PR #62, opened 2018-08-29
Use `any` instead of `map`

mwajeeh:
Save some CPU cycle by not doing a `map` first and then filtering.


gpeal:
@mwajeeh Can you rebase this to pick up the CI checks which we just re-enabled? LGTM to me if it goes green.

mwajeeh:
@gpeal I just did rebase and its still not picking it up. Am I missing something here?

gpeal:
It turns out that rebasing doesn't add the checks for whatever reason. LGTM though

PR #64, opened 2018-08-29
Reenable CI checks

gpeal:
Also fixed a few lint errors


.travis.yml
```diff
@@ -0,0 +1,45 @@
+language: android
+
+android:
+  components:
+  - tools
+  - platform-tools
+  - build-tools-27.0.3
```
elihart: should we be using 28?

gpeal: @elihart actually, everything else is on 27 so let's do that separately.

PR #67, opened 2018-08-29
Remove unnecessary marshaling

Oknesif:
It is better to store Class fields in MvRxPersistedViewModelHolder than String, it is more readable and error-prone.


gpeal:
@Oknesif This looks great. Can you rebase to pick up the CI checks which we just reenabled? Other than that, LGTM!

gpeal:
Thanks @Oknesif I'll just run the checks locally but thanks for doing this!

PR #72, opened 2018-08-30
Potential approaches for allowing mocked state

elihart:
Playing with some ways to mock out state for testing. Each commit here is a separate approach.

Use an intermediate mocker singleton to control all mocked state
https://github.com/airbnb/MvRx/pull/72/commits/6ffc9c77be63994a3b1d83b446b2710f97db5d8f

Set a mock provider on the view model
https://github.com/airbnb/MvRx/pull/72/commits/d508554af7a894be4b220de14f1af4e18678a111

Definitely want to avoid making it too easy to set state so people don't use this outside of testing.

### DIFF
--- a/mvrx/src/main/kotlin/com/airbnb/mvrx/BaseMvRxViewModel.kt
+++ b/mvrx/src/main/kotlin/com/airbnb/mvrx/BaseMvRxViewModel.kt
@@ -10,6 +10,7 @@ import io.reactivex.Single
 import io.reactivex.android.schedulers.AndroidSchedulers
 import io.reactivex.disposables.CompositeDisposable
 import io.reactivex.disposables.Disposable
+import io.reactivex.disposables.Disposables
 import io.reactivex.functions.Consumer
 import io.reactivex.schedulers.Schedulers
 import kotlin.reflect.KProperty1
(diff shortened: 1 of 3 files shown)


mvrx/src/main/kotlin/com/airbnb/mvrx/BaseMvRxViewModel.kt
```diff
@@ -57,17 +58,16 @@ abstract class BaseMvRxViewModel<S : MvRxState>(
...
-                val firstState = this.reducer()
-                val secondState = this.reducer()
-                if (firstState != secondState) throw IllegalArgumentException("Your reducer must be pure!")
-                firstState
-            }
-        } else {
-            stateStore.set(reducer)
+        when {
+            debugMode -> // Must use `set` to ensure the validated state is the same as the actual state used in reducer
+                // Do not use `get` since `getState` queue has lower priority and the validated state would be the state after reduced
```
hellohuanlin: these 2 comments are outdated. can you delete them?

mmcwong:
I agree both approaches seem like they would work. I kind of like being able to have all the mock states for different view models in one place and a single place to enable/disable all of them though.

elihart:
Closed in favor or https://github.com/airbnb/MvRx/pull/81

PR #73, opened 2018-08-30
Use correct base classes in README.md

markushi:
I just added two minor tweaks to the README.md to correctly reflect the actual API.

### DIFF
--- a/README.md
+++ b/README.md
@@ -11,11 +11,11 @@ This is what it looks like:
 
 data class HelloWorldState(val title: String = "Hello World") : MvRxState
 
-class HelloWorldViewModel(initialState: HelloWorldState) : MvRxViewModel<HelloWorldState>(initialState) {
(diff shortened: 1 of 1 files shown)


gpeal:
It's not correct to use the Base one directly. The [integration instructions](https://github.com/airbnb/MvRx/wiki/Integrating-MvRx-In-Your-App#steps-required) suggest that you create your own base class.

I think that could be clearer though. How could we update the readme to make that more clear?

gpeal:
@markushi

markushi:
Oh right, looks like I missed that - sorry. 
Well I guess it's a bit tricky. As a first time user I'd expect to see a working example. On the other side it's important to see that MvRx doesn't force you to a specific inheritance model.

I guess the confusing part for me was that `MvRxViewModel` is not part of MvRx. How about renaming it to `BaseViewModel` or `MyBaseViewModel`? Then it should be clear that this class is not part of MvRx.

gpeal:
Done :) Thanks for the suggestion.

PR #75, opened 2018-08-31
Remove subscribeOn and observeOn thread for observable execution

hellohuanlin:
It can be unexpected to change upstream thread in VM. MvRx should only switch downstream thread. 

For example: 

```
// in thread A context
inMemoryDB
  .fetchStuffWhichIsFastOperation() // Developer intend to run it on thread A, we shouldn't run it on merge thread. (Assuming this function doesn't have a `subscribeOn`)
  .execute {
    // merge thread here, which is expected
  }
```

Also removed `observeOn(backgroundThread)`, since `setState` is thread safe, and it can be run on any thread. (Note that `setState` only enqueues the block, while the actual execution of block is still in merge thread. )

@gpeal @BenSchwab @elihart

### DIFF
--- a/mvrx/src/main/kotlin/com/airbnb/mvrx/BaseMvRxViewModel.kt
+++ b/mvrx/src/main/kotlin/com/airbnb/mvrx/BaseMvRxViewModel.kt
@@ -27,7 +27,6 @@ abstract class BaseMvRxViewModel<S : MvRxState>(
 ) : ViewModel() {
     private val tag by lazy { javaClass.simpleName }
     private val disposables = CompositeDisposable()
-    private val backgroundScheduler = Schedulers.single()
     private lateinit var mutableStateChecker: MutableStateChecker<S>
 
     init {
@@ -144,7 +143,6 @@ abstract class BaseMvRxViewModel<S : MvRxState>(
         successMetaData: ((T) -> Any)? = null,
         stateReducer: S.(Async<V>) -> S
     ): Disposable {
-        // This will ensure that Loading is dispatched immediately rather than being posted to `backgroundScheduler` before emitting Loading.
         setState { stateReducer(Loading()) }
 
         return map {
--- a/mvrx/src/main/kotlin/com/airbnb/mvrx/BaseMvRxViewModel.kt
+++ b/mvrx/src/main/kotlin/com/airbnb/mvrx/BaseMvRxViewModel.kt
@@ -148,7 +148,6 @@ abstract class BaseMvRxViewModel<S : MvRxState>(
         setState { stateReducer(Loading()) }
 
         return observeOn(backgroundScheduler)
-            .subscribeOn(backgroundScheduler)
             .map {
                 val success = Success(mapper(it))
                 success.metadata = successMetaData?.invoke(it)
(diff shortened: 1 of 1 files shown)


BenSchwab:
Makes sense to me 👍

hellohuanlin:
As discussed offline, let's remove both `subscribeOn` and `observeOn`

gpeal:
@hellohuanlin @BenSchwab I think this is the right move overall but I'm afraid that it might make the default/simple case for many apps more cumbersome. 

I have a feeling that as-is, people will either start putting the schedulers in every execute call which adds a lot of boilerplate for a concept we would like to abstract away or people will accidentally do some operations on the main thread that shouldn't be there.

Thoughts?

hellohuanlin:
@gpeal @BenSchwab @elihart 
I think there are 2 separate problems here: 

1. subscribeOn
Setting it in MvRx feels like encouraging bad design and hide bugs. Whoever provides the source is "supposed" to decide which thread to run. MvRx knows nothing about what the source is, thus can't assume that background thread is the best choice for all. 

It also makes the API confusing if developers rely on our `subscribeOn` to decide what thread the source should run on. Because it's possible that the source already specified `subscribeOn`, and the second one would be a no-op. i.e. From developers point of view, "sometimes it works. sometimes it doesn't". 

2. observeOn
Specifying observeOn is unnecessary. `setState` which enqueues (not execute) the reducer is a fast operation and thread safe. It's designed to be called from any thread (including main thread)

mvrx/src/main/kotlin/com/airbnb/mvrx/BaseMvRxViewModel.kt
```diff
@@ -140,9 +140,7 @@ abstract class BaseMvRxViewModel<S : MvRxState>(
         // This will ensure that Loading is dispatched immediately rather than being posted to `backgroundScheduler` before emitting Loading.
         setState { stateReducer(Loading()) }
 
-        return observeOn(backgroundScheduler)
-            .subscribeOn(backgroundScheduler)
```
BenSchwab: @hellohuanlin Can we delete the `backgroundScheduler` now?

hellohuanlin:
@BenSchwab @gpeal @elihart 
re: the risk of a main thread observable being slowed down due to lock contention from other observables

Lock is only held for enqueue/dequeue, which is short lock. In fact we do enqueue from different threads in current implementation: 

1. merge thread here (when `withState` enqueues a `setState` reducer)
https://github.com/airbnb/MvRx/blob/88e9ac22d4e0cab23acb5418cbd07a7acac06186/mvrx/src/main/kotlin/com/airbnb/mvrx/MvRxStateStore.kt#L123

2. any thread (most likely main thread) that calls this function
https://github.com/airbnb/MvRx/blob/88e9ac22d4e0cab23acb5418cbd07a7acac06186/mvrx/src/main/kotlin/com/airbnb/mvrx/BaseMvRxViewModel.kt#L59

3. VM's `background` thread that i'm trying to remove in this PR.

BenSchwab:
@hellohuanlin LGTM after deleting the background scheduler

hellohuanlin:
updated PR @gpeal @BenSchwab @elihart

gpeal:
@hellohuanlin Can you create and update a CHANGELOG. This is an important change to note.

@BenSchwab Can you make sure that the AirRequest extension includes the right schedulers then?

gpeal:
@BenSchwab @hellohuanlin what is the status of this?

hellohuanlin:
@gpeal @BenSchwab Just rebased. Let's include a ChangeLog for `0.6.0` release

PR #81, opened 2018-09-05
Hookup MvRxMocker and add tests

mmcwong:
This PR builds on to https://github.com/airbnb/MvRx/pull/72
It finishes up setting state directly and through args. 
It also adds a test for the `MvRxMocker`

@elihart

### DIFF
--- a/mvrx/src/main/kotlin/com/airbnb/mvrx/MvRxStateStore.kt
+++ b/mvrx/src/main/kotlin/com/airbnb/mvrx/MvRxStateStore.kt
@@ -78,6 +78,8 @@ internal open class MvRxStateStore<S : Any>(
      * all of the code required.
      */
     fun set(stateReducer: S.() -> S) {
+        if (MvRxMocker.getMockedState(this) != null) return
+
         jobs.enqueueSetStateBlock(stateReducer)
         flushQueueSubject.onNext(Unit)
     }
--- a/mvrx/src/main/kotlin/com/airbnb/mvrx/MvRxStateStore.kt
+++ b/mvrx/src/main/kotlin/com/airbnb/mvrx/MvRxStateStore.kt
@@ -20,7 +20,6 @@ internal open class MvRxStateStore<S : Any>(
 ) : Disposable {
 
     internal val isMocked = MvRxMocker.enabled
-
     /**
      * The subject is where state changes should be pushed to.
      */
@@ -79,8 +78,6 @@ internal open class MvRxStateStore<S : Any>(
      * all of the code required.
      */
     fun set(stateReducer: S.() -> S) {
-        if (mocker != null) return
-
         jobs.enqueueSetStateBlock(stateReducer)
         flushQueueSubject.onNext(Unit)
     }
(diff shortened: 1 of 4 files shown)


mvrx/src/main/kotlin/com/airbnb/mvrx/TestUtil.kt
```diff
@@ -1,23 +1,21 @@
...
 
 object MvRxMocker {
 
     var enabled: Boolean = false
-    private val mockedState: MutableMap<BaseMvRxViewModel<*>, Any> = mutableMapOf()
+    private val mockedState: MutableMap<MvRxStateStore<*>, MvRxState> = mutableMapOf()
 
-     fun <S : MvRxState> setMockedState(viewModel: BaseMvRxViewModel<S>, state: S) {
-        mockedState[viewModel] = state
+    fun <S : MvRxState> setMockedState(viewModel: BaseMvRxViewModel<S>, state: S) {
```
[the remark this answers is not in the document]
mmcwong: yeah that makes sense will address in another PR

mmcwong:
@elihart updated with your feedback and changed the base branch

mvrx/src/main/kotlin/com/airbnb/mvrx/MvRxStateStore.kt
```diff
@@ -15,7 +15,11 @@ import java.util.LinkedList
  * conditions with each other.
  *
  */
-internal open class MvRxStateStore<S : Any>(initialState: S) : Disposable {
+internal open class MvRxStateStore<S : Any>(
+    initialState: S
+) : Disposable {
+
+    internal val isMocked = MvRxMocker.enabled
```
elihart: I think you need `get() =`, otherwise this won't update

mvrx/src/main/kotlin/com/airbnb/mvrx/MvRxStateStore.kt
```diff
@@ -72,6 +78,8 @@ internal open class MvRxStateStore<S : Any>(initialState: S) : Disposable {
      * all of the code required.
      */
     fun set(stateReducer: S.() -> S) {
+        if (MvRxMocker.getMockedState(this) != null) return
```
elihart: do you see reasons to only block this if mocked state is set? I think it should be just
```
if (isMocked)
```

Because the view model or fragment might do things to set state before the mock is in place (the mock can only be set once the view model is created). Since setting state is async and can trigger subscribe callbacks it would be nice to eliminate any side effects from that

mvrx/src/main/kotlin/com/airbnb/mvrx/TestUtil.kt
```diff
@@ -0,0 +1,38 @@
...
+ * This singleton can be used to provide mocked states to viewModels for testing.
+ * State can be directly set or set via arguments. Set the mocker to enabled
+ * to have the viewModels use these mocked states.
+ */
+object MvRxMocker {
+
+    /**
+     * Enable or disable whether the viewModels use mocked states
+     */
+    var enabled: Boolean = false
```
BenSchwab: Using static variables can cause a lot of headaches for tests. This will not be reset between tests. So if some test turns this on and forgets to turn it off it will stay on for all remaining tests.

mmcwong:
Closing as https://github.com/airbnb/MvRx/pull/88 is a more testable approach

PR #89, opened 2018-09-15
Fix ordering issue of setState calls within withState 

BenSchwab:
`push` adds to the front of the queue, but we want to add to the end to preserve order of `setState`. The only time the queue will have multiple elements is within a call to `withState`. This is because the `withState` block itself will be on the background thread, so `setState` calls will not run immediately. Currently this test fails with the value of `foo` on the second call being `1` as the ordering of the setState was reversed:

```
 withState {
            selectSubscribe(ViewModelTestState::foo) {
                callCount++
                assertEquals(when (callCount) {
                    1 -> 0
                    2 -> 2
                    else -> throw IllegalArgumentException("Unexpected call count $callCount")
                }, it)
            }
            // These will be coalesced into one update
            setState { copy(foo = 1) }
            setState { copy(foo = 2) }
        }
        assertEquals(2, callCount)
```

Now it passes. Note -- that this also tests the coalescing logic of setState's within with state.


BenSchwab:
This may be related to what was seen here: https://github.com/airbnb/MvRx/issues/76

gpeal:
@BenSchwab Nice!

PR #90, opened 2018-09-15
Add debug validation for state immutability

elihart:
Somebody recently was facing buggy behavior in their UI code and asked for help to figure it out. Taking a look at their code their state looked like this

```kotlin
data class FormState(
    val countryInputs: MutableMap<BusinessAccountFormInput, Country?> = HashMap(),
    val textInputs: MutableMap<BusinessAccountFormInput, String?> = HashMap(),
    val dateInputs: MutableMap<BusinessAccountFormInput, AirDate?> = HashMap(),
```

and they were updating it like this
```kotlin
val currentTextInputs = withState(verificationViewModel) { it.textInputs }
currentTextInputs[input] = it
 verificationViewModel.setTextInput(currentTextInputs)
```

This was not working because their state objects were sharing mutable data and weren't actually changing.

Our current debug checks can only validate that types like HashMap are not used, but cannot check for simple MutableMap. This change lets us detect at runtime any mutable state (as long as it causes hashcode changes).

### DIFF
--- a/mvrx/src/main/kotlin/com/airbnb/mvrx/BaseMvRxViewModel.kt
+++ b/mvrx/src/main/kotlin/com/airbnb/mvrx/BaseMvRxViewModel.kt
@@ -28,10 +28,14 @@ abstract class BaseMvRxViewModel<S : MvRxState>(
     private val disposables = CompositeDisposable()
     private val backgroundScheduler = Schedulers.single()
     private val stateStore: MvRxStateStore<S> = MvRxStateStore(initialState)
+    private lateinit var mutableStateChecker: MutableStateChecker<S>
 
     init {
         if (debugMode) {
-            Observable.fromCallable { validateState(initialState) }.subscribeOn(Schedulers.computation()).subscribe()
+            mutableStateChecker = MutableStateChecker(initialState)
+
+            Observable.fromCallable { validateState(initialState) }
+                .subscribeOn(Schedulers.computation()).subscribe()
         }
     }
 
@@ -63,7 +67,10 @@ abstract class BaseMvRxViewModel<S : MvRxState>(
             stateStore.set {
                 val firstState = this.reducer()
                 val secondState = this.reducer()
+
                 if (firstState != secondState) throw IllegalArgumentException("Your reducer must be pure!")
+                mutableStateChecker.onStateChanged(firstState)
+
                 firstState
             }
         } else {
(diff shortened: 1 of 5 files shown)


mvrx/src/main/kotlin/com/airbnb/mvrx/MvRxMutabilityHelper.kt
```diff
@@ -41,3 +44,31 @@ internal fun KClass<*>.assertImmutability() {
...
+
+/**
+ * Checks that a state's value is not changed over its lifetime.
+ */
+internal class MutableStateChecker<S : MvRxState>(val initialState: S) {
+
+    data class StateWrapper<S : MvRxState>(val state: S) {
+        private val originalHashCode = hashCode()
+
+        fun validate() = require(originalHashCode == hashCode()) {
```
hellohuanlin: looks like every state hash is computed twice

mvrx/src/main/kotlin/com/airbnb/mvrx/BaseMvRxViewModel.kt
```diff
@@ -28,10 +28,14 @@ abstract class BaseMvRxViewModel<S : MvRxState>(
     private val disposables = CompositeDisposable()
     private val backgroundScheduler = Schedulers.single()
     private val stateStore: MvRxStateStore<S> = MvRxStateStore(initialState)
+    private lateinit var mutableStateChecker: MutableStateChecker<S>
```
hellohuanlin: is lateinit necessary? can it simply be `private val mutableStateChecker = MutableStateChecker(initialState)`

mvrx/src/main/kotlin/com/airbnb/mvrx/MvRxMutabilityHelper.kt
elihart: yes, it has to be. once when it is first set and once when the next state is set, then we can compare the hashcodes to see if they have changed.

mvrx/src/main/kotlin/com/airbnb/mvrx/BaseMvRxViewModel.kt
elihart: you're right, it's not necessary, it was done this way to avoid calculating the hashcode  of the state object on class init in production.

that's a bit minor, but I didn't want to affect prod at all. I could add a comment about the purpose

mvrx/src/main/kotlin/com/airbnb/mvrx/MvRxMutabilityHelper.kt
hellohuanlin: ic. i think the implementation can be simplified as just 2 lines of code: 

```
            stateStore.set {
                val oldHash = hash() // add this
                val firstState = reducer()
                val secondState = reducer()
                 if (firstState != secondState) throw IllegalArgumentException("Your reducer must be pure!")
                if (oldHash != hash()) throw IllegalArgumentException("You should not mutate the state")  // add this
                 firstState
            }
```

And in this way we don't have to worry about lateinit the MutableStateChecker in the comment below.

hellohuanlin: ah, you are right!

PR #113, opened 2018-10-04
Don't re-emit state value stop->start if the state did not change

BenSchwab:
As brought up in https://github.com/airbnb/MvRx/issues/94 there seems to be no need to remit the state value during state transitions of `stopped` -> `started` unless the value changed.

We will alway call `invalidate` on any call to `started`.

Closes https://github.com/airbnb/MvRx/issues/94

### DIFF
--- a/mvrx/src/main/kotlin/com/airbnb/mvrx/BaseMvRxViewModel.kt
+++ b/mvrx/src/main/kotlin/com/airbnb/mvrx/BaseMvRxViewModel.kt
@@ -146,10 +146,10 @@ abstract class BaseMvRxViewModel<S : MvRxState>(
         setState { stateReducer(Loading()) }
 
         return map {
-                val success = Success(mapper(it))
-                success.metadata = successMetaData?.invoke(it)
-                success as Async<V>
-            }
+            val success = Success(mapper(it))
+            success.metadata = successMetaData?.invoke(it)
+            success as Async<V>
+        }
             .onErrorReturn { Fail(it) }
             .subscribe { asyncData -> setState { stateReducer(asyncData) } }
             .disposeOnClear()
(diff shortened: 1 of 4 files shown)


mvrx/src/test/kotlin/com/airbnb/mvrx/ViewSubscriberTest.kt
```diff
@@ -15,20 +15,20 @@ class ViewSubscriberFragment : BaseMvRxFragment() {
     private val viewModel: ViewSubscriberViewModel by fragmentViewModel()
 
     var subscribeCallCount = 0
-    var selectSubscribeCalled = 0
+    var selectSubscribeCalled = -1
```
BenSchwab: This needs to be -1 to properly assert what we want, otherwise the first call to the selectSusbcribe lambda sets 0 to 0

mvrx/src/test/kotlin/com/airbnb/mvrx/ViewModelSubscriberTest.kt
```diff
@@ -475,38 +475,54 @@ class ViewModelSubscriberTest : BaseTest() {
...
+
+        viewModel.setFoo(1)
+        assertEquals(1, callCount)
+
+        owner.lifecycle.handleLifecycleEvent(Lifecycle.Event.ON_START)
         assertEquals(2, callCount)
     }
 
     @Test
-    fun testSubscribeCalledOnRestart() {
```
BenSchwab: This assertion this test makes is no longer true.

mvrx/src/main/kotlin/com/airbnb/mvrx/BaseMvRxViewModel.kt
```diff
@@ -167,35 +167,37 @@ abstract class BaseMvRxViewModel<S : MvRxState>(
      * For ViewModels that want to subscribe to itself.
      */
     protected fun subscribe(subscriber: (S) -> Unit) =
-        stateStore.observable.subscribeLifecycle(null, subscriber)
+        stateStore.observable.subscribeLifecycle(null, false, subscriber)
```
BenSchwab: ViewModels subscriptions don't have to worry about lifecycles, so this is not relevant.

BenSchwab:
Updated to make `uniqueOnly` configurable. The default behavior is backwards compatible.

gpeal:
@BenSchwab

BenSchwab:
Change log here: https://github.com/airbnb/MvRx/pull/124

PR #138, opened 2018-11-15
Add default proguard config for sample app

rossbacher:
Update AGP to 3.2.0
Update Kotlin to 1.2.51
Include consumer proguard file in every variant
Add default proguard config for sample app
Run Proguard on release build type for sample app

Fixes: https://github.com/airbnb/MvRx/issues/126

### DIFF
--- a/mvrx/build.gradle
+++ b/mvrx/build.gradle
@@ -12,11 +12,7 @@ android {
     defaultConfig {
         minSdkVersion 16
         targetSdkVersion 27
-    }
-    buildTypes {
-        release {
-            consumerProguardFiles 'proguard-rules.pro'
-        }
+        consumerProguardFiles 'proguard-rules.pro'
     }
     lintOptions {
         abortOnError true
--- a/sample/proguard-project.pro
+++ b/sample/proguard-project.pro
@@ -5,6 +5,13 @@
 # Print out the full proguard config used for every build
 -printconfiguration build/outputs/fullProguardConfig.pro
 
+
+#
+# Keep rules that are used because MvRx uses Kotlin and Kotlin reflection. These are not defined in the lib
+# as they are not specific to the lib itself but need to be most likley present in any project that uses Kotlin 
+# and Kotlin reflection.
+#
+
 # These classes are used via kotlin reflection and the keep might not be required anymore once Progurad supports
 # Kotlin reflection directlty.
 -keep class kotlin.reflect.jvm.internal.impl.builtins.BuiltInsLoaderImpl
@@ -27,14 +34,6 @@
 # https://stackoverflow.com/questions/33547643/how-to-use-kotlin-with-proguard
 -dontwarn kotlin.**
 
-# https://app.bugsnag.com/airbnb/android-1/errors/59f6c07487a5c5001a49d882?filters[event.since][0]=all&filters[event.severity][0][value]=error&filters[event.severity][0][type]=eq
-# https://github.com/square/okhttp/issues/3582
--keep class okhttp3.internal.publicsuffix.PublicSuffixDatabase
-
-# The code is safeguarded against API28 use so we can just ignore the warning bout Handler.createAsync(looper) not existing.
-# Can be removed when compiled against API >=28
--dontwarn com.airbnb.epoxy.EpoxyAsyncUtil
-
 # RxJava
 -keepclassmembers class rx.internal.util.unsafe.*ArrayQueue*Field* {
    long producerIndex;
@@ -43,6 +42,14 @@
 
 # Oddly need to keep that even though Evernote state is not used in the app.
 -keepnames class * { @com.evernote.android.state.State *;}
+
+#
+# Keep rules that are unique to the sample project because it uses epoxy, okhttp3, moshi and retrofit.
+#
+
+# The code is safeguarded against API28 use so we can just ignore the warning bout Handler.createAsync(looper) not existing.
+# Can be removed when compiled against API >=28
+-dontwarn com.airbnb.epoxy.EpoxyAsyncUtil
 
 #
 # From: https://github.com/square/retrofit/blob/master/retrofit/src/main/resources/META-INF/proguard/retrofit2.pro
@@ -70,7 +77,6 @@
 -dontwarn retrofit2.-KotlinExtensions
 
 
-
 #
 # From https://github.com/square/okhttp/blob/master/okhttp/src/main/resources/META-INF/proguard/okhttp3.pro
 #
@@ -86,7 +92,6 @@
 
 # OkHttp platform used only on JVM and when Conscrypt dependency is available.
 -dontwarn okhttp3.internal.platform.ConscryptPlatform
-
 
 
 #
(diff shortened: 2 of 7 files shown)


sample/proguard-project.pro
```diff
@@ -0,0 +1,163 @@
...
+# https://app.bugsnag.com/airbnb/android-1/errors/59f6c07487a5c5001a49d882?filters[event.since][0]=all&filters[event.severity][0][value]=error&filters[event.severity][0][type]=eq
+# https://github.com/square/okhttp/issues/3582
+-keep class okhttp3.internal.publicsuffix.PublicSuffixDatabase
+
+# The code is safeguarded against API28 use so we can just ignore the warning bout Handler.createAsync(looper) not existing.
+# Can be removed when compiled against API >=28
+-dontwarn com.airbnb.epoxy.EpoxyAsyncUtil
+
+# RxJava
+-keepclassmembers class rx.internal.util.unsafe.*ArrayQueue*Field* {
```
bernaferrari: Is this really needed? I thought RxJava already worked great with ProGuard.

sample/proguard-project.pro
```diff
@@ -0,0 +1,163 @@
...
+   long producerIndex;
+   long consumerIndex;
+}
+
+# Keep the models, this can be done in a better way but for this sample we just keep all the models
```
bernaferrari: What is the 'better way' you recommend?

sample/proguard-project.pro
```diff
@@ -0,0 +1,163 @@
...
+-keep class com.airbnb.mvrx.sample.models.** {
+  *;
+}
+
+# Oddly need to keep that even though Evernote state is not used in the app.
```
bernaferrari: I understand this as much as you, i.e., nothing. Why is it necessary?

mvrx/build.gradle
```diff
@@ -12,11 +12,7 @@ android {
     defaultConfig {
         minSdkVersion 16
         targetSdkVersion 27
```
bernaferrari: Since you are updating everything, you could also change this to 28.

sample/proguard-project.pro
rossbacher: We found that we needed that in our app, it might not be needed here (so is the full config for retrofit, moshi and okhttp3) it is more here for completeness, as this app uses RxJava, it should have the config that is needed for full support.

bernaferrari: Oh, I see, you just copied the Proguard from Airbnb?

sample/proguard-project.pro
rossbacher: The better way  is to define the model in a way that the getter has the `@Json` annotation with the name of the field. That way the default Moshi rules work already:

```
@get:Json(name = "id") @Json(name = "id") val id: String,
```

I'll make this modification. I was just to lazy to do it the right way in a simple demo project but then on the other hand, the demo project is where it should be done right.

sample/proguard-project.pro
rossbacher: It is really odd, and I sadly do not have enough time to dig what exactly is going wrong here, it was not an easy thing.

mvrx/build.gradle
rossbacher: For the sdk it does not make sense to have an artificially high target SDK. As it might pull up the garget SDK requirements for implementers without any need.

sample/proguard-project.pro
bernaferrari: Awesome. I just tested your commit, working perfectly!

sample/proguard-project.pro
rossbacher: I did not, our file is way more complicated. :-) 

I just copied the sections for the libs that overlap and are not taken care of the libs themselves.

sample/proguard-project.pro
rossbacher: I just pushed the change with the model change.

bernaferrari:
A few more suggestions: you named proguard-project.pro instead of proguard-rules.pro. Also, I counted `-dontwarn javax.annotation.**` there 3 times.

rossbacher:
@bernaferrari This is on purpose, I want to make clear that these are project specific rules. The name of the file is not really a problem.

`-dontwarn javax.annotation.**` is in there multiple times as it is in the different configs I had to copy for `okhttp3`, `moshi` and `retrofit2`. These configs are unchanged on purpose. It is ok to have duplicative config.

bernaferrari:
Fine. I tried the new configurations on my app that crashes with MvRx. I was able to make it stop crashing with the following configuration:

```
# Print out the full proguard config used for every build
-printconfiguration build/outputs/fullProguardConfig.pro

# These classes are used via kotlin reflection and the keep might not be required anymore once Progurad supports
# Kotlin reflection directlty.
-keep class kotlin.reflect.jvm.internal.impl.builtins.BuiltInsLoaderImpl
-keep class kotlin.reflect.jvm.internal.impl.load.java.FieldOverridabilityCondition
-keep class kotlin.reflect.jvm.internal.impl.load.java.ErasedOverridabilityCondition
-keep class kotlin.reflect.jvm.internal.impl.load.java.JavaIncompatibilityRulesOverridabilityCondition

# If Companion objects are instantiated via Kotlin reflection and they extend/implement a class that Proguard
# would have removed or inlined we run into trouble as the inheritance is still in the Metadata annoation
# read by Kotlin reflection.
# FIXME Remove if Kotlin reflection is supported by Pro/Dexguard
-if class **$Companion extends **
-keep class <2>
-if class **$Companion implements **
-keep class <2>

# https://medium.com/@AthorNZ/kotlin-metadata-jackson-and-proguard-f64f51e5ed32
-keep class kotlin.Metadata { *; }

# Oddly need to keep that even though Evernote state is not used in the app.
-keepnames class * { @com.evernote.android.state.State *;}
```

I suggest adding it to the library, since everyone using mvrx+epoxy+proguard will get a crash without it (probably because of reflection), which is not fun.

bernaferrari:
I think you should at least add on wiki that mvrx crashes without some proguard rules, else this will catch a lot of people on surprise.

rossbacher:
Well the rules are now in the demo app. But I agree, let me update the README...

bernaferrari:
If I were you I would probably split in the demo what is necessary for the
library and what is necessary for the demo (which include retrofit and
okhttp that is independent of mvrx).

rossbacher:
I also updated the Wiki with a Proguard/Dexguard section here: https://github.com/airbnb/MvRx/wiki#proguarddexguard

rossbacher:
> I think you should at least add on wiki that mvrx crashes without some proguard rules, else this will catch a lot of people on surprise.

Done: https://github.com/airbnb/MvRx/wiki#proguarddexguard

rossbacher:
> I also updated the Wiki with a Proguard/Dexguard section here: https://github.com/airbnb/MvRx/wiki#proguarddexguard

Done.

bernaferrari:
Great, awesome, thanks!

PR #147, opened 2018-11-29
Move MvRx to AndroidX

rossbacher:
- Update all dependencies and usages to AndroidX
- Update to AGP 3.2 and Kotlin 1.2.41

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -1,6 +1,6 @@
 buildscript {
     ext.kotlinVersion = '1.2.41'
-    ext.supportLibVersion = '27.1.1'
+    ext.supportLibVersion = '28.0.0'
     ext.lifecycleVersion = '1.1.1'
     ext.robolectricVersion = '3.8'
     ext.epoxyVersion = '2.16.4'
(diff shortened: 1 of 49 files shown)


build.gradle
```diff
@@ -1,21 +1,26 @@
 buildscript {
-    ext.kotlinVersion = '1.2.41'
-    ext.supportLibVersion = '27.1.1'
-    ext.lifecycleVersion = '1.1.1'
+    ext.kotlinVersion = '1.2.51'
+    ext.androidxVersion = '1.0.0'
```
bernaferrari: While this works, they are being updated independently (there is already, for example, androix.appcompat 1.0.2, which fixes important issues with PrecomputedTextCompat, while RecyclerView is still at 1.0.0)

rossbacher: Good point and I wanted to break this outer later. I did not realize that an independent bugfix release already happened. Let me update that.

gpeal:
@rossbacher Let's hold off on merging this until we do one final release with any changes we want right before it.

rossbacher:
Replaced by https://github.com/airbnb/MvRx/pull/194

Issue #69, opened 2018-08-30
Work with Dagger


chrisbanes:
At the moment I see no way of integrating with an existing ViewModelProvider.Factory. This is necessary to be able to use Dagger to inject ViewModels.

[Example fragment](https://github.com/chrisbanes/tivi/blob/master/app/src/main/java/app/tivi/showdetails/details/ShowDetailsFragment.kt#L63)

Is there a way?

chrisbanes:
I settled on the following for now [commit](https://github.com/chrisbanes/tivi/blob/06727520434034e6c54b4a8b112d050872c96a31/app/src/main/java/app/tivi/showdetails/details/ShowDetailsFragmentViewModel.kt#L58)

I inject the ViewModel.Factory into the Activity, then that to create the VM. Works nicely but we do lose the initial state nice-ness.

gpeal:
@chrisbanes The reason we moved away from that is because initialState is pretty critical to MvRx. This would prevent a ViewModel from being able to use Fragment args or for MvRx to do proper ViewModel restoration in a new process.

Thoughts?

chrisbanes:
Yeah, we don't have assisted DI with Dagger so it's hard to get the initialState + fragment args working nicely.

I managed to get it sort of working, but I can't rely on `init {}` and have to manually poke the VM to use the fragment args: [here](https://github.com/chrisbanes/tivi/blob/cacf79fed811956b121fd17e742aca764e5a136e/app/src/main/java/app/tivi/showdetails/details/ShowDetailsFragmentViewModel.kt#L58).

I might have a play with AutoFactory and see if I can get it to work.

chrisbanes:
Renaming this to be clearer what the goal is (for me anyway).

I got AutoFactory working [here](https://github.com/chrisbanes/tivi/pull/212/commits/aff85d779d26db1ccbbe274ea60b3bd7e8e293ed). The only downside I see is that `MvRxViewModelFactory` only passes along the Activity, whereas I would much rather inject the related fragment instead.

Can we get a version of `MvRxViewModelFactory` which provides the Fragment?

mzgreen:
@chrisbanes have you seen this: https://github.com/square/AssistedInject ? Maybe it'll be helpful in this scenario?

chrisbanes:
@mzgreen Ah yes! I've been wanting a reason to try it but totally forgot about it. I'll try on Monday.

chrisbanes:
[...]

PR #148, opened 2018-12-07
Allow for Fragment ViewModel factory (#69)

marukami:
Injecting with Dagger #69 is very hard without the Fragment been passed for Fragment `ViewModels`. So, I added a `MvRxFragmentViewModelFactory` as another factory type. To keep backward compatibility if the `Fragment` factory can't be found we can fall back to the `Activity` factory. If no factory can be found then like before try the single state arg `ViewModel` constructor before crashing as it would before.


marukami:
Sure, no worries. I added a few tests and fixed an issue they found.

PR #150, opened 2018-12-11
Upgrade to support lib 28.0.0 

BenSchwab:
This PR bumps MvRx to 28 sdk and support libs. It includes a necessary internal change of always restoring `fragmentArgsForActivityViewModelState` map during `restoreViewModels`.

When a fragment viewModel is scoped to an activity, we store the fragment args in the savedInstanceState of the activity. This is to make it accessible during the restoration process.

To repro crash:
- use an `activityViewModel` in a fragment
- cause a configuration change (rotate phone)
- Background app, make sure process is killed
- restart app.

During the configuration change `fragmentArgsForActivityViewModelState` was never fully restored, because the viewModelMap was empty. Therefore, in `onSaveInstanceState` during the process death, the args are lost, causing a crash when the app restarts.

This PR, moves the restoration before the early terminate in `restoreViewModels`.

@gpeal @elihart

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,7 +6,7 @@ android:
   - platform-tools
   - build-tools-28.0.2
   - build-tools-28.0.3
-  - android-27
+  - android-28
   - extra-google-google_play_services
   - extra-android-m2repository
   - extra-android-support
--- a/build.gradle
+++ b/build.gradle
@@ -7,7 +7,7 @@ buildscript {
     ext.moshiVersion = '1.6.0'
     ext.koinVersion = '0.9.3'
     ext.retrofitVersion = '2.4.0'
-    ext.navVersion = '1.0.0-alpha04'
+    ext.navVersion = '1.0.0-alpha08'
     ext.roomVersion = '1.1.1'
     ext.buildToolsVersion = '28.0.3'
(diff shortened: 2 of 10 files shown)


BenSchwab:
@gpeal @elihart 
So after doing more research, I found that the crash only happens with support lib 28.0.0. (The crash first started appearing at Airbnb after our upgrade.)

The root cause is a bit complex, will document if for posterity's sake:

FragmentActivity essentially has [a bug prior to 28.0.0](https://github.com/aosp-mirror/platform_frameworks_support/commit/76366948d0551a322262612404eca4c60fa29224#diff-1f343e9c6419dde9b204081fce5510f5) where if a fragment accesses an activity view model store during `onCreate`, it will get a different instance, because FragmentActivity used to only restore the view model store after calling super.onCreate.

So this actually caused the bug I first explained is the PR to not occur, because the ViewModelStore map would always be empty during a call to `restoreViewModels`. 

After 28.0.0. this map is restored when `getViewModel()` is called, which is the expected behavior. Thus, the view model map is no longer null, and the restore crash occurs.

So it seems appropriate to also bump the target/compile sdk and support libs in this PR to 28.

PR #92, opened 2018-09-16
Todomvrx unit tests

rom4ek:
Hey guys, thanks for the nice library! I was just curious about testability of the architecture (Closes #31), and decided to give it a try, so don't judge me hard :) As you don't expose the `debugMode` parameter, I was not sure how do you guys mock the `BuildConifg` behavior at airbnb, but I've done that via small `SuperReflect` lib, just so I don't like PowerMock, and that was a minor thing that prevented me from unit-testing the view model. I also didn't add tests for combining different data sources, but it can be ofc easily done.

PR #154, opened 2018-12-13
Created a testing artifact and MvRxTestRule

gpeal:
This PR creates `MvRxTestRule` and has two options:
1) Force debug mode to be enabled/disabled for testing (fixes an issues for #92).
2) Set the rx schedulers to be immediate so all reducers and other operations run synchronously.

I also simplified the travis steps since we would have to add steps for every module. They also run in parallel so it ends up doing a lot of duplicate recompilation as separate steps.


testing/src/main/kotlin/com/airbnb/mvrx/test/MvRxTestRule.kt
```diff
@@ -0,0 +1,68 @@
...
+
+    private fun setRxImmediateSchedulers() {
+        RxJavaPlugins.reset()
+        val immediate = object : Scheduler() {
+            // this prevents StackOverflowErrors when scheduling with a delay
+            override fun scheduleDirect(@NonNull run: Runnable, delay: Long, @NonNull unit: TimeUnit): Disposable = super.scheduleDirect(run, 0, unit)
+
+            override fun createWorker(): Scheduler.Worker = ExecutorScheduler.ExecutorWorker(Executor { it.run() })
+        }
+        RxJavaPlugins.setNewThreadSchedulerHandler { immediate }
```
rom4ek: Shouldn't the similar thing be also done for `RxAndroidPlugins` since it's used in `subscribeLifecycle` function?

gpeal:
@BenSchwab I updated the test rule but I'm having some issues getting it to work reliably. It doesn't seem to perfectly clean itself up. If I run these separately, they pass:
`g :testing:testDebugUnitTest --tests MvRxTestRuleTest*`
`g :testing:testDebugUnitTest --tests MvRxTestRuleSchedulersNotSet`
they pass but running them all fails. Could you take a look?

BenSchwab:
@gpeal AndroidSchedulers#MAIN_THREAD is a static reference which is not being cleaned up by RxPlugin code. I need to think more of how to handle, but nothing great is jumping out to me. I'm a little surprised that this isn't more widely complained about. The way it's set up means that the first `initMainThread` plugin that runs, will win for the rest of the lifecycle of the test suite.

BenSchwab:
Interesting read, seems to be be design: https://medium.com/@peter.tackage/overriding-rxandroid-schedulers-in-rxjava-2-5561b3d14212

BenSchwab:
I think we will need to change some of the asserts in the test

rom4ek:
Also just FYI, when I ran the aforementioned tests with a `trampoline` scheduler instead of the custom one, they worked fine.

BenSchwab:
@rom4ek Could you explain a bit more? We do use a trampoline scheduler for the `initMainThread`. Do you mean for the normal RxSchedulers?

rom4ek:
Yes I meant for the normal Rx schedulers which are being set via RxJavaPlugins

testing/src/main/kotlin/com/airbnb/mvrx/test/MvRxTestRule.kt
```diff
@@ -0,0 +1,83 @@
...
+        override fun createWorker(): Scheduler.Worker = ExecutorScheduler.ExecutorWorker(Executor { it.run() })
+    }
+
+    private var defaultExceptionHandler: Thread.UncaughtExceptionHandler? = null
+
+    override fun apply(base: Statement, description: Description): Statement {
+        return object : Statement() {
+            override fun evaluate() {
+                RxAndroidPlugins.reset()
+                RxAndroidPlugins.setInitMainThreadSchedulerHandler { Schedulers.trampoline() }
```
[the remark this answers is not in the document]
BenSchwab: @gpeal Not a super strong reason. I chose trampoline because the main thread scheduler gets stored as a static variable in AndroidSchedulers, so I thought it was better to store an RxJava scheduler than our own. But, Trampoline and immediate are very similar. Do you recall if there was a reason why we were using our own immediate scheduler? If not, we should consider just switching to all trampolines. I verified the tests still pass with using trampoline everywhere.

testing/src/main/kotlin/com/airbnb/mvrx/test/MvRxTestRule.kt
```diff
@@ -0,0 +1,83 @@
...
+                }
+            }
+        }
+    }
+
+    private fun setRxImmediateSchedulers() {
+        RxJavaPlugins.reset()
+        RxJavaPlugins.setNewThreadSchedulerHandler { immediateScheduler }
+        RxJavaPlugins.setComputationSchedulerHandler { immediateScheduler }
+        RxJavaPlugins.setInitIoSchedulerHandler { immediateScheduler }
```
BenSchwab: Just noticed, but this should probably be just `setIoSchedulerHandler` and not initIo. It probably doesn't have the same leaking issue as mainThread, but for consistency sake.

PR #161, opened 2018-12-31
Make it easier to test state

gpeal:
This adds a `getState` helper to `MvRxTestRule` to make it possible to test state updates without nasty or awkward hacks.

The code to proxy the state is a bit funky to prevent it from getting exposed but that is done to keep the public API cleaner.

@BenSchwab @hellohuanlin @elihart


gpeal:
@rom4ek Yeah, looks like this will simplify some of your tests 😄 How do you like the API?

rom4ek:
@gpeal to be honest, I'm totally fine with the way I did it.. :smile: For me this looks even more readable:
```
withState(viewModel) { // assertions over 'it' }
```
Especially when it comes to several assertions at time. With your current implementation, we'd most likely use Kotlin's standard `with(mvrxRule.getState(viewModel))` operator, which looks pretty similar actually :) Therefore, not sure about the benefits

gpeal:
@rom4ek Actually, I agree with you. Let's go with that as the recommended approach.

PR #163, opened 2019-01-02
update dad joke search function to be async

hellohuanlin:
What's expected: 

If we call `fetchNextPage()` twice in init: 
```
init {
  fetchNextPage()
  fetchNextPage()
}
```
The second call should be short circuited by `is Loading` check

In original implementation, this doesn't work (i.e. `is Loading` check is never true), because the dummy search function emits Success state synchronously. This PR simply defer the execution to make it async. 

@BenSchwab @elihart @gpeal

### DIFF
--- a/sample/src/main/java/com/airbnb/mvrx/sample/features/dadjoke/DadJokeIndexViewModel.kt
+++ b/sample/src/main/java/com/airbnb/mvrx/sample/features/dadjoke/DadJokeIndexViewModel.kt
@@ -12,6 +12,7 @@ import com.airbnb.mvrx.sample.models.Joke
 import com.airbnb.mvrx.sample.models.JokesResponse
 import com.airbnb.mvrx.sample.network.DadJokeService
 import org.koin.android.ext.android.inject
+import java.util.concurrent.TimeUnit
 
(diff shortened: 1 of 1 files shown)


gpeal:
@hellohuanlin I'm not sure I follow. That is actually executing a network request through a retrofit service.

pmecho:
The description under https://github.com/airbnb/MvRx/wiki#accessing-state doesn't seem to match up with how the sample works. The RxJava calls will be made synchronously since we're using the `create()` method when creating the `RxJava2CallAdapterFactory`. It seems that the sample should be using an asynchronous observable to reflect the wiki description. It seems like this should be done either by using the `createAsync()` method when creating the `RxJava2CallAdapterFactory` or by using a `subscribeOn()` call when searching on the `DadJokeService`.

hellohuanlin:
@pmecho yeah `subscribeOn(io)` looks cleaner than delay 1 millisecond

gpeal:
@hellohuanlin Yeah, this is correct now. Execute used to call subscribeOn(Schedulers.io()) automatically but it was never added here when that was removed.

PR #164, opened 2019-01-04
Add select function.

jinxinzheng:
Another option for the `selectSubscribe` methods. It eliminates the use
of reflection and allows to write more concise code.

Sample usage looks like

```
viewModel
    .select { + prop1 + prop2 + prop3 }
    .subscribe{ (p1, p2, p3) -> ... }
```

### DIFF
--- a/mvrx/src/main/kotlin/com/airbnb/mvrx/BaseMvRxViewModel.kt
+++ b/mvrx/src/main/kotlin/com/airbnb/mvrx/BaseMvRxViewModel.kt
@@ -505,6 +505,31 @@ abstract class BaseMvRxViewModel<S : MvRxState>(
             .distinctUntilChanged()
             .subscribeLifecycle(owner, uniqueOnly, subscriber)
 
+    /**
+     * Select state changes for subscription.
+     */
+    protected fun <T> select(selector: S.() -> T) = selectInternal(null, false, selector)
+
+    @RestrictTo(RestrictTo.Scope.LIBRARY)
+    fun <T> select(
+        owner: LifecycleOwner,
+        uniqueOnly: Boolean,
+        selector: S.() -> T
+    ) = selectInternal(owner, uniqueOnly, selector)
+
+    private fun <T> selectInternal(
+        owner: LifecycleOwner?,
+        uniqueOnly: Boolean,
+        selector: S.() -> T
+    ): MvRxSubscriber<T> {
+        val subscriber = DefaultSubscriber<T>()
+        stateStore.observable
+            .map { it.selector() }
+            .distinctUntilChanged()
+            .subscribeLifecycle(owner, uniqueOnly) { subscriber.emit(it) }
+        return subscriber
+    }
+
     private fun <T> Observable<T>.subscribeLifecycle(
         lifecycleOwner: LifecycleOwner? = null,
         uniqueOnly: Boolean,
--- a/mvrx/src/main/kotlin/com/airbnb/mvrx/BaseMvRxViewModel.kt
+++ b/mvrx/src/main/kotlin/com/airbnb/mvrx/BaseMvRxViewModel.kt
@@ -479,6 +479,32 @@ abstract class BaseMvRxViewModel<S : MvRxState>(
             .distinctUntilChanged()
             .subscribeLifecycle(owner, uniqueOnly) { (a, b, c, d, e, f, g) -> subscriber(a, b, c, d, e, f, g) }
 
+    /**
+     * Subscribe to state changes for custom mapping of the state.
+     */
+    protected fun <T> selectSubscribe(
+            selector: S.() -> T,
+            subscriber: (T) -> Unit
+    ) = selectSubscribeInternal(null, selector, false, subscriber)
+
+    @RestrictTo(RestrictTo.Scope.LIBRARY)
+    fun <T> selectSubscribe(
+            owner: LifecycleOwner,
+            selector: S.() -> T,
+            uniqueOnly: Boolean,
+            subscriber: (T) -> Unit
+    ) = selectSubscribeInternal(owner, selector, uniqueOnly, subscriber)
+
+    private fun <T> selectSubscribeInternal(
+            owner: LifecycleOwner?,
+            selector: S.() -> T,
+            uniqueOnly: Boolean,
+            subscriber: (T) -> Unit
+    ) = stateStore.observable
+            .map { it.selector() }
+            .distinctUntilChanged()
+            .subscribeLifecycle(owner, uniqueOnly, subscriber)
+
     private fun <T> Observable<T>.subscribeLifecycle(
         lifecycleOwner: LifecycleOwner? = null,
         uniqueOnly: Boolean,
--- a/mvrx/src/main/kotlin/com/airbnb/mvrx/MvRxState.kt
+++ b/mvrx/src/main/kotlin/com/airbnb/mvrx/MvRxState.kt
@@ -9,4 +9,7 @@ package com.airbnb.mvrx
  * that takes a parcelable object. If the Fragment that initializes this ViewModel has an argument
  * at [MvRx.KEY_ARG], it will be passed to your secondary constructor automatically.
  */
-interface MvRxState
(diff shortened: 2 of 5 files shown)


hellohuanlin:
I like this direction. This is more concise syntax, and allows more complex mapping of state from calling site, without the need to create a getter property in the state. 

Since the new function is a more generalized version of the old `selectSubscribe`s, how about making the old `selectSubscribe`s to directly call this function internally? 

Or we can completely remove old functions, since it's could be confusing to have 2 ways of doing the same thing.

gpeal:
@jinxinzheng @hellohuanlin @BenSchwab Interesting idea. However, it doesn't have the type safety of selectSubscribe so you could so something like:
```
viewModel
    .select { + prop1 + 5 + "foo" }
    .subscribe { (prop1, iAm5, iAmFoo)  -> }
```

This might be fixable with generic bounds though.

One other issue is that while the syntax is clever, the leading `+` feels unconventional and undiscoverable. One of the nice things about MvRx is that it provides some strong opinions on syntax so it is easy to learn and write. However, this creates a totally separate syntax that is clever but less discoverable and potentially confusing.

hellohuanlin:
@gpeal I am not a fan of the `+` sign either. I think this is better: 

```
viewModel
   .selectSubscribe({ it.prop1 }, { prop1 -> ... }
```

```

**Comment by jinxinzheng** [comment]
@gpeal I agree it's not as type safe as selectSubscribe. But it's intended to get arbitrary mapping of the state. Maybe the following syntax is less confusing while still allowing arbitrary mapping,

viewModel.select { tuple(prop1, prop2, prop3) }...

Specifically, if one wants to subscribe to only one property, he/she can remove the use of tuple:

viewModel.select { property }. subscribe.{ it -> ..}

And about the generic bounds, I have no idea currently how could I use it to help improve type safety?

@hellohuanlin the suggested syntax is easy to add. I will add it later. 

I just think the select.subscribe syntax looks more streamlined and more concise. However, I think they could both exist.

**Comment by jinxinzheng** [comment]
I have rebased this and the first commit only adds an overload of the `selectSubscribe` which follows the  suggestion by @hellohuanlin, and the `select` function is rewritten.

It's also possible to implement the `select` as extension function to `MvRxView` and `BaseMvRxViewModel` based on the new `selectSubscribe`.

**Comment by gpeal** [comment]
@jinxinzheng @hellohuanlin I'm still unsure of how this is materially better vs just different. Can you provide some real world examples where this is better?

**Comment by jinxinzheng** [comment]
Firstly I just use it for cleaner syntax. and later I found it might be useful in the following example.

`data class State(val isA: Boolean, val isB: Boolean)`

and you want to react to when either isA or isB is true. With `selectSubscribe(State::isA, State::isB)` you get every notification when either isA or isB changed. With the new syntax `selectSubscribe({isA || isB}) { value -> ...} ` you only get one notification if the mapped value is changed.

**Comment by gpeal** [comment]
@jinxinzheng In that case, the function is more of a mapper than a selector then so then the API would become:
```
viewModel
    .mapState { ... }
    .subscribe { ... }
```
and from inside of a view model:
```
class MyViewModel ... {
    init {
        mapState { ... }
        subscribe { ... }
    }
}
```

Thinking of it as a mapper makes it seem more reasonable to me. @BenSchwab @elihart what do you think?

**Comment by elihart** [comment]
I think the perspective of a mapper is reasonable

**Comment by jinxinzheng** [comment]
That makes sense. I can do this. I will also delete the `+` syntaxes since they are confusing.

Which of the following is better?

```
mapState{..}
.subscribe{..}
```

Or

```
mappedSubscribe(
    mapper = {..},
    subscriber = {..}
)
```

**Comment by jinxinzheng** [comment]
@gpeal 
> `mapState` multiple times 

So you mean like this: `mapState{ .. }.mapState{ .. }.subscribe {..}` ?

Then that would require some more works.

**Comment by hellohuanlin** [comment]
(1) @gpeal yes that's pretty much the old `selectSubscribe` API. We later decided to favor KProperty and removed that API. I think it's good to have it back, though my concern is having both APIs doing similar things may lead to confusion and more steep learning curve. 

That being said, if we were to have both API, I'd prefer to have the existing `selectSubscribe(kProps)` to internally call this new function, since it's more generalized version of existing `selectSubscribe`. 

(2) About `mapState{..}.subscribe{..}` vs `selectSubscribe()`
My guess is that most of the time developers are going to use exactly 1 `mapState` and immediately call `subscribe()` after that. If that assumption is true, `mappedSubscribe()` may be easier to use.

**Comment by gpeal** [comment]
@jinxinzheng Fair but maybe that isn't so important after all.

Also, is this API something you have encountered while working with MvRx `selectSubscribe({isA || isB}) { value -> ...}` makes sense on its own but what's a real world use case for it?

@BenSchwab @elihart @hellohuanlin Can you think of any cases from Airbnb in which this would have helped?

**Comment by jinxinzheng** [comment]
I'm not aware of that old `selectSubscribe` API. If you have decided to remove it for some decent reason, maybe it's really not so important to add it back.

However, for the real world use case. I can provide one of my own case here.

For a state with two async requests

```kotlin
data class State(
    val request1: Async<Foo>,
    val request2: Async<Bar>
)
```

We want to show **one and only one** toast when either of the two requests fails, or both of them succeeds. Something like
```kotlin
viewModel.selectSubscribe(State::request1, State::request2) {r1, r2 ->
    if (r1 is Fail || r2 is Fail) {
        toast("failed")
    } else if (r1 is Success && r2 is Success) {
        toast("success")
    }
} 
```
will show two failed toasts if the two requests failed one after the other.

This could be fixed by 
```kotlin
viewModel.mapState { request1 is Fail || request2 is Fail }
    .subscribe { toast("failed") }
viewModel.mapState { request1 is Success && request2 is Success }
    .subscribe { toast("success") }
```

while the code is still clean and readable.

gpeal:
@jinxinzheng In that case, it could be the same number of lines of code if you do:
```
viewModel.selectSubscribe(State::request1, State::request2) { r1, r2 ->
    if (r1 is Fail || r2 is Fail)  toast("failed")
    else if (r1 is Success && r2 is Success)  toast("success")
} 
```
and you have the added bonus of still having access to the raw request async objects that you could and likely would want to use in most cases.

jinxinzheng:
@gpeal It's not about the number of lines of code. The problem is that there will be two "Failed" toasts if both requests fails, which is not desired.

gpeal:
@jinxinzheng You could do something like [this](https://github.com/airbnb/MvRx/issues/149#issuecomment-451837588)

gpeal:
@jinxinzheng I'm not strongly opposed to it but I want to avoid adding APIs unless they have a strong use case to keep MvRx small and easy to learn.

@BenSchwab @elihart @hellohuanlin care to chime in here?

hellohuanlin:
@gpeal as @jinxinzheng pointed out, the difference is just declaration site or calling site. I think it's more readable and maintainable to be on calling site, to have the mapper code to be closer to subscriber code. 

However, it does introduce steeper learning curve. I am wondering are you strongly against removing the existing API with `KProperty`? Because most of the cases, we only subscribe to 1 property, and this new API syntax is more concise, without having to specify the scope `StateClassName::`.

gpeal:
@hellohuanlin It would make the multiple `selectSubscribe` case more awkward if we remove the API though

BenSchwab:
if I recall the rational against including `map` was we can easily emulate map by creating a derived property on the state class, which I think @gpeal is mentioning [here](https://github.com/airbnb/MvRx/issues/149#issuecomment-451837588). 

I think that is more compelling than mapping because state class are easily testable. Subscribe sites are often in Fragments, which are not as easily testable. My bias for MvRx is to push as much business logic to State/ViewModels, keeping fragments/views as purely reactive as possible.

gpeal:
@BenSchwab Ah yes, that was the reason!

hellohuanlin:
@BenSchwab yeah that's a solid reason.

gpeal:
@BenSchwab @hellohuanlin @jinxinzheng Should we close this then?

jinxinzheng:
OK

PR #169, opened 2019-01-10
Add initialState to MvRxFactory, give access to args in create ViewModel

BenSchwab:
Closes https://github.com/airbnb/MvRx/issues/83

Motivation:

1) If you need to use a value from args only in the viewModel (say an id to call an API) the only way to access in the view model is to manually find the arguments (painful) or add a "dead" field to state class.

2) If you need to access information to create `initialState` that is not stored in args, but say in your dagger component, there is no way to access it via a secondary constructor.

## Breaking changes

```
MvRxFactory<State>#create(activity: Activity, state: State) : BaseMvRxViewModel<State> ->
MvRxFactory<State, ViewModel>#create(viewModelContext: ViewModelContext, state: State) : ViewModel
```
Note, activity is now referenceable as `viewModelContext.activity.`

The second parameter was added to `MvRxFactory` because it guarantees that you cannot accidentally return a ViewModel of the wrong type in `create`.

- (unreleased) Removes `MvRxFragmentViewModelFactory` https://github.com/airbnb/MvRx/pull/148. I prefer the single factory in this PR, as if you used a `MvRxFragmentViewModelFactory` but scoped your ViewModel to an activity, there will be a crash. This new approach forces that possibility to be acknowledged in code.

## New
- Factory now has a method `initialState(viewModelContext: ViewModelContext)`. This gives access to the args _and_ the activity/fragment before creating the state. You can now use dagger to inject dependencies into your state. The existing constructor based state creation still works, however, it has lower precedence that `initialState` in the factory. If both are defined, the initialState from the factory will be used.
- No more requirement for `@JvmStatic`. I wrote some java reflection to access the Companion object. It seems to be working fine, and should make onboarding a lot easier, as forgetting `@JvmStatic` is very common.

### Open questions
- Does anyone have a better idea for a name than `ViewModelContext`. I hate overloading the word context, but it feels like that is what the combination of the owner and args is.
- Any typed args. In the old manner, args would be automatically cast to the type of a matching secondary constructor. I couldn't figure out a clean way to get this to work without forcing on generic on `ViewModelFactory`. As many times you will use a ViewModelFactory without needing args, it felt bad to force many declarations to have a meaningless `<Any>`. Also the Factory has two generics, a third is a lot... 
Currently, the context has `rawArgs` which is just the untype arg object, and `fun <T> args()` a convenience method to type args. 
Note: it is totally possible, and expected to be able to handle many types of args, which can be done cleaning with a `when` on `rawArgs`.

@gpeal @hellohuanlin @elihart


mvrx/src/test/kotlin/com/airbnb/mvrx/MvRxExtensionsTest.kt
```diff
@@ -1,28 +0,0 @@
...
-  @Parcelize
-  data class Arg(val value: String) : Parcelable
-
-  class DemoState1(val arg: Parcelable) : MvRxState
-
-  class DemoState2(val arg: Arg) : MvRxState
-
-  @Test
-  fun testInitialDemoState1() {
-    val state = _initialStateProvider(DemoState1::class.java, Arg("Hello"))
```
BenSchwab: These providers are gone. Lot's of other tests cover this code too.

mvrx/src/main/kotlin/com/airbnb/mvrx/MvRxExtensions.kt
```diff
@@ -48,91 +47,37 @@ inline fun <T, reified VM : BaseMvRxViewModel<S>, reified S : MvRxState> T.activ
...
 
 /**
  * For internal use only. Public for inline.
  *
- * Looks for [MvRx.KEY_ARG] on the arguments of the fragment receiver to create an instance of the State class.
- * The state class must have a matching single arg constructor.
+ * Looks for [MvRx.KEY_ARG] on the arguments of the fragments.
+ */
+@Suppress("FunctionName")
+fun <T : Fragment> T._fragmentArgsProvider(): Any? = arguments?.get(MvRx.KEY_ARG)
```
gpeal: @BenSchwab `@RestrictTo(Library)`

mvrx/src/main/kotlin/com/airbnb/mvrx/MvRxExtensions.kt
```diff
@@ -48,91 +47,37 @@ inline fun <T, reified VM : BaseMvRxViewModel<S>, reified S : MvRxState> T.activ
...
  *
- * If no MvRx fragment args exist it attempts to use a empty constructor. Otherwise an exception is thrown.
+ * Looks for [MvRx.KEY_ARG] on the arguments of the fragment receiver.
  *
  * Also adds the fragment's MvRx args to the host Activity's [MvRxViewModelStore] so that they can be used to recreate initial state
  * in a new process.
  */
 @Suppress("FunctionName")
-inline fun <reified S : MvRxState, T : Fragment> T._activityViewModelInitialStateProvider(keyFactory: () -> String): S {
+inline fun <T : Fragment> T._activityArgsProvider(keyFactory: () -> String): Any? {
```
gpeal: `@RestrictTo(Library)` this too

elihart: the idea of "activity args" is confusing to me since the args are still coming from a fragment, and the function receiver is a fragment

mvrx/src/main/kotlin/com/airbnb/mvrx/MvRxExtensions.kt
```diff
@@ -48,91 +47,37 @@ inline fun <T, reified VM : BaseMvRxViewModel<S>, reified S : MvRxState> T.activ
...
     val args: Any? = arguments?.get(MvRx.KEY_ARG)
```
elihart: could this chain to `_fragmentArgsProvider` instead?

mvrx/src/main/kotlin/com/airbnb/mvrx/MvRxViewModelFactory.kt
```diff
@@ -4,26 +4,84 @@ import android.support.v4.app.Fragment
...
+     */
+    abstract val rawArgs: Any?
+
     /**
-     * This will be called when your ViewModel needs to created. This *needs* to be annotated with [JvmStatic].
-     * @param state: The initial state for your ViewModel. This will be populated from fragment args and persisted state.
+     * Convenience method to type [rawArgs].
      */
-    fun create(fragment: Fragment, state: S): BaseMvRxViewModel<S>
+    abstract fun <A> args(): A
```
elihart: is the implementation the same both subclasses? can we not have a default implementation?

mvrx/src/main/kotlin/com/airbnb/mvrx/MvRxViewModelFactory.kt
```diff
@@ -4,26 +4,84 @@ import android.support.v4.app.Fragment
...
+}
+
+/**
+ * The [ViewModelContext] for a ViewModel created with an
+ * activity scope (`val viewModel by activityViewModel<MyViewModel>`).
+ */
+class ActivityViewModelContext(
```
elihart: It's not immediately clear to me why we need to separate fragment and activity contexts, since an activity scoped viewmodel is still initially created for a specific fragment, and with arguments from that fragment.

I suppose it's just during process restart that we don't have the original fragment anymore? That's subtlety makes this requirement/distinction not immediately clear

mvrx/src/main/kotlin/com/airbnb/mvrx/MvRxViewModelProvider.kt
```diff
@@ -156,4 +113,38 @@ object MvRxViewModelProvider {
         }
         return null
     }
+
+    /**
+     * For internal use only. Public for inline.
```
elihart: you can make this `internal` and use `@PublishedApi` instead of needing to make it public

mvrx/src/main/kotlin/com/airbnb/mvrx/MvRxViewModelProvider.kt
```diff
@@ -156,4 +113,38 @@ object MvRxViewModelProvider {
...
+            ?: try {
+                stateClass.newInstance()
+            } catch (@Suppress("TooGenericExceptionCaught") e: Throwable) {
+                null
+            }
+            ?: throw IllegalStateException(
+                "Attempt to create the MvRx state class ${stateClass.simpleName} has failed. One of the following must be true:" +
+                    "\n 1) The state class has default values for every property." +
+                    "\n 2) The state class has a secondary constructor for ${args?.javaClass?.simpleName ?: "a fragment argument"}." +
+                    "\n 3) The ViewModel using the state must has a companion object implementing MvRxFactory with an initialState function " +
```
elihart: nit: `must has`

mvrx/src/main/kotlin/com/airbnb/mvrx/MvRxViewModelFactory.kt
```diff
@@ -4,26 +4,84 @@ import android.support.v4.app.Fragment
...
-     * @param state: The initial state for your ViewModel. This will be populated from activity persisted state.
+     * @param viewModelContext [ViewModelContext] which contains the ViewModel owner and arguments.
+     * @param state The initial state to pass to the ViewModel. In a new process, state will have all [PersistState] annotated members restored,
+     * therefore you should never create a custom state in this method. To customize the initial state, override [initialState].
+     *
+     * @return The ViewModel. If you return `null` the ViewModel must have a single argument
+     * constructor only taking the initial state.
      */
-    fun create(activity: FragmentActivity, state: S): BaseMvRxViewModel<S>
+    fun create(viewModelContext: ViewModelContext, state: S): VM? = null
```
marukami: The issue I ran into with having a single create in #148 was casting. If the consumer wants the `FragmentViewModelContext` then you need to downcast the `ViewModelContext`. 

For me, cause Dagger injects the `ViewModel` Factory implementation into the fragment, I would need to downcast for virtually all my `ViewModel`s factories to use the `FragmentViewModelContext`.

mvrx/src/test/kotlin/com/airbnb/mvrx/FactoryTest.kt
```diff
@@ -35,114 +32,246 @@ class FactoryTest : BaseTest() {
...
+    }
+}
+
+/**
+ * Test a factory which only uses a custom ViewModel create.
+ */
+class FactoryViewModelTest : BaseTest() {
+
+    private class TestFactoryViewModel(initialState: FactoryState, val otherProp: Long) : TestMvRxViewModel<FactoryState>(initialState) {
+        companion object : MvRxViewModelFactory<TestFactoryViewModel, FactoryState> {
```
marukami: Should `TestFactoryViewModel` be using `FragmentViewModelContext`?

mvrx/src/main/kotlin/com/airbnb/mvrx/MvRxViewModelFactory.kt
BenSchwab: @elihart Yes, during process activity scoped ViewModels will be recreated when the activity is created. So even in a normal creation sequence we will pass it only the activity (even though a fragment reference is technically accessible).

mvrx/src/main/kotlin/com/airbnb/mvrx/MvRxExtensions.kt
BenSchwab: Agreed, let me explain the reasons a bit more in the comments.

mvrx/src/main/kotlin/com/airbnb/mvrx/MvRxViewModelFactory.kt
BenSchwab: @marukami Yeah, that is a downside, but I don't see a safe way around this. It seems too risky to have a ViewModelFactory that will crash if you use an activity scope.

I think, if in your application you know that you will only use fragment scoped view models, you could do something like:

```
interface MvRxViewModelFragmentFactory<VM: BaseMvRxViewModel<S>, S : MvRxState> : MvRxViewModelFactory<VM, S> {
    
    fun createFragmentScopedViewModel(viewModelContext: FragmentViewModelContext, state: S) : VM?
    
    override fun create(viewModelContext: ViewModelContext, state: S) = createFragmentScopedViewModel(viewModelContext as FragmentViewModelContext, state)

    override fun initialState(viewModelContext: ViewModelContext) : S? = initialStateFragmentScoped(viewModelContext as FragmentViewModelContext)

    fun initialStateFragmentScoped(viewModelContext: FragmentViewModelContext) : S? 
}
```

That way you are explicitly performing the unsafe cast. Does this seem reasonable?

mvrx/src/main/kotlin/com/airbnb/mvrx/MvRxViewModelFactory.kt
```diff
@@ -4,26 +4,76 @@ import android.support.v4.app.Fragment
...
+) : ViewModelContext()
+
+/**
+ * The [ViewModelContext] for a ViewModel created with a
+ * fragment scope (`val viewModel by fragmentViewModel<MyViewModel>`).
+ */
+class FragmentViewModelContext(
+    override val activity: FragmentActivity,
+    override val rawArgs: Any?,
+    val fragment: Fragment
```
gpeal: @BenSchwab You could make a `<T : Fragment> fragment()` and `<A : Activity> activity()` function that could help with casting to your own app's Activity/Fragment

mvrx/src/main/kotlin/com/airbnb/mvrx/MvRxViewModelFactory.kt
marukami: With the changes @gpeal suggestion for the Fragment been a generic, I think an extension function like this should work nicely. 

```kotlin
@Suppress("UNCHECKED_CAST")
inline fun <F: Fragment> ViewModelContext.withFragmentFactory(factory: (F) -> ViewModel) =
        (this as FragmentViewModelContext).let { it.fragment as F }.let(factory)
```

PR #187, opened 2019-01-19
Small cleanup to get rid of Kotlin nullity warning

tasomaniac:
Cleanup restoreViewModels to get rid of Kotlin nullity warning.

Here was the warning. This happens because `getParcelable` returns Nullable.

```
w: MvRxViewModelStore.kt: (107, 68): Type mismatch: inferred type is MvRxPersistedViewModelHolder? but MvRxPersistedViewModelHolder was expected
w: MvRxViewModelStore.kt: (108, 67): Type mismatch: inferred type is MvRxPersistedViewModelHolder? but MvRxPersistedViewModelHolder was expected
```

Additionally, used when statement instead of optional casting to `Fragment` and then force casting to `FragmentActivty` if casting to `Fragment` was not successful. Now, the type checking is more clear. Everything is smart-casted.

### DIFF
--- a/mvrx/src/main/kotlin/com/airbnb/mvrx/MvRxViewModelStore.kt
+++ b/mvrx/src/main/kotlin/com/airbnb/mvrx/MvRxViewModelStore.kt
@@ -1,4 +1,3 @@
-
 package com.airbnb.mvrx
 
 import android.arch.lifecycle.ViewModel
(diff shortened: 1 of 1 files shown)


mvrx/src/main/kotlin/com/airbnb/mvrx/MvRxViewModelStore.kt
```diff
@@ -103,9 +103,12 @@ class MvRxViewModelStore(private val viewModelStore: ViewModelStore) {
             } else {
                 ownerArgs
             }
-            map[it] = (host as? Fragment)
-                    ?.let { fragment -> restoreViewModel(fragment, viewModelsState.getParcelable(it), arguments) }
-                    ?: restoreViewModel(host as FragmentActivity, viewModelsState.getParcelable(it), arguments)
+            val holder = viewModelsState.getParcelable<MvRxPersistedViewModelHolder>(it)!!
+            map[it] = when(host) {
```
gpeal: nit: missing a space before after `when`

mvrx/src/main/kotlin/com/airbnb/mvrx/MvRxViewModelStore.kt
```diff
@@ -103,9 +103,12 @@ class MvRxViewModelStore(private val viewModelStore: ViewModelStore) {
             } else {
                 ownerArgs
             }
-            map[it] = (host as? Fragment)
-                    ?.let { fragment -> restoreViewModel(fragment, viewModelsState.getParcelable(it), arguments) }
-                    ?: restoreViewModel(host as FragmentActivity, viewModelsState.getParcelable(it), arguments)
+            val holder = viewModelsState.getParcelable<MvRxPersistedViewModelHolder>(it)!!
```
gpeal: Since a null value will crash anyway, can you give it a good error message with `?: throw IllegalStateException(...)`?

tasomaniac: Makes sense, will do.

mvrx/src/main/kotlin/com/airbnb/mvrx/MvRxViewModelStore.kt
tasomaniac: Are you sure about this?

tasomaniac:
@gpeal thanks for the review. I think this is now good to go. 

Also gave a named param to lambda instead of using `it`.

PR #195, opened 2019-02-12
Delete unused example test

tasomaniac:
These tests are not used and can be confusing to the developers browsing the repo. 

I was looking for sample unit tests for MvRx for example.


gpeal:
Thanks!

Issue #190, opened 2019-01-21
Remove the default value for debugMode in BaseMvRxViewModel


gpeal:
As it is, it is easy for somebody to make the mistake of using BaseMvRxVIewModel without the debugMode param set so it will always be false.

PR #196, opened 2019-02-12
Require debug flag

marukami:
Updated `BaseMvRxViewModel` to now require the debug flag #190

Had to fix formatting reported by Detekt before was able to push with git hooks enabled.

Update `Readme.me` with
* To have the most recent release version
* Added debug flag using `BuildConfig` as a base guideline.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,8 @@
 # Change log
 
+## Version 0.8.0 (Current development)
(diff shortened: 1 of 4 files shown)


gpeal:
@marukami Thanks! Can you add something to the changelog indicating that this *may* be a breaking change if you weren't already using debug mode?

PR #197, opened 2019-02-12
 Allow for injectable ViewModel factories

marukami:
Instead of accessing an injected ViewModel Factory via the companion object context I've since found it easier to just sidestep the reflection and directly inject the ViewModel Factory into the ViewModelProvider.

This makes it even easier when using dagger #69 cause you can inject the Factory like this

```kotlin
@Inject lateinit var viewModelFactory: UserDetailViewModel.Factory

private val userDetailViewModel: UserDetailViewModel by fragmentViewModel {
    viewModelFactory.create(arguments!!.getLong("USER_ID"))
}
```

Feedback on adding this approch would be appreciated.

### DIFF
--- a/mvrx/src/main/kotlin/com/airbnb/mvrx/MvRxExtensions.kt
+++ b/mvrx/src/main/kotlin/com/airbnb/mvrx/MvRxExtensions.kt
@@ -23,9 +23,10 @@ import kotlin.reflect.KProperty
  */
 inline fun <T, reified VM : BaseMvRxViewModel<S>, reified S : MvRxState> T.fragmentViewModel(
     viewModelClass: KClass<VM> = VM::class,
-    crossinline keyFactory: () -> String = { viewModelClass.java.name }
+    crossinline keyFactory: () -> String = { viewModelClass.java.name },
+    noinline viewModelFactory: MvRxViewModelProviderFactory<VM>? = null
 ) where T : Fragment, T : MvRxView = lifecycleAwareLazy(this) {
-    MvRxViewModelProvider.get(viewModelClass.java, S::class.java, FragmentViewModelContext(this.requireActivity(), _fragmentArgsProvider(), this), keyFactory())
+    MvRxViewModelProvider.get(viewModelClass.java, S::class.java, FragmentViewModelContext(this.requireActivity(), _fragmentArgsProvider(), this), keyFactory(), viewModelFactory)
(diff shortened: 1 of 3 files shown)


gpeal:
@marukami We remove the ability to do this for activity view models intentionally.
My recollection is that it was for this case:
1) Create this Fragment in Fragment A
2) Navigate to Fragment B that accesses the view model created in A with `existingViewModel()`
3) Disable background processes in developer settings
4) Return to your app. MvRx needs to be able to recreate your ViewModel without that code running in Fragment A.

gpeal:
cc @BenSchwab

marukami:
Ah, Yes good point. 

It's late here but I did a quick hack that might work. I'll try and see how I can make a test to validate this next version would work.
https://github.com/airbnb/MvRx/compare/master...marukami:injectable-factory-interface?expand=1

gpeal:
@marukami Another challenge is that you can only implement an interface once per class. You can't overload an interface multiple times with different generic bounds. That means that you would only be able to use this with a single ViewModel class per Fragment/Activity.

What is the advantage of this vs having it tied to your ViewModel companion object. You have a reference to your Activity/Fragment so why does the code have to actually run there?

marukami:
> What is the advantage of this vs having it tied to your ViewModel companion object.

When using Dagger, it's a lot simpler to inject the Fragment/Activity with the factory implementation. Then you need to downcast for a Fragment and downcast again to get access to the injected factory. 
```kotlin
// This Ext fun helps a little 
ViewModelContext.withFragmentFactory(factory: (F, A) -> VM) =
  (this as FragmentViewModelContext).let { factory(it.fragment as F, it.args as A) }

viewModelContext
        .withFragmentFactory { fragment: RegisterPasscodeFragment, args: RegisterPasscodeArgs ->
          fragment.viewModelFactory.create(state, args)
        }
```

The main advantage of the dagger use case is the compile time validation.

> Another challenge is that you can only implement an interface once per class.

Ah, Good point. Right now for legacy reasons I have a `ViewModelProvider.Factory` as a Singleton. It has the same issue. The way we resolved the issue was to inject a map of class -> ViewModelFactories. Doing that feels like I'm adding more than I want from having the ability to inject a compile-time factory.

@gpeal Thank you so much for all the feedback. I will close this PR for now since I need to resolve a few more issues and validate all the edge cases are handled. Thank you again.

gpeal:
Thanks for the ideas @marukami. It's always healthy to consider new ideas.

PR #205, opened 2019-03-10
Add a test that ensures that we fail in debug mode if state is an optional param

### LINKED ISSUE #95: ViewModelFactory error uses wrong class name
```kotlin
viewModelClass.kotlin.primaryConstructor?.parameters?.get(0)?.type != state::class -> {
                    "${MvRxViewModelFactory::class.java.simpleName} must have primary constructor with a " +
                        "single parameter that takes initial state of ${state::class.java.simpleName}. Found type " +
                        "${viewModelClass.kotlin.primaryConstructor?.parameters?.get(0)?.type?.javaClass?.simpleName}"
                }
```

this is using the viewmodelfactory class in the error message instead of the view model class

gpeal:
Fixes #95

### DIFF
--- a/mvrx/src/main/kotlin/com/airbnb/mvrx/MvRxViewModelProvider.kt
+++ b/mvrx/src/main/kotlin/com/airbnb/mvrx/MvRxViewModelProvider.kt
@@ -64,8 +64,8 @@ object MvRxViewModelProvider {
                     "It must have companion object implementing ${MvRxViewModelFactory::class.java.simpleName} " +
                     "with a create method returning a non-null ViewModel."
             } else {
-                "${viewModelClass.simpleName} must have primary constructor with a " +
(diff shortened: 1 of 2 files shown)


mvrx/src/main/kotlin/com/airbnb/mvrx/MvRxViewModelProvider.kt
```diff
@@ -65,7 +65,7 @@ object MvRxViewModelProvider {
                     "with a create method returning a non-null ViewModel."
             } else {
                 "${MvRxViewModelFactory::class.java.simpleName} must have primary constructor with a " +
```
elihart: I think the main point of the issue was that `${MvRxViewModelFactory::class.java.simpleName} ` is used, which is not the right class,

`viewModelClass` . should be used instead

gpeal:
@elihart Oh, you're right. Fixed.

PR #206, opened 2019-03-10
Remove coalescing

### LINKED ISSUE #142: Do not collapse state change notifications
I think it's more expected behavior to have 1 reducer to trigger 1 state change. Collapsing notifications can result in missing changes. 

For example, if we have value changes from `0 -> 1 -> 0`, after collapsing, that state change information is gone. 

I agree that collapsing is more performant since it reduces unnecessary UI rendering. But it's possible that developers subscribe to state not just for UI purpose. 

Maybe we can allow a customization on whether to collapse or not. 

@BenSchwab @elihart @gpeal

gpeal:
Fixes #142

### DIFF
--- a/mvrx/src/main/kotlin/com/airbnb/mvrx/RealMvRxStateStore.kt
+++ b/mvrx/src/main/kotlin/com/airbnb/mvrx/RealMvRxStateStore.kt
@@ -95,17 +95,12 @@ class RealMvRxStateStore<S : Any>(initialState: S) : MvRxStateStore<S> {
 
         @Synchronized
         fun dequeueGetStateBlock(): ((state: S) -> Unit)? {
-            if (getStateQueue.isEmpty()) return null
-
-            return getStateQueue.removeFirst()
+            return getStateQueue.poll();
         }
 
         @Synchronized
         fun dequeueSetStateBlock(): (S.() -> S)? {
-            // do not allocate empty queue for no-op flushes
-            if (setStateQueue.isEmpty()) return null
-
-            return setStateQueue.removeFirst()
+            return setStateQueue.poll()
         }
     }
 
@@ -124,9 +119,6 @@ class RealMvRxStateStore<S : Any>(initialState: S) : MvRxStateStore<S> {
         flushQueues()
     }
 
-    /**
-     * Coalesce all updates on the setState queue and clear the queue.
-     */
     private fun flushSetStateQueue() {
         val block = jobs.dequeueSetStateBlock() ?: return
         val newState = block(state)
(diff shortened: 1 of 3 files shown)


mvrx/src/main/kotlin/com/airbnb/mvrx/RealMvRxStateStore.kt
```diff
@@ -101,13 +101,11 @@ class RealMvRxStateStore<S : Any>(initialState: S) : MvRxStateStore<S> {
         }
 
         @Synchronized
-        fun dequeueAllSetStateBlocks(): List<(S.() -> S)>? {
+        fun dequeueSetStateBlock(): (S.() -> S)? {
             // do not allocate empty queue for no-op flushes
```
elihart: I don't think this comment applies anymore.

instead of doing the empty check, can you just `setStateQueue.poll()`?

mvrx/src/main/kotlin/com/airbnb/mvrx/RealMvRxStateStore.kt
```diff
@@ -130,11 +128,11 @@ class RealMvRxStateStore<S : Any>(initialState: S) : MvRxStateStore<S> {
      * Coalesce all updates on the setState queue and clear the queue.
```
elihart: comment `Coalesce all updates ` doesn't apply anymore?

Might be worth having a comment documenting why we chose to not coalesce

mvrx/src/main/kotlin/com/airbnb/mvrx/RealMvRxStateStore.kt
```diff
@@ -126,15 +119,12 @@ class RealMvRxStateStore<S : Any>(initialState: S) : MvRxStateStore<S> {
...
-        val blocks = jobs.dequeueAllSetStateBlocks() ?: return
-
-        blocks
-            .fold(state) { state, reducer -> state.reducer() }
-            .run { if (this != state) subject.onNext(this) }
+        val block = jobs.dequeueSetStateBlock() ?: return
+        val newState = block(state)
+        if (newState != state) {
+            subject.onNext(newState)
+        }
```
hellohuanlin: @gpeal @BenSchwab @elihart I think we still need to flush the whole `setStateQueue` here to prioritize reducers over accessors, to avoid a race condition we discussed a while back. 

To avoid coalescing the change, I think we can simply emit the new value once for every reducers in the queue.

```
private fun flushSetStateQueue() {
	        val blocks = jobs.dequeueAllSetStateBlocks() ?: return
                for (block in blocks) {
                    val newState = state.reducer()
                    if (newState != state) { 
                          subject.onNext(newState) 
                    }
                }
```

hellohuanlin: Let me put up a PR today and also add some comment on this race condition.

PR #207, opened 2019-03-11
Respect unique only on configuration changes.

BenSchwab:
closes https://github.com/airbnb/MvRx/issues/204

Currently there is a bug where `uniqueOnly` will redeliver the last value, even if unchanged, during an orientation change. The intent of `uniqueOnly` was to emit a value only once, ever, to allow for actions that need to be taken only once, such as showing a PopTart or logging. (Just for posterity sake I want to note I recommend logging in ViewModel, as subscriptions there have a much simpler lifecycle).

In order to dedup a repeated value after a configuration change, the view model maintains a map between "subscriptionIds" and the last delivered value.

When you choose to subscribe with `uniqueOnly`, you must provide a subscription id, so the view model can perform the dedupping logic. An extension function `uniqueOnly()` on MvRxView will handle creating and persisting a unique subscription id across configuration changes.

@gpeal @elihart

### DIFF
--- a/mvrx/src/main/kotlin/com/airbnb/mvrx/BaseMvRxViewModel.kt
+++ b/mvrx/src/main/kotlin/com/airbnb/mvrx/BaseMvRxViewModel.kt
@@ -13,6 +13,7 @@ import io.reactivex.disposables.CompositeDisposable
 import io.reactivex.disposables.Disposable
 import io.reactivex.functions.Consumer
 import io.reactivex.schedulers.Schedulers
+import java.util.concurrent.ConcurrentHashMap
 import kotlin.reflect.KProperty1
 import kotlin.reflect.KVisibility
 import kotlin.reflect.full.declaredMemberProperties
@@ -35,6 +36,7 @@ abstract class BaseMvRxViewModel<S : MvRxState>(
     private val tag by lazy { javaClass.simpleName }
     private val disposables = CompositeDisposable()
     private lateinit var mutableStateChecker: MutableStateChecker<S>
+    private val lastDeliveredStates = ConcurrentHashMap<String, Any>()
 
     init {
         // Kotlin reflection has a large overhead the first time you run it
@@ -211,11 +213,11 @@ abstract class BaseMvRxViewModel<S : MvRxState>(
      * For ViewModels that want to subscribe to itself.
      */
     protected fun subscribe(subscriber: (S) -> Unit) =
-        stateStore.observable.subscribeLifecycle(null, false, subscriber)
+        stateStore.observable.subscribeLifecycle(null, Standard, subscriber)
 
     @RestrictTo(RestrictTo.Scope.LIBRARY)
-    fun subscribe(owner: LifecycleOwner, uniqueOnly: Boolean = false, subscriber: (S) -> Unit) =
-        stateStore.observable.subscribeLifecycle(owner, uniqueOnly, subscriber)
+    fun subscribe(owner: LifecycleOwner, deliveryMode: DeliveryMode = Standard, subscriber: (S) -> Unit) =
+        stateStore.observable.subscribeLifecycle(owner, deliveryMode, subscriber)
 
     /**
      * Subscribe to state changes for only a single property.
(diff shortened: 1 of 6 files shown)


mvrx/src/main/kotlin/com/airbnb/mvrx/BaseMvRxViewModel.kt
```diff
@@ -35,6 +36,7 @@ abstract class BaseMvRxViewModel<S : MvRxState>(
     private val tag by lazy { javaClass.simpleName }
     private val disposables = CompositeDisposable()
     private lateinit var mutableStateChecker: MutableStateChecker<S>
+    private val lastDeliveredStates = ConcurrentHashMap<String, Any>()
```
elihart: I don't think there is any check that the same id isn't used twice in separate subscriptions. The potential bug where different subscriptions with the same properties leads to collision worries me, especially since it doesn't seem like it would be an obvious issue.

Could we store a map of Id to Disposable here, and it is only valid to register the same id if the disposable is disposed?

mvrx/src/main/kotlin/com/airbnb/mvrx/MvRxLifecycleAwareObserver.kt
```diff
@@ -105,7 +116,12 @@ internal class MvRxLifecycleAwareObserver<T>(
             return
         }
         if (!isDisposed) {
-            val valueToDeliverOnUnlock = if (alwaysDeliverLastValueWhenUnlocked && lastValue != null) lastValue else lastUndeliveredValue
+            val valueToDeliverOnUnlock = when {
+                deliveryMode is UniqueOnly -> lastUndeliveredValue
+                deliveryMode is Standard && lastUndeliveredValue != null -> lastUndeliveredValue
+                deliveryMode is Standard && lastUndeliveredValue == null -> lastValue
```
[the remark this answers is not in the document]
BenSchwab: We always wrap properties in tuples, so they can never be null. I updated the generics to reflect this (RxJava doesn't allow null anyway)

mvrx/src/main/kotlin/com/airbnb/mvrx/BaseMvRxViewModel.kt
BenSchwab: Good idea! I'll add this, but this makes it very important for us to fix the duplicate subscriptions if you subscribe outside of onCreate.

I have a fix in progress for that.

mvrx/src/main/kotlin/com/airbnb/mvrx/MvRxLifecycleAwareObserver.kt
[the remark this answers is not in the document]
elihart: ah, great. In that case it seems the `deliveredFirstValue` atomic boolean is unnecessary because it will always be true when `lastDeliveredValueFromPriorObserver` is non null (I think), so checking both seems redundant.

it is maybe more clear though

mvrx/src/main/kotlin/com/airbnb/mvrx/MvRxView.kt
```diff
@@ -23,6 +23,15 @@ private val handler = Handler(Looper.getMainLooper(), Handler.Callback { message
  * will automatically subscribe to all state changes in the ViewModel and call [invalidate].
  */
 interface MvRxView : MvRxViewModelStoreOwner, LifecycleOwner {
+
+    /**
+     * A globally unique id for this MvRxView. If your MvRxView is being recreated due to a lifecycle event (e.g. rotation)
+     * you should assign a consistent id. Likely this means you should save the id in onSaveInstance state.
+     * The viewId will not be accessed until a subscribe method is called. Accessing mvrxViewId before calling
+     * super.onCreate() will cause a crash.
```
gpeal: You might want to give an example implementation of this for people who are only looking at this and are implementing MvRxView themself.

mvrx/src/main/kotlin/com/airbnb/mvrx/BaseMvRxViewModel.kt
```diff
@@ -523,3 +550,37 @@ abstract class BaseMvRxViewModel<S : MvRxState>(
...
+
+/**
+ * Defines what updates a subscription should receive.
+ * See: [Standard], [UniqueOnly].
+ */
+sealed class DeliveryMode {
+
+    internal fun appendPropertiesToId(vararg properties: KProperty1<*, *>) : DeliveryMode {
+        return when (this) {
+            is Standard -> Standard
```
gpeal: How about something more explicit like `RedeliverOnStart`?

mvrx/src/main/kotlin/com/airbnb/mvrx/BaseMvRxViewModel.kt
```diff
@@ -506,10 +513,30 @@ abstract class BaseMvRxViewModel<S : MvRxState>(
...
-        val lifecycleAwareObserver = MvRxLifecycleAwareObserver(
+        @Suppress("UNCHECKED_CAST") val lifecycleAwareObserver = MvRxLifecycleAwareObserver(
             lifecycleOwner,
-            alwaysDeliverLastValueWhenUnlocked = !uniqueOnly,
-            onNext = Consumer<T> { subscriber(it) }
+            deliveryMode = deliveryMode,
+            lastDeliveredValue = if (deliveryMode is UniqueOnly) {
+                if (activeSubscriptions.contains(deliveryMode.subscriptionId)) {
+                    throw IllegalStateException("Subscribing with a duplicate subscription id: ${deliveryMode.subscriptionId}. " +
+                        "If you have multiple unique only subscriptions in a MvRx view that listen to the same")
```
elihart: looks like this comment wasn't finished

mvrx/src/main/kotlin/com/airbnb/mvrx/BaseMvRxViewModel.kt
```diff
@@ -506,10 +513,30 @@ abstract class BaseMvRxViewModel<S : MvRxState>(
...
+                }
+                activeSubscriptions.add(deliveryMode.subscriptionId)
+                lastDeliveredStates[deliveryMode.subscriptionId] as? T
```
elihart: can we remove the value from the map `lastDeliveredStates` too, or does it need to be stored for later?

mvrx/src/main/kotlin/com/airbnb/mvrx/BaseMvRxViewModel.kt
```diff
@@ -506,10 +513,30 @@ abstract class BaseMvRxViewModel<S : MvRxState>(
...
+            },
+            onNext = Consumer { value ->
+                if (deliveryMode is UniqueOnly) {
+                    lastDeliveredStates[deliveryMode.subscriptionId] = value
+                }
+                subscriber(value)
+            },
+            onDispose = {
+                if (deliveryMode is UniqueOnly) {
+                    activeSubscriptions.remove(deliveryMode.subscriptionId)
```
elihart: nice

mvrx/src/main/kotlin/com/airbnb/mvrx/BaseMvRxViewModel.kt
BenSchwab: Good catch! we can delete it now since we enforce subscriptions to be unique

PR #208, opened 2019-03-12
Fixed race condition in the state reducers. Also fixed missing reducer

hellohuanlin:
Fix the race condition introduced here https://github.com/airbnb/MvRx/pull/206
Also added detailed explanation on the job scheduling algorithm

1. Fixed the race condition

let's say we have `[G1, G2, G3]` in getStateQueue and `[S1, S2, S3]` in setStateQueue. we want to purge all [S1, S2, S3] before taking any task in getStateQueue. With the original implementation, the execution order will be `S1 -> G1 -> S2 -> G2 -> S3 -> G3`, which has the race condition. With this change, the execution order will be `S1 -> S2 -> S3 -> G1 -> G2 -> G3`

2. Also fix a missing reducer. 

Let's say we have [G1] and [S1, S2, S3], in the original implementation, execution order will be `S1 -> G1 -> S2` (`S3` is missing). with this PR update, execution order will be `S1 -> S2 -> S3 -> G1`

@gpeal @BenSchwab @elihart

### DIFF
--- a/mvrx/src/main/kotlin/com/airbnb/mvrx/RealMvRxStateStore.kt
+++ b/mvrx/src/main/kotlin/com/airbnb/mvrx/RealMvRxStateStore.kt
@@ -82,11 +82,11 @@ class RealMvRxStateStore<S : Any>(initialState: S) : MvRxStateStore<S> {
     /**
      * Job scheduling algorithm
      * We use double-queue design to prioritize `setState` blocks over `getState` blocks.
-     * `setStateQueue` has higher priority and needs to be flushed before taking tasks from `setStateQueue`.
+     * `setStateQueue` has higher priority and needs to be flushed before taking tasks from `getStateQueue`.
      * If a `getState` block enqueues a `setState` block, it should be executed before executing the next `getState` block.
      * This is to prevent a race condition when `getState` itself enqueues a `setState`, for example:
      * ```
-     * withState { state ->
+     * getState { state ->
      *   if (state.isLoading) return
      *   setState { state ->
      *     state.copy(isLoading = true)
@@ -94,7 +94,7 @@ class RealMvRxStateStore<S : Any>(initialState: S) : MvRxStateStore<S> {
      *   // make a network call
      * }
      * ```
-     * In the above example, we have to run the inner `setState` before the next `withState`.
+     * In the above example, we have to run the inner `setState` before the next `getState`.
      * Otherwise if we call this code twice consecutively, we could end up with making network call twice.
      *
      * Let's simplify the scenario as following:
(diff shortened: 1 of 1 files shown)


mvrx/src/main/kotlin/com/airbnb/mvrx/RealMvRxStateStore.kt
```diff
@@ -120,10 +187,13 @@ class RealMvRxStateStore<S : Any>(initialState: S) : MvRxStateStore<S> {
     }
 
     private fun flushSetStateQueue() {
```
BenSchwab: @hellohuanlin isn't this called `flushQueues()` which does drain the setState queue. Not sure what is changing here. Could you clarify?

mvrx/src/main/kotlin/com/airbnb/mvrx/RealMvRxStateStore.kt
```diff
@@ -78,6 +78,67 @@ class RealMvRxStateStore<S : Any>(initialState: S) : MvRxStateStore<S> {
         flushQueueSubject.onNext(Unit)
     }
 
+
+    /**
```
BenSchwab: Thanks for adding this comment!

mvrx/src/main/kotlin/com/airbnb/mvrx/RealMvRxStateStore.kt
hellohuanlin: let's say we have `[G1, G2, G3]` in `getStateQueue` and `[S1, S2, S3]` in `setStateQueue`. we want to purge all `[S1, S2, S3]` before taking any task in `getStateQueue`. With the original implementation, the execution order will be `S3 -> G3 -> S2 -> G2 -> S1 -> G1`, which has the race condition. With this change, the execution order will be `S3 -> S2 -> S1 -> G3 -> G2 -> G1`

hellohuanlin: Oh, actually if you follow the execution path, the original implementation actually may have missing reducers. 

Let's say we have `[G1]` and `[S1, S2, S3]`, in the original implementation, execution order will be `S1 -> G1 -> S2` (`S3` is missing)

So this PR not just fix the race condition, but also fix the missing reducer.

hellohuanlin: I will update the PR description

hellohuanlin:
@gpeal bump

hellohuanlin:
@gpeal yeah i figured that it was probably introduced by mistake. Sorry I didn't see your comment earlier. Will add more tests for this in case of future regression.

PR #209, opened 2019-03-12
Fix saved state of editing TODO sample

elihart:
Addresses https://github.com/airbnb/MvRx/issues/203

Epoxy state cannot be used as it clashes with the text we need to set from the view model - the best approach is to have a local fragment view model to store changed text.

This is a better pattern that getting text off of the recyclerview too, and also provides a good example for how to work with edit texts and epoxy.

@gpeal @BenSchwab

### DIFF
--- a/todomvrx/src/main/java/com/airbnb/mvrx/todomvrx/core/BaseFragment.kt
+++ b/todomvrx/src/main/java/com/airbnb/mvrx/todomvrx/core/BaseFragment.kt
@@ -33,6 +33,11 @@ abstract class BaseFragment : BaseMvRxFragment() {
     // Used to keep track of task changes to determine if we should show a snackbar.
     private var oldTasks: Tasks? = null
 
+    override fun onCreate(savedInstanceState: Bundle?) {
+        super.onCreate(savedInstanceState)
+        epoxyController.onRestoreInstanceState(savedInstanceState)
+    }
+
     override fun onCreateView(inflater: LayoutInflater, container: ViewGroup?, savedInstanceState: Bundle?): View? =
         inflater.inflate(R.layout.fragment_base, container, false).apply {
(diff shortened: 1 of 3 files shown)


todomvrx/src/main/java/com/airbnb/mvrx/todomvrx/core/BaseFragment.kt
```diff
@@ -33,6 +33,11 @@ abstract class BaseFragment : BaseMvRxFragment() {
     // Used to keep track of task changes to determine if we should show a snackbar.
     private var oldTasks: Tasks? = null
 
+    override fun onCreate(savedInstanceState: Bundle?) {
+        super.onCreate(savedInstanceState)
+        epoxyController.onRestoreInstanceState(savedInstanceState)
```
elihart: this isn't used, but it's good as an example

todomvrx/src/main/java/com/airbnb/mvrx/todomvrx/views/AddEditView.kt
```diff
@@ -1,36 +1,115 @@
...
+    var onDescriptionChanged: ((newText: String) -> Unit)? = null
+
+}
+
+/**
+ * Set the given text on the textview.
+ *
+ * @return True if the given text is different from the previous text on the textview.
+ */
+fun EditText.setTextIfDifferent(newText: CharSequence?): Boolean {
```
gpeal: I've actually considered making a library of idempotent base classes like TextViews, ImageViews, etc with this built in.

elihart: That would be great, people ask about that for Epoxy a lot

PR #210, opened 2019-03-12
For subscriptions made after onCreateView use View lifecycle observer

BenSchwab:
Closes https://github.com/airbnb/MvRx/issues/118

When a Fragment is added to the backstack, certain lifecycle methods are called multiple times on the same instance of the fragment (e.g. onCreateView). Currently, if you create a MvRx subscription there, it will be duplicated.

This PR creates the concept of a `subscriptionLifecycleOwner` which can be dynamically changed depending on the lifecycle a fragment is in to return a ViewLifecycle owner instead.

Adds a test for backstack behavior.

@gpeal @elihart


mvrx/src/main/kotlin/com/airbnb/mvrx/MvRxView.kt
```diff
@@ -57,7 +68,7 @@ interface MvRxView : MvRxViewModelStoreOwner, LifecycleOwner {
      *
      * Default: [RedeliverOnStart].
      */
-    fun <S : MvRxState> BaseMvRxViewModel<S>.subscribe(deliveryMode: DeliveryMode = RedeliverOnStart, subscriber: (S) -> Unit) = subscribe(this@MvRxView, deliveryMode, subscriber)
+    fun <S : MvRxState> BaseMvRxViewModel<S>.subscribe(deliveryMode: DeliveryMode = RedeliverOnStart, subscriber: (S) -> Unit) = subscribe(this@MvRxView.subscriptionLifecycleOwner, deliveryMode, subscriber)
```
elihart: it seems a little dangerous to continue to have MvRxView implement LifecycleOwner, since it can easily be passed instead of `subscriptionLifecycleOwner` on accident.

Since we're making breaking changes for 1.0, do you think we should consider changing this too?

BenSchwab: I thought about that (moving it as a property instead of an interface), but it seems like it would still be easy to use the wrong one.

However, almost all subscriptions should be going through the functions in MvRxView so I don't think there is too much risk.

gpeal: In debug mode, we could check if the view is an instance of Fragment and if the subscriptionLifecycleOwner == lifecycleOwner

BenSchwab: @gpeal The debug flag is only in the ViewModel, so no clean way to do this check in the Fragment. I tried to do it in the BaseMvrxViewModel, but a fragment may properly subscribe with its own or its view lifecycle owner depending on when the subscribe is called.

If someone makes a mistake in a custom MvRxView, it will result in a duplicate subscription, which we do throw an error for. I updated the error message to explain this possibility.

PR #225, opened 2019-04-10
Allow Fail to be used with Exceptions with no stack trace

gpeal:
Coroutine CancellationException can have an empty stack trace.

Also fixed Detekt

### DIFF
--- a/mvrx/src/main/kotlin/com/airbnb/mvrx/Async.kt
+++ b/mvrx/src/main/kotlin/com/airbnb/mvrx/Async.kt
@@ -69,7 +69,7 @@ data class Fail<out T>(val error: Throwable) : Async<T>(complete = true, shouldL
         val otherError = other.error
         return error::class == otherError::class &&
             error.message == otherError.message &&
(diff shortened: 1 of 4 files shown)


mvrx/src/main/kotlin/com/airbnb/mvrx/Async.kt
```diff
@@ -68,8 +68,8 @@ data class Fail<out T>(val error: Throwable) : Async<T>(complete = true, shouldL
 
         val otherError = other.error
         return error::class == otherError::class &&
-                error.message == otherError.message &&
-                error.stackTrace[0] == otherError.stackTrace[0]
+            error.message == otherError.message &&
+            (error.stackTrace.size == otherError.stackTrace.size && (error.stackTrace.isEmpty() || error.stackTrace[0] == otherError.stackTrace[0]))
```
elihart: `error.stackTrace.firstOrNull() == otherError.stackTrace.firstOrNull`?

gpeal: Good idea!

PR #235, opened 2019-04-24
Not using lifecycle aware observer for unit tests

### LINKED ISSUE #198: MvRxLifecycleAwareObserver breaks unit testability
Currently all kind of `subscribe` and `asyncSubscribe` methods strongly depend on `MvRxLifecycleAwareObserver` which then depends on Android APIs. 

Example call:

```
viewModel.asyncSubscribe(LoginState::token, onSuccess = {
    navController.navigate(R.id.open_search)
 })
```

It is almost impossible to unit test the implementation of functions given into `asyncSubscribe` and its siblings. Basically in unit tests, the fragment lifecycle will never be in resumed state and that's why `MvRxLifecycleAwareObserver` never calls the `subsriber` function.

kojadin:
Using MvRxTestRule to force not using life cycle aware observable for unit tests. 
Fixes #198


mvrx/src/main/kotlin/com/airbnb/mvrx/BaseMvRxViewModel.kt
```diff
@@ -541,44 +542,49 @@ abstract class BaseMvRxViewModel<S : MvRxState>(
...
-                if (deliveryMode is UniqueOnly) {
-                    lastDeliveredStates[deliveryMode.subscriptionId] = value
-                }
-                subscriber(value)
-            },
-            onDispose = {
-                if (deliveryMode is UniqueOnly) {
-                    activeSubscriptions.remove(deliveryMode.subscriptionId)
+    private fun <T : Any> Observable<T>.resolveSubscription(
+        subscriber: (T) -> Unit,
```
tasomaniac: Small comment: I would keep the order of these params consistent with the above functions.

mvrx/src/main/kotlin/com/airbnb/mvrx/BaseMvRxViewModel.kt
```diff
@@ -33,6 +33,7 @@ abstract class BaseMvRxViewModel<S : MvRxState>(
     private val stateStore: MvRxStateStore<S> = RealMvRxStateStore(initialState)
 ) : ViewModel() {
     private val debugMode = if (MvRxTestOverrides.FORCE_DEBUG == null) debugMode else MvRxTestOverrides.FORCE_DEBUG
+    private val forceUserTestObserver = MvRxTestOverrides.FORCE_USE_TEST_OBSERVER
```
gpeal: typo: `forceUseTestObserver`

testing/src/main/kotlin/com/airbnb/mvrx/test/MvRxTestRule.kt
```diff
@@ -32,6 +32,7 @@ class MvRxTestRule(
         RxAndroidPlugins.setMainThreadSchedulerHandler { Schedulers.trampoline() }
         if (setRxImmediateSchedulers) setRxImmediateSchedulers()
         MvRxTestOverridesProxy.forceMvRxDebug(debugMode.value)
+        MvRxTestOverridesProxy.forceUseTestObserver(true)
```
gpeal: Did you intentionally make this not configurable? You can include it in the TestRule constructor as an optional and default-true param.

mvrx/src/main/kotlin/com/airbnb/mvrx/BaseMvRxViewModel.kt
elihart: looks like a spelling mistake `user` instead of `use`

Is it necessary to add a property instead of referencing `FORCE_USE_TEST_OBSERVER` directly?

kojadin: Hm, actually on second thought , maybe I should call this `DISABLE_LIFECYCLE_AWARE_OBSERVER` . 
About using directly flag I was just trying to fallow same code style as you guys have for `debugMode` 
WYT ?

kojadin: Actually, I was thinking to rename to `forceDisableLifecycleAwareObserver` and flag to `FORCE_DISABLE_LIFECYCLE_AWARE_OBSERVER` . WYT?

testing/src/main/kotlin/com/airbnb/mvrx/test/MvRxTestRule.kt
kojadin: I was thinking to make it configurable, similar as for `debugMode`, but I could think on case when you will need lifecycle aware observer for unit tests. You can't use `MvRxTestRule` for automation tests because everything would happen in main thread. Maybe robolectric ?

mvrx/src/main/kotlin/com/airbnb/mvrx/BaseMvRxViewModel.kt
kojadin: Make sense.

testing/src/main/kotlin/com/airbnb/mvrx/test/MvRxTestRule.kt
kojadin: Actually on second look Rx immediate schedulers are also configurable, so I make this also configurable.

kojadin:
Tests added.

mvrx/src/main/kotlin/com/airbnb/mvrx/BaseMvRxViewModel.kt
gpeal: @kojadin you can use it directly inline and the new name sounds good 👍

testing/src/test/kotlin/com/airbnb/android/mvrx/test/TestRuleView.kt
```diff
@@ -0,0 +1,21 @@
...
+
+class TestRuleView : BaseMvRxFragment() {
+
+    var subscribeCallCount = 0
+
+    private val viewModel = TestRuleViewModel()
+
+    init {
+        viewModel.subscribe {
+            subscribeCallCount++
```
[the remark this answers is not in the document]
BenSchwab: This is actually a fragment, I think the fact that it is in `init` is confusing, but I understand that's probably easier than driving lifecycle in the tests.

mvrx/src/main/kotlin/com/airbnb/mvrx/MvRxTestOverrides.java
```diff
@@ -8,4 +8,12 @@
      * This is Java so it can be packgage private.
      */
     static Boolean FORCE_DEBUG = null;
+
+    /**
+     * This should only be set by the MvRxTestRule from the mvrx-testing artifact.
+     *
+     * This can be used to force MvRxViewModels to disable lifecycle aware observer for unit testing.
+     * This is Java so it can be packgage private.
```
BenSchwab: typo `packgage`

testing/src/test/kotlin/com/airbnb/android/mvrx/test/TestRuleView.kt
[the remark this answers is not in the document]
tasomaniac: So, this was the exact problem. This subscribe is an extension function available inside MvRxView. Without the TestRule, this callback was never called in unit tests.

mvrx/src/main/kotlin/com/airbnb/mvrx/BaseMvRxViewModel.kt
kojadin: 👍 Done

testing/src/test/kotlin/com/airbnb/android/mvrx/test/TestRuleView.kt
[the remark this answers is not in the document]
kojadin: Exactly, this is a view and it would not be called before. There is a test `MvRxTestRuleTestLifecycleAwareObserverEnabled` where you can see that this wouldn't be called. The reason I used `init` is because I was trying to be consistent and use same approach that you guys have in `TestRuleViewModel`

mvrx/src/main/kotlin/com/airbnb/mvrx/MvRxTestOverrides.java
kojadin: Fixed 👍

testing/src/test/kotlin/com/airbnb/android/mvrx/test/TestRuleView.kt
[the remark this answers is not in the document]
gpeal: Got it. Thanks for the explanation

PR #246, opened 2019-06-21
[WIP] Allow usage of fragmentViewModel and activityViewModel in custom views

gpeal:
**This PR is still undergoing active development**

This PR enables the retrieving and subscribing to ViewModels in a view-lifeycle-aware way. To use MvRx in a custom view, make your custom view implement `StatefulView` and override `onInvalidate()`. Naming is tough here because `MvRxView` is already taken and View already has an `invalidate()` function... But I'm open to alternative names.

Fragment retrieval works by walking up the view hierarchy from the custom view until it reaches the root view of a Fragment that is in any FragmentManager or child FragmentManager of the view's Activity.

Thoughts? Any reasons you can think of that this wouldn't work?

I'll add tests once we agree on this as a path forward.

As a demo, I created a counter where the text views subscribe to the state rather than needing the state passed in from the Fragment. 
CustomView

### Potential TODO
* Remove the ability to change a key and VM memoization
* Create a function on BaseMvRxFragment to fetch a ViewModel and pass it optional args but just return it without subscribing. This can be passed to a View
* Remove StatefulView and the concept of invalidate on VIews
* Create subscribe and selectSubscribe for Views. If a new VM is passed in, it is responsible for managing the old subscription disposables.
* We might want to make ViewViewModelStore clear a ViewModel when the key changes because each one might be holding on to SQL triggers and if the key changes frequently, they will just accumulate.

### DIFF
--- a/mvrx/src/main/kotlin/com/airbnb/mvrx/Memoize.kt
+++ b/mvrx/src/main/kotlin/com/airbnb/mvrx/Memoize.kt
@@ -1,0 +1,24 @@
+package com.airbnb.mvrx
+
+import io.reactivex.disposables.Disposable
+
+class Memoize<T>(private val factory: (key: String) -> Pair<T, Disposable>) {
+    private var key: String? = null
+    private var value: T? = null
+    private var disposable: Disposable? = null
+
+    val isInitialized get() = value != null
+
+    @Synchronized
+    fun get(key: String): T {
+        val currentValue = value
+        if (this.key == key && currentValue != null && disposable?.isDisposed != true) return currentValue
+
+        disposable?.dispose()
+        this.key = key
+        val (value, disposable) = factory(key)
+        this.value = value
+        this.disposable = disposable
+        return value
+    }
+}
--- a/mvrx/src/main/kotlin/com/airbnb/mvrx/MvRxExtensions.kt
+++ b/mvrx/src/main/kotlin/com/airbnb/mvrx/MvRxExtensions.kt
@@ -1,10 +1,14 @@
 package com.airbnb.mvrx
 
-import androidx.lifecycle.ViewModelProviders
+import android.app.Activity
+import android.view.View
 import androidx.annotation.RestrictTo
 import androidx.annotation.RestrictTo.Scope.LIBRARY
 import androidx.fragment.app.Fragment
 import androidx.fragment.app.FragmentActivity
+import androidx.lifecycle.ViewModelProviders
+import io.reactivex.internal.disposables.EmptyDisposable
+import java.lang.IllegalArgumentException
 import kotlin.properties.ReadOnlyProperty
 import kotlin.reflect.KClass
 import kotlin.reflect.KProperty
@@ -23,10 +27,10 @@ import kotlin.reflect.KProperty
  */
 inline fun <T, reified VM : BaseMvRxViewModel<S>, reified S : MvRxState> T.fragmentViewModel(
     viewModelClass: KClass<VM> = VM::class,
-    crossinline keyFactory: () -> String = { viewModelClass.java.name }
-) where T : Fragment, T : MvRxView = lifecycleAwareLazy(this) {
-    MvRxViewModelProvider.get(viewModelClass.java, S::class.java, FragmentViewModelContext(this.requireActivity(), _fragmentArgsProvider(), this), keyFactory())
-        .apply { subscribe(this@fragmentViewModel, subscriber = { postInvalidate() }) }
+    noinline keyFactory: () -> String = { viewModelClass.java.name }
+) where T : Fragment, T : MvRxView = lifecycleAwareLazy(this, keyFactory) { key ->
+    val viewModel = MvRxViewModelProvider.get(viewModelClass.java, S::class.java, FragmentViewModelContext(this.requireActivity(), _fragmentArgsProvider(), this, key), key)
+    viewModel to viewModel.subscribe(this@fragmentViewModel, subscriber = { postInvalidate() })
 }
 
 /**
@@ -35,11 +39,11 @@ inline fun <T, reified VM : BaseMvRxViewModel<S>, reified S : MvRxState> T.fragm
  */
 inline fun <T, reified VM : BaseMvRxViewModel<S>, S : MvRxState> T.existingViewModel(
     viewModelClass: KClass<VM> = VM::class,
-    crossinline keyFactory: () -> String = { viewModelClass.java.name }
-) where T : Fragment, T : MvRxView = lifecycleAwareLazy(this) {
-    val factory = MvRxFactory { throw IllegalStateException("ViewModel for ${requireActivity()}[${keyFactory()}] does not exist yet!") }
-    ViewModelProviders.of(requireActivity(), factory).get(keyFactory(), viewModelClass.java)
-        .apply { subscribe(this@existingViewModel, subscriber = { postInvalidate() }) }
+    noinline keyFactory: () -> String = { viewModelClass.java.name }
+) where T : Fragment, T : MvRxView = lifecycleAwareLazy(this, keyFactory) { key ->
+    val factory = MvRxFactory { error("ViewModel for ${requireActivity()}[${key}] does not exist yet!") }
+    val viewModel = ViewModelProviders.of(requireActivity(), factory).get(key, viewModelClass.java)
+    viewModel to viewModel.subscribe(this@existingViewModel, subscriber = { postInvalidate() })
 }
 
 /**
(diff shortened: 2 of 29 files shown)


mvrx/src/main/kotlin/com/airbnb/mvrx/viewLifecycleAwareLazy.kt
```diff
@@ -0,0 +1,89 @@
...
+import androidx.appcompat.app.AppCompatActivity
+import androidx.fragment.app.Fragment
+import java.io.Serializable
+
+/**
+ * This was copied from SynchronizedLazyImpl but modified to automatically initialize in onAttachedToWindow.
+ */
+@RestrictTo(RestrictTo.Scope.LIBRARY)
+@SuppressWarnings("Detekt.ClassNaming")
+class viewLifecycleAwareLazy<out T>(private val view: View, initializer: (AppCompatActivity, Fragment) -> T) : Lazy<T>, Serializable {
```
gpeal: This is mostly the same as `lifecycleAwareLazy` except it was modified to use window attach state rather than a lifecycle owner.

mvrx/src/main/kotlin/com/airbnb/mvrx/MvRxExtensions.kt
```diff
@@ -54,6 +55,53 @@ inline fun <T, reified VM : BaseMvRxViewModel<S>, reified S : MvRxState> T.activ
...
+inline fun <T, reified VM : BaseMvRxViewModel<S>, reified S : MvRxState> T.fragmentViewModel(
+    viewModelClass: KClass<VM> = VM::class,
+    crossinline keyFactory: () -> String = { viewModelClass.java.name }
+) where T : View, T : StatefulView = viewLifecycleAwareLazy(this) { activity, fragment ->
+    MvRxViewModelProvider.get(viewModelClass.java, S::class.java, FragmentViewModelContext(activity, fragment._fragmentArgsProvider(), fragment), keyFactory())
+        .apply { subscribe(fragment.viewLifecycleOwner, subscriber = { onInvalidate() }) }
+}
+
+/**
+ * [fragmentViewModel] except scoped to the current Activity. Use this to share state between different Fragments.
```
elihart: Some people use view architectures directly without fragments - it would be nice to make activity/existing view model not related to fragments (ie remove the `share state between different Fragments.` wording)

mvrx/src/main/kotlin/com/airbnb/mvrx/MvRxExtensions.kt
```diff
@@ -54,6 +55,53 @@ inline fun <T, reified VM : BaseMvRxViewModel<S>, reified S : MvRxState> T.activ
...
+ *
+ * If the ViewModel has additional dependencies, implement [MvRxViewModelFactory] in its companion object.
+ * You will be given the initial state as well as a FragmentActivity with which you can access other dependencies to
+ * pass to the ViewModel's constructor.
+ *
+ * MvRx will also handle persistence across process restarts. Refer to [PersistState] for more info.
+ *
+ * Use [keyFactory] if you have multiple ViewModels of the same class in the same scope.
+ */
+inline fun <T, reified VM : BaseMvRxViewModel<S>, reified S : MvRxState> T.fragmentViewModel(
```
elihart: So there is no concept of `existing` view model for views? If the fragment, or another view in the fragment, also uses this then it will grab the existing fragment view model?

I think it's confusing to have `activityViewModel` and `existinViewModel` when we don't make that distinction here.

This feels related to the discussion I brought on slack for whether `existingViewModel` is necessary

mvrx/src/main/kotlin/com/airbnb/mvrx/MvRxExtensions.kt
```diff
@@ -54,6 +55,53 @@ inline fun <T, reified VM : BaseMvRxViewModel<S>, reified S : MvRxState> T.activ
...
+}
+
+/**
+ * [fragmentViewModel] except scoped to the current Activity. Use this to share state between different Fragments.
+ */
+inline fun <T, reified VM : BaseMvRxViewModel<S>, reified S : MvRxState> T.activityViewModel(
+    viewModelClass: KClass<VM> = VM::class,
+    crossinline keyFactory: () -> String = { viewModelClass.java.name }
+) where T : View, T : StatefulView = viewLifecycleAwareLazy(this) { activity, fragment ->
+    if (activity !is MvRxViewModelStoreOwner) throw IllegalArgumentException("Your Activity must be a MvRxViewModelStoreOwner!")
```
elihart: a little more idiomatic - `require(activity is MvRxViewModelStoreOwner) { "Your activity..." }`

this is shorter, and still throws the IllegalArgumentException

mvrx/src/main/kotlin/com/airbnb/mvrx/MvRxExtensions.kt
```diff
@@ -54,6 +55,53 @@ inline fun <T, reified VM : BaseMvRxViewModel<S>, reified S : MvRxState> T.activ
...
+
+/**
+ * [fragmentViewModel] except scoped to the current Activity. Use this to share state between different Fragments.
+ */
+inline fun <T, reified VM : BaseMvRxViewModel<S>, reified S : MvRxState> T.existingViewModel(
+    viewModelClass: KClass<VM> = VM::class,
+    crossinline keyFactory: () -> String = { viewModelClass.java.name }
+) where T : View, T : StatefulView = viewLifecycleAwareLazy(this) { activity, fragment ->
+    if (activity !is MvRxViewModelStoreOwner) throw IllegalArgumentException("Your Activity must be a MvRxViewModelStoreOwner!")
+    val factory = MvRxFactory { throw IllegalStateException("ViewModel for ${activity}[${keyFactory()}] does not exist yet!") }
```
elihart: replace `throw IllegalStateException` with just `error`

mvrx/src/main/kotlin/com/airbnb/mvrx/UninitializedValue.kt
```diff
@@ -0,0 +1,3 @@
+package com.airbnb.mvrx
+
+object UninitializedValue
```
elihart: meant to be public? can it be internal?

mvrx/src/main/kotlin/com/airbnb/mvrx/StatefulView.kt
```diff
@@ -0,0 +1,12 @@
+package com.airbnb.mvrx
+
+/**
+ * Implement this in a custom view to expose [fragmentViewModel] and [activityViewModel] in a custom view. Similar to usage in Fragments,
+ * this will create a view-lifecycle-aware subscription.
+ * For [fragmentViewModel], this will walk up the view hierarchy until it finds a View that is the root view for a Fragment.
+ * [StatefulView] uses [onInvalidate] instead of `invalidate()` because `invalidate()` is already a function on View. It is also
+ * not called [MvRxView] because the name was already taken.
+ */
+interface StatefulView {
```
elihart: kind of a shame that we can't better combine this with the existing `MvRxView` - I wonder if that should be named `MvRxFragment`? it would also be nice if we could rename `MvRxView#invalidate` to match this.

honestly I'm kind of ok with breaking changes at this point to keep this more in sync

mvrx/src/main/kotlin/com/airbnb/mvrx/viewLifecycleAwareLazy.kt
```diff
@@ -0,0 +1,89 @@
...
+
+        var view: View? = this
+        while (view != null) {
+            val fragment = viewFragmentMap[view]
+            if (fragment != null) {
+                return fragment
+            }
+            view = view.parent as? ViewGroup
+        }
+        throw java.lang.IllegalStateException("Unable to find a Fragment in any of the parents of $this.")
```
elihart: replace `throw java.lang.IllegalStateException` with `error`

littleGnAl:
Seems it doesn't re-invalidate when the `@ModelView` rebind when using it with epoxy.

mvrx/src/main/kotlin/com/airbnb/mvrx/StatefulView.kt
gpeal: @elihart I'm open to have a full discussion about this. The original idea behind MvRxView is that you could use it for other things like conductor or other custom architectures. However, the full functionality of MvRxView is explicitly not needed here which is now awkward.

Renaming invalidate to onInvalidate everywhere is possible but would require a lot of updates but each one would be a simple copy/replace. What if we renamed it `render` to make it more like React?

gpeal:
@littleGnAl Can you explain what you mean?

littleGnAl:
@gpeal  I add a new fragment, then navigate to it, after clicking the `statefulBasicRow`, then navigate to it again, the behavior is so strange, I think this may be the problem relates to the view lifecycle:
```kotlin
class StatefulViewEpoxyFragment : BaseFragment() {
    private val fragmentScopedViewModel: StatefulViewViewModel by fragmentViewModel()

    override fun epoxyController(): MvRxEpoxyController = simpleController {
        repeat(5) {
            val index = it + 1
            statefulBasicRow {
                id("statefulBasicRow_$index")
                title("statefulBasicRow $index")
                clickListener { _, _, _, _ ->
                    fragmentScopedViewModel.setTitle("New Title $index")
                }
            }
        }
    }
}
```

gpeal:
@littleGnAl This PR is still under heavy development but yes, you are right, there are a number of issues wit this PR as-is

mvrx/src/main/kotlin/com/airbnb/mvrx/StatefulView.kt
elihart: @gpeal I would be ok with a rename to `render` I think. or is there a keyword that Compose is using that we could match now?

thoughts @BenSchwab

mvrx/src/main/kotlin/com/airbnb/mvrx/ViewLifecycleOwner.kt
```diff
@@ -0,0 +1,105 @@
...
+                handleAttachedToWindow()
+            }
+            return _fragment
+        }
+    val activity get() = view.context as? AppCompatActivity
+
+    init {
+        view.addOnAttachStateChangeListener(object : View.OnAttachStateChangeListener {
+            override fun onViewDetachedFromWindow(v: View) {
+                lifecycleRegistry.handleLifecycleEvent(Lifecycle.Event.ON_STOP)
```
elihart: onAttach does `onresume`, which is not the corollary to `onStop` - this should be `onPause` if we want to make them symmetric

mvrx/src/main/kotlin/com/airbnb/mvrx/ViewViewModelStore.kt
```diff
@@ -0,0 +1,95 @@
...
+            store.clear()
+        }
+    }
+
+    private lateinit var activity: AppCompatActivity
+
+    private val referenceQueue = ReferenceQueue<View>()
+    private val viewMap = mutableMapOf<ViewReference, ViewModelStore>()
+    /** Helper to find the right key for [viewMap]. */
+    private val viewIdeneityRefMap = mutableMapOf<Int, ViewReference>()
```
elihart: typo? `Ideneity`

mvrx/src/main/kotlin/com/airbnb/mvrx/Memoize.kt
```diff
@@ -0,0 +1,24 @@
...
+    private var key: String? = null
+    private var value: T? = null
+    private var disposable: Disposable? = null
+
+    val isInitialized get() = value != null
+
+    @Synchronized
+    fun get(key: String): T {
+        val currentValue = value
+        if (this.key == key && currentValue != null && disposable?.isDisposed != true) return currentValue
```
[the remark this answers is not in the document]
gpeal: @elihart Not in the context of using this as memoization for ViewModels and the fact that the return type is T not T?.

mvrx/src/main/kotlin/com/airbnb/mvrx/Memoize.kt
```diff
@@ -0,0 +1,24 @@
+package com.airbnb.mvrx
+
+import io.reactivex.disposables.Disposable
+
+class Memoize<T>(private val factory: (key: String) -> Pair<T, Disposable>) {
```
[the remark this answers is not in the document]
gpeal: If we want to move forward with this approach, I'm going to add a bunch of docs and tests. I'm not fully convinced that this is the right path forward yet though since it introduces a number of complexities in the API an handling manual subscriptions.

mvrx/src/main/kotlin/com/airbnb/mvrx/MvRxInvalidator.kt
```diff
@@ -0,0 +1,27 @@
...
+        val obj = message.obj
+        pendingInvalidates.remove(System.identityHashCode(obj))
+        when {
+            obj is MvRxView -> if (obj.lifecycle.currentState.isAtLeast(Lifecycle.State.STARTED)) obj.invalidate()
+            obj is StatefulView && obj is View -> if (obj.isAttachedToWindow()) obj.onInvalidate()
+        }
+        true
+    })
+
+    fun post(obj: Any) {
```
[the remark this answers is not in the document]
gpeal: Totally agree. Again, I'll add a bunch of docs when we feel good about a direction for this.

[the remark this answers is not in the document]
gpeal: It probably should but I just pulled this from the implementation in MvRxView which has worked seemingly without issue since launch.

mvrx/src/main/kotlin/com/airbnb/mvrx/ViewLifecycleOwner.kt
gpeal: @elihart True but it wouldn't have the correct behavior of locking the lifecycle aware observer. This is one of the challenges of mapping the view lifecycle to the fragment/activity lifecycle

gpeal:
Closing this for now. I can't think of a way to do this reliably without becoming error prone.

PR #247, opened 2019-06-23
Added parentFragmentViewModel()

### LINKED ISSUE #227: Parent Fragment ViewModel
What is the best way to get the view model for the parent fragment?
Would prefer to not scope it to the activity.

I did try to create my own extension function based on `fragmentViewModel` but `lifecycleAwareLazy`, `MvRxViewModelProvider` and a few others are private.

gpeal:
Fixes #227

### DIFF
--- a/mvrx/src/main/kotlin/com/airbnb/mvrx/MvRxExtensions.kt
+++ b/mvrx/src/main/kotlin/com/airbnb/mvrx/MvRxExtensions.kt
@@ -1,10 +1,10 @@
 package com.airbnb.mvrx
 
-import androidx.lifecycle.ViewModelProviders
 import androidx.annotation.RestrictTo
 import androidx.annotation.RestrictTo.Scope.LIBRARY
 import androidx.fragment.app.Fragment
 import androidx.fragment.app.FragmentActivity
+import androidx.lifecycle.ViewModelProviders
 import kotlin.properties.ReadOnlyProperty
 import kotlin.reflect.KClass
 import kotlin.reflect.KProperty
@@ -27,6 +27,40 @@ inline fun <T, reified VM : BaseMvRxViewModel<S>, reified S : MvRxState> T.fragm
 ) where T : Fragment, T : MvRxView = lifecycleAwareLazy(this) {
     MvRxViewModelProvider.get(viewModelClass.java, S::class.java, FragmentViewModelContext(this.requireActivity(), _fragmentArgsProvider(), this), keyFactory())
         .apply { subscribe(this@fragmentViewModel, subscriber = { postInvalidate() }) }
+}
+
+/**
+ * Gets or creates a ViewModel scoped to a parent fragment. This delegate will walk up the parentFragment hierarchy
+ * until it finds a Fragment that can provide the correct ViewModel. If no parent fragments can provide the ViewModel,
+ * a new one will be created in the direct parent of the curent Fragment.
+ */
+inline fun <T, reified VM : BaseMvRxViewModel<S>, reified S : MvRxState> T.parentFragmentViewModel(
+    viewModelClass: KClass<VM> = VM::class,
+    crossinline keyFactory: () -> String = { viewModelClass.java.name }
+): Lazy<VM> where T : Fragment, T : MvRxView = lifecycleAwareLazy(this) {
+    requireNotNull(parentFragment) { "There is no parent fragment!" }
+    val factory = MvRxFactory { error("No ViewModel for this Fragment.") }
+    var fragment: Fragment? = this
+    var viewModel: VM? = null
+    val key = keyFactory()
+    while (fragment != null) {
+        try {
+            viewModel = ViewModelProviders.of(fragment, factory).get(key, viewModelClass.java)
+                .apply { subscribe(this@parentFragmentViewModel, subscriber = { postInvalidate() }) }
+            break
+        } catch (e: IllegalStateException) {
+            fragment = fragment.parentFragment
+        }
+    }
+    if (viewModel == null) {
+        val viewModelContext = FragmentViewModelContext(this.requireActivity(), _fragmentArgsProvider(), parentFragment as Fragment)
+        viewModel = MvRxViewModelProvider.get(viewModelClass.java, S::class.java, viewModelContext, keyFactory())
+    }
+    // There is a mismatch between the compiler inference and Android Studio inference.
+    // This code doesn't compile without the cast.
+    // May be related to: https://blog.jetbrains.com/kotlin/2019/06/kotlin-1-3-40-released/
+    @Suppress("USELESS_CAST")
+    viewModel as VM
 }
 
 /**
--- a/mvrx/src/main/kotlin/com/airbnb/mvrx/MvRxExtensions.kt
+++ b/mvrx/src/main/kotlin/com/airbnb/mvrx/MvRxExtensions.kt
@@ -41,31 +41,41 @@ inline fun <T, reified VM : BaseMvRxViewModel<S>, reified S : MvRxState> T.paren
     requireNotNull(parentFragment) { "There is no parent fragment!" }
     val factory = MvRxFactory { error("No ViewModel for this Fragment.") }
     var fragment: Fragment? = parentFragment
-    var viewModel: VM? = null
     val key = keyFactory()
     while (fragment != null) {
         try {
-            viewModel = ViewModelProviders.of(fragment, factory).get(key, viewModelClass.java)
+            return@lifecycleAwareLazy ViewModelProviders.of(fragment, factory).get(key, viewModelClass.java)
                 .apply { subscribe(this@parentFragmentViewModel, subscriber = { postInvalidate() }) }
-            break
         } catch (e: IllegalStateException) {
             fragment = fragment.parentFragment
         }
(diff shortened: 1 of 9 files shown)


mvrx/src/main/kotlin/com/airbnb/mvrx/MvRxExtensions.kt
```diff
@@ -29,6 +29,40 @@ inline fun <T, reified VM : BaseMvRxViewModel<S>, reified S : MvRxState> T.fragm
...
+ * until it finds a Fragment that can provide the correct ViewModel. If no parent fragments can provide the ViewModel,
+ * a new one will be created in the direct parent of the curent Fragment.
+ */
+inline fun <T, reified VM : BaseMvRxViewModel<S>, reified S : MvRxState> T.parentFragmentViewModel(
+    viewModelClass: KClass<VM> = VM::class,
+    crossinline keyFactory: () -> String = { viewModelClass.java.name }
+): Lazy<VM> where T : Fragment, T : MvRxView = lifecycleAwareLazy(this) {
+    requireNotNull(parentFragment) { "There is no parent fragment!" }
+    val factory = MvRxFactory { error("No ViewModel for this Fragment.") }
+    var fragment: Fragment? = this
```
littleGnAl: I think it should be `var fragment: Fragment? = parentFragment`, or it will always throw `"There is no parent fragment!"` in the first time.

mvrx/src/main/kotlin/com/airbnb/mvrx/MvRxExtensions.kt
```diff
@@ -29,6 +29,40 @@ inline fun <T, reified VM : BaseMvRxViewModel<S>, reified S : MvRxState> T.fragm
...
+    crossinline keyFactory: () -> String = { viewModelClass.java.name }
+): Lazy<VM> where T : Fragment, T : MvRxView = lifecycleAwareLazy(this) {
+    requireNotNull(parentFragment) { "There is no parent fragment!" }
+    val factory = MvRxFactory { error("No ViewModel for this Fragment.") }
+    var fragment: Fragment? = parentFragment
+    var viewModel: VM? = null
+    val key = keyFactory()
+    while (fragment != null) {
+        try {
+            viewModel = ViewModelProviders.of(fragment, factory).get(key, viewModelClass.java)
```
elihart: can you return here instead of storing it in a `var`?

mvrx/src/main/kotlin/com/airbnb/mvrx/MvRxExtensions.kt
```diff
@@ -29,6 +29,40 @@ inline fun <T, reified VM : BaseMvRxViewModel<S>, reified S : MvRxState> T.fragm
         .apply { subscribe(this@fragmentViewModel, subscriber = { postInvalidate() }) }
 }
 
+/**
+ * Gets or creates a ViewModel scoped to a parent fragment. This delegate will walk up the parentFragment hierarchy
+ * until it finds a Fragment that can provide the correct ViewModel. If no parent fragments can provide the ViewModel,
+ * a new one will be created in the direct parent of the curent Fragment.
```
elihart: It doesn't feel quite right that an existing viewmodel can exist anywhere in the parent fragment hierarchy, but creating a new one also puts it in the direct parent.

What if instead by default it creates it in the top level parent fragment? That seems less likely to cause issues

gpeal: @elihart @BenSchwab and I chatted about this yesterday with the same concern. Do you think the top would be a more sane default than the first parent?

elihart: Yeah, feels safer since any other child fragment is guaranteed to be able to reach the same viewmodel.

This could be important in an edge case like process recreation where we are starting with a nested child fragment that wouldn't normally be created first

gpeal: @elihart would that be the case for restoration? The ViewModel won't be created until ON_CREATE. Would the child onCreate be called before the parent? That seems impossible because onAttach is called before onCreate.

mvrx/src/main/kotlin/com/airbnb/mvrx/MvRxExtensions.kt
```diff
@@ -29,6 +29,55 @@ inline fun <T, reified VM : BaseMvRxViewModel<S>, reified S : MvRxState> T.fragm
...
+    while (topParentFragment?.parentFragment != null) {
+        topParentFragment = topParentFragment.parentFragment
+    }
+    val viewModelContext = FragmentViewModelContext(this.requireActivity(), _fragmentArgsProvider(), topParentFragment!!)
+    return@lifecycleAwareLazy MvRxViewModelProvider.get(viewModelClass.java, S::class.java, viewModelContext, keyFactory())
+        .apply { subscribe(this@parentFragmentViewModel, subscriber = { postInvalidate() }) }
+}
+
+/**
+ * Gets or creates a ViewModel scoped to a target fragment. Throws [IllegalStateException] if there is no target fragment.
```
elihart: this use case seems weird to me, because the lifecycle of the target fragment isn't necessarily tied to the fragment that is calling this. Couldn't the target fragment be destroyed, and the viewmodel cleared, while this fragment is still using it?

is it that bad to share viewmodels via an activity scoped viewmodel in this case?

I'm not particularly opposed to it, but am curious to hear more about the need. Once we have these new delegates it would be great if we updated the wiki with a little guide about when we recommend using each

gpeal: @elihart I think it alleviates some of the potential ambiguity in the parentFragment case because it allows you to be explicit what the ViewModel will be scoped to.

I haven't actually needed it myself but @littleGnAl suggested it and it seems reasonable and was simple enough.

mvrx/src/main/kotlin/com/airbnb/mvrx/MvRxExtensions.kt
```diff
@@ -29,6 +29,55 @@ inline fun <T, reified VM : BaseMvRxViewModel<S>, reified S : MvRxState> T.fragm
...
+        try {
+            return@lifecycleAwareLazy ViewModelProviders.of(fragment, factory).get(key, viewModelClass.java)
+                .apply { subscribe(this@parentFragmentViewModel, subscriber = { postInvalidate() }) }
+        } catch (e: IllegalStateException) {
+            fragment = fragment.parentFragment
+        }
+    }
+
+    // ViewModel was not found. Create a new one in the top-most parent.
+    var topParentFragment = parentFragment
```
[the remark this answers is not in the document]
BenSchwab: seems like `fragment` would be `null`, but if seems like we could eliminate this loop if we track parentFragment in the above loop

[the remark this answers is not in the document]
gpeal: @BenSchwab @elihart Yes, it would be null. I wrote it that way but the code was harder to read despite being shorter. Writing it this way made the intention of each section more explicit and easier to grok.

mvrx/src/main/kotlin/com/airbnb/mvrx/MvRxExtensions.kt
```diff
@@ -29,6 +29,55 @@ inline fun <T, reified VM : BaseMvRxViewModel<S>, reified S : MvRxState> T.fragm
...
+): Lazy<VM> where T : Fragment, T : MvRxView = lifecycleAwareLazy(this) {
+    requireNotNull(parentFragment) { "There is no parent fragment!" }
+    val factory = MvRxFactory { error("No ViewModel for this Fragment.") }
+    var fragment: Fragment? = parentFragment
+    val key = keyFactory()
+    while (fragment != null) {
+        try {
+            return@lifecycleAwareLazy ViewModelProviders.of(fragment, factory).get(key, viewModelClass.java)
+                .apply { subscribe(this@parentFragmentViewModel, subscriber = { postInvalidate() }) }
+        } catch (e: IllegalStateException) {
```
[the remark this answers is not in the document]
gpeal: @BenSchwab I did this by matching the exception name now. Otherwise, we'd have to expose a new type since this is an inline function

mvrx/src/main/kotlin/com/airbnb/mvrx/MvRxExtensions.kt
littleGnAl: @gpeal @elihart In our project, the `targetFragment` is generally used with `DialogFragment`, but MvRx does not officially support `DialogFragment`, I think this is not necessary.

 If someone needs to get the `ViewModel` of the `targetFragment `, they can explicitly cast the `targetFragment` such as `(targetFragment as MyTargetFragment).myTargetViewModel`. 

But if this function is needed, I think the logic should be similar to the logic of `parentFragmentViewModel`, because the `targetFragment`'s `ViewModel` maybe already exist.

gpeal: @littleGnAI, ah, the DialogFragment case makes sense to me

PR #248, opened 2019-06-24
Add Completable execute without the function to handle output

tasomaniac:
### Problem

Only when I have a `Completable` source, I couple of time had the condition where I had no way to handle `Async<Unit>` that comes with the callback from `Completable.execute`.

### Solution

The solution that I had locally is to have `Completable.subscribe().disposeOnClear()`. This is far from ideal. 

The solution that I'm proposing here is to have `execute` function overloaded for `Completable` with no arguments.

### Considerations

This is definitely going to be handy. But I was wondering what you may all think about it. It allows you to (very easily) skip/ignore any failures. Is that something you would be fine. At the end, it will be up to the user.

### DIFF
--- a/mvrx/src/main/kotlin/com/airbnb/mvrx/BaseMvRxViewModel.kt
+++ b/mvrx/src/main/kotlin/com/airbnb/mvrx/BaseMvRxViewModel.kt
@@ -156,14 +156,14 @@ abstract class BaseMvRxViewModel<S : MvRxState>(
     }
 
     /**
-     * Helper to map an Single to an Async property on the state object.
+     * Helper to map a Single to an Async property on the state object.
      */
     fun <T> Single<T>.execute(
         stateReducer: S.(Async<T>) -> S
     ) = toObservable().execute({ it }, null, stateReducer)
 
     /**
-     * Helper to map an Single to an Async property on the state object.
+     * Helper to map a Single to an Async property on the state object.
      * @param mapper A map converting the observable type to the desired AsyncData type.
      * @param stateReducer A reducer that is applied to the current state and should return the
      *                     new state. Because the state is the receiver and it likely a data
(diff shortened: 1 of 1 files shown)


mvrx/src/main/kotlin/com/airbnb/mvrx/BaseMvRxViewModel.kt
```diff
@@ -186,7 +186,12 @@ abstract class BaseMvRxViewModel<S : MvRxState>(
      */
     fun Completable.execute(
         stateReducer: S.(Async<Unit>) -> S
-    ) = toSingle { Unit }.execute(stateReducer)
+    ) = toObservable<Unit>().execute(stateReducer)
```
tasomaniac: Also changed `toSingle { Unit }` to `toObservable`. Single version already calls the same one.

gpeal:
@tasomaniac Can you explain the use case where you would want this in more detail? It seems specific enough that it could lead people to not properly handle error/loading states in other cases and the workaround if you really need this is fairly trivial.

tasomaniac:
This is useful when you want to trigger something that is not expected to error. Although now I think about it, potential errors in that case should throw rather then being ignored.

RxJava's subscribe with no argument exactly does that. It allows you to ignore emission and completion events but throws an exception in case an error happen. 

Would it be fine if I change it in a way that it is closer to RxJava's subscribe?

BenSchwab:
First, agreed that this should throw an exception if there is one. 

Second, I see the argument behind this change. But my main concern is that putting something in the BaseMvRxViewModel implicitly condones it as a MvRx pattern. For most use cases MvRx is of the opinion that you should model actions you take in your `state`. This API would encourage doing things that are not modeled, (even with a Unit return type, the fact that there is an outgoing Completable is often something you would want to know about, for example, to code defensive checks for duplicate completables).

@tasomaniac Do you have a base view model class in your application which extends the BaseMvRxViewModel and provide this function for you application? At Airbnb we use our own `MvRxViewModel` base class to provide extensions that are opinionated uses of MvRx which we don't feel are general enough to open source.

tasomaniac:
I was originally going to open an issue to ask your opinion on this but I thought it is even better with a PR. It makes a lot of sense to have this in our base ViewModel. Thanks

PR #249, opened 2019-06-24
Some RxJava clean up

tasomaniac:
I've seen this little piece that I can improve in `execute` function. Let me know what you think. 

- Use Kotlin's `also` to set `metadata` as a side-effect
- Use RxJava's `startWith(Loading())`. I believe it is more clear to read.

### DIFF
--- a/mvrx/src/main/kotlin/com/airbnb/mvrx/BaseMvRxViewModel.kt
+++ b/mvrx/src/main/kotlin/com/airbnb/mvrx/BaseMvRxViewModel.kt
@@ -207,13 +207,12 @@ abstract class BaseMvRxViewModel<S : MvRxState>(
         successMetaData: ((T) -> Any)? = null,
         stateReducer: S.(Async<V>) -> S
     ): Disposable {
-        setState { stateReducer(Loading()) }
-
-        return map { value ->
-            val success = Success(mapper(value))
-            success.metadata = successMetaData?.invoke(value)
-            success as Async<V>
+        return map<Async<V>> { value ->
+            Success(mapper(value)).also {
+                it.metadata = successMetaData?.invoke(value)
+            }
         }
+            .startWith(Loading())
             .onErrorReturn { Fail(it) }
             .subscribe { asyncData -> setState { stateReducer(asyncData) } }
             .disposeOnClear()


mvrx/src/main/kotlin/com/airbnb/mvrx/BaseMvRxViewModel.kt
```diff
@@ -207,13 +207,12 @@ abstract class BaseMvRxViewModel<S : MvRxState>(
...
-        return map { value ->
-            val success = Success(mapper(value))
-            success.metadata = successMetaData?.invoke(value)
-            success as Async<V>
+        return map<Async<V>> { value ->
+            Success(mapper(value)).also {
+                it.metadata = successMetaData?.invoke(value)
+            }
         }
+            .startWith(Loading())
```
gpeal: We originally had Loading done with a `startWith` but there was a specific timing reason why we switched it but it was in the git history before the GitHub migration. @BenSchwab do you remember or can you check git?

tasomaniac: Interesting. Loading starts when we subscribe to the RxJava type, not necessarily when the execute function is called.  Even if there is a timing issue, using startWith seems to be more correct.

Having said that, I'm really curious about what was the case before.

gpeal: @tasomaniac startWith will wait for the subscribe which will happen on the subcribeOn scheduler thread. That means that if you call `withState` right after `execute` then the loading reducer won't be enqueued yet if you use `startWith`

gpeal: Thanks @BenSchwab for digging up that original commit.

tasomaniac: Hmm, would it make sense to change this PR to have a comment about this? And keep the other clean up around `also`? Or should we just close it. I'm fine with both.

gpeal: A comment would definitely be useful. I don't personally find the also block more readable because it is more nested and it is less clear what exactly the type applies to

tasomaniac:
I will close this for now. Will open another smaller one in the future.

PR #254, opened 2019-07-29
Use SavedStateRegistry for persisting state

mmcwong:
This PR demonstrates how we could use the `SavedStateRegistry` to persist viewModel state due to process recreation. 

I believe this would allow us to remove the concept of the `MvRxViewModelStoreOwner` as we no longer need to hook into the activity, or fragment's `onSaveInstanceState` and `onCreate` methods to directly save and restore the viewModels. Instead, we can register a `SavedStateProvider` that directly hooks in to the lifecycle when we create the viewModel. When we go to recreate the viewModel, we check the saved state registry to see if there is a persisted state we can restore. 

The `restoreViewModel` method in `MvRxViewModelStoreOwner` helped ensure that all `existingViewModels` and `parentViewModels` were created before being accessed after system-initiated process death. Without the `MvRxViewModelStoreOwner` now, `existingViewModel` and `parentViewModel` could potentially access the viewModel before they have been recreated. This requires both these extensions to call into the modified `MvRxViewModelProvider.get` method and provide their own viewModel factories instead of using `ViewModelProviders.of`

Would love to get your thoughts on this @elihart @BenSchwab 
Thanks!


mvrx/src/main/kotlin/com/airbnb/mvrx/MvRxExtensions.kt
```diff
@@ -103,8 +112,7 @@ inline fun <T, reified VM : BaseMvRxViewModel<S>, reified S : MvRxState> T.activ
     viewModelClass: KClass<VM> = VM::class,
     noinline keyFactory: () -> String = { viewModelClass.java.name }
 ) where T : Fragment, T : MvRxView = lifecycleAwareLazy(this) {
-    if (requireActivity() !is MvRxViewModelStoreOwner) throw IllegalArgumentException("Your Activity must be a MvRxViewModelStoreOwner!")
-    MvRxViewModelProvider.get(viewModelClass.java, S::class.java, ActivityViewModelContext(requireActivity(), _activityArgsProvider(keyFactory)), keyFactory())
+    MvRxViewModelProvider.get(viewModelClass.java, S::class.java, ActivityViewModelContext(requireActivity(), _fragmentArgsProvider()), keyFactory())
```
elihart: put these on separate lines to match the above methods?

mvrx/src/main/kotlin/com/airbnb/mvrx/MvRxView.kt
```diff
@@ -22,7 +22,7 @@ private val handler = Handler(Looper.getMainLooper(), Handler.Callback { message
  * When you get a ViewModel with fragmentViewModel, activityViewModel, or existingViewModel, it
  * will automatically subscribe to all state changes in the ViewModel and call [invalidate].
  */
-interface MvRxView : MvRxViewModelStoreOwner, LifecycleOwner {
+interface MvRxView: LifecycleOwner {
```
elihart: nit: wrong spacing

mvrx/src/main/kotlin/com/airbnb/mvrx/MvRxViewModelProvider.kt
```diff
@@ -28,10 +30,26 @@ object MvRxViewModelProvider {
         stateClass: Class<out S>,
         viewModelContext: ViewModelContext,
         key: String = viewModelClass.name,
-        initialStateFactory: MvRxStateFactory<VM, S> = RealMvRxStateFactory()
+        initialStateFactory: MvRxStateFactory<VM, S> = RealMvRxStateFactory(),
```
elihart: does `initialStateFactory` need to be a parameter? seems like it is only used in the `viewModelFactory` default value

mvrx/src/main/kotlin/com/airbnb/mvrx/MvRxViewModelProvider.kt
```diff
@@ -28,10 +30,26 @@ object MvRxViewModelProvider {
...
+            val restoredArgs = savedStateBundle.get(KEY_MVRX_SAVED_ARGS)
+            val restoredState = savedStateBundle.getBundle(KEY_MVRX_SAVED_INSTANCE_STATE)
+
+            requireNotNull(restoredState) { "State was not saved prior to restoring!"}
+
+            val restoredContext = when (viewModelContext) {
+                is ActivityViewModelContext -> ActivityViewModelContext(viewModelContext.activity, restoredArgs)
+                is FragmentViewModelContext -> FragmentViewModelContext(viewModelContext.activity, restoredArgs, viewModelContext.fragment)
+            }
+            MvRxFactory { createViewModel(key, viewModelClass, stateClass, restoredContext, restoredState::restorePersistedState) }
```
elihart: this pattern `MvRxFactory { createViewModel` seems to be used several times, there could be a function or constructor to simplify it to just do `createMvRxFactory(key, viewModelClass, stateClass...)`

mvrx/src/main/kotlin/com/airbnb/mvrx/MvRxViewModelStore.kt
```diff
@@ -105,8 +105,8 @@ class MvRxViewModelStore(private val viewModelStore: ViewModelStore) {
             val holder = viewModelsState.getParcelable<MvRxPersistedViewModelHolder>(key)
                 ?: throw IllegalStateException("ViewModel key: $key expected to be in stored ViewModels but was not found.")
             map[key] = when (host) {
-                is Fragment -> restoreViewModel(host, holder, arguments)
-                is FragmentActivity -> restoreViewModel(host, holder, arguments)
+                is Fragment -> restoreViewModel(key, host, holder, arguments)
```
elihart: I think it would be great to delete this class if we can now (and move these helper functions into a file just for dealing with saved state.

mmcwong: I'm able to actually delete this class and `MvRxViewModelStoreOwner` as well!

mvrx/src/main/kotlin/com/airbnb/mvrx/MvRxViewModelProvider.kt
mmcwong: this was actually added in https://github.com/airbnb/MvRx/pull/219 as we override this in the Airbnb app for mocking

mmcwong:
@elihart I currently have it so that we're always persisting both the state and the initial args of the viewModel when saving to the bundle. This removes having to special case for saving fragment args in activityViewModels.

mvrx/src/main/kotlin/com/airbnb/mvrx/MvRxViewModelProvider.kt
```diff
@@ -70,6 +88,17 @@ object MvRxViewModelProvider {
                     "single non-optional parameter that takes initial state of ${stateClass.simpleName}."
             }
         }
+
+        viewModelContext.savedStateRegistry.registerSavedStateProvider(key) {
+            withState(viewModel) { state ->
+                Bundle().apply {
+                    putBundle(KEY_MVRX_SAVED_INSTANCE_STATE, state.persistState())
+                    putParcelable(KEY_MVRX_SAVED_ARGS, viewModelContext.args as Parcelable?)
```
mmcwong: there isn't a specific interface that the arguments conform to, is there a way to store in the bundle this without switching on the type?

mvrx/src/main/kotlin/com/airbnb/mvrx/MvRxViewModelProvider.kt
elihart: Right, we need to make sure we can override the necessary behavior or mocking, but we also want a clean interface here.

As far as I can tell the only times we override `viewModelFactory` are when we want to throw if the view model doesn't already exist - ie factory that throws when it is used instead of actually doing anything. perhaps we don't need to allow `viewModelFactory` to be set, and instead this can have a boolean parameter for whether it should throw if the viewmodel doesn't already exist

we should try to have a clean interface here and not need both `initialStateFactory` and `viewModelFactory`, but we do need to design it in a way that works for our mocking (ie it needs to be overridable

mvrx/src/main/kotlin/com/airbnb/mvrx/MvRxViewModelProvider.kt
elihart: I guess before we were treating it as a serializable.

I _think_ it is safe to treat it as either a serializable or a parcelable, so you could use a `when` statement to check for those two cases (primitives, lists, and strings would fall under the Serializable case), and I suppose it would be best to throw in the else case so we are aware of issues

mmcwong:
@elihart updated with your feedback! Sorry for the delay, I restructured my code a bit to check the saved state registry in the `create` method of the `MvRxFactory` so that we are registering/consuming the state in one place that is attached to the lifecycle of the viewModel. I think in my initial implementation I was registering it only when the viewModel was created, but consuming it each time `MvRxViewModelProvider.get` was called which is tied to the fragment/activity creation instead. I also updated/added some tests.

mvrx/src/main/kotlin/com/airbnb/mvrx/BaseMvRxActivity.kt
```diff
@@ -1,41 +1,18 @@
...
-        super.onCreate(savedInstanceState)
-    }
-
-    override fun onSaveInstanceState(outState: Bundle) {
-        super.onSaveInstanceState(outState)
-        mvrxViewModelStore.saveViewModels(outState)
-    }
-}
\ No newline at end of file
+abstract class BaseMvRxActivity : AppCompatActivity()
```
mmcwong: do we still need this class? Maybe we'll need it in the future?

mvrx/src/main/kotlin/com/airbnb/mvrx/MvRxFactory.kt
```diff
@@ -1,10 +1,113 @@
...
+            val restoredArgs = savedStateBundle.get(KEY_MVRX_SAVED_ARGS)
+            val restoredState = savedStateBundle.getBundle(KEY_MVRX_SAVED_INSTANCE_STATE)
+
+            requireNotNull(restoredState) { "State was not saved prior to restoring!" }
+
+            val restoredContext = when (viewModelContext) {
+                is ActivityViewModelContext -> ActivityViewModelContext(viewModelContext.activity, restoredArgs)
+                is FragmentViewModelContext -> FragmentViewModelContext(viewModelContext.activity, restoredArgs, viewModelContext.fragment)
+            }
+            restoredContext to createViewModel(viewModelClass, stateClass, restoredContext, key, restoredState::restorePersistedState)
```
elihart: the `createViewModel` code looks mostly the same. to remove some duplication, could this return `ViewModelContext` and the state factory? then we can call `createViewModel` in just one place

mvrx/src/main/kotlin/com/airbnb/mvrx/MvRxFactory.kt
```diff
@@ -1,10 +1,113 @@
...
+    private val viewModelContext: ViewModelContext,
+    private val key: String,
+    private val forExistingViewModel: Boolean,
+    private val initialStateFactory: MvRxStateFactory<VM, S>
+) : ViewModelProvider.Factory {
+
+    @Suppress("UNCHECKED_CAST")
+    override fun <T : ViewModel?> create(modelClass: Class<T>): T {
+        val (viewModelContext, viewModel) = viewModelContext.savedStateRegistry.consumeRestoredStateForKey(key)?.let { savedStateBundle ->
+            val restoredArgs = savedStateBundle.get(KEY_MVRX_SAVED_ARGS)
```
elihart: might be nice to pull the code in this `let` block into a separate function, since this create function is a bit large and complex.

mvrx/src/main/kotlin/com/airbnb/mvrx/MvRxFactory.kt
```diff
@@ -1,10 +1,113 @@
...
+            restoredContext to createViewModel(viewModelClass, stateClass, restoredContext, key, restoredState::restorePersistedState)
+        } ?: viewModelContext to createViewModel(viewModelClass, stateClass, viewModelContext, key, forExistingViewModel = forExistingViewModel, initialStateFactory = initialStateFactory)
+
+        viewModelContext.savedStateRegistry.registerSavedStateProvider(key) {
+            withState(viewModel) { state ->
+                Bundle().apply {
+                    putBundle(KEY_MVRX_SAVED_INSTANCE_STATE, state.persistState())
+                    when (val args = viewModelContext.args) {
+                        is Parcelable -> putParcelable(KEY_MVRX_SAVED_ARGS, args)
+                        is Serializable -> putSerializable(KEY_MVRX_SAVED_ARGS, args)
```
elihart: can you throw in the else case? better to fail and detect it, then fail without knowing I think

mvrx/src/main/kotlin/com/airbnb/mvrx/MvRxFactory.kt
```diff
@@ -1,10 +1,139 @@
...
+private fun <VM : BaseMvRxViewModel<S>, S : MvRxState> createViewModel(
+    viewModelClass: Class<out VM>,
+    stateClass: Class<out S>,
+    viewModelContext: ViewModelContext,
+    key: String,
+    stateRestorer: (S) -> S,
+    forExistingViewModel: Boolean,
+    initialStateFactory: MvRxStateFactory<VM, S>
+): VM {
+    if (forExistingViewModel) {
```
BenSchwab: I feel like it is a little odd to have a parameter which if true throws an exception -- could the caller perform this check? This keeps the function more true to it's name `createViewModel`.

mmcwong:
Thanks for reviewing! I've removed the `BaseMvRxActivity` and addressed PR feedback, but other than that I didn't have anything else to add to this PR.

MatushkaRossiya:
New versions of Lifecycle (2.2.0-alpha03) and Lifecycle ViewModel SavedState (1.0.0-alpha03) were just released, this may be relevant to this PR.

elihart:
I don't think we've decided completely, but imo we a lifecycle beta release would be a reasonable time to merge this. I think that alpha would be too early though.

gpeal:
@elihart @mradzinski Agreed. We might want to consider beta but I would be cautious to require that for any users of MvRx.

There is an official definition of alpha/beta/rc here: https://developer.android.com/jetpack/androidx/versions

mradzinski:
Checking this again, is the latest alpha version of Lifecycle truly needed or it was updated per request? I think changes contained in this PR are available in the latest stable version of Google's library as well, so I would actually revert to stable and leave updating that library for another PR.

mmcwong:
A newer version of `androidx.fragment` (stable release is currently `1.0.0`) is needed to reference the `SavedStateRegistry` from the fragment/activity. `androidx.fragment:fragment:1.1.0-rc04` requires a lifecycle version of `2.2.0-alpha02` which is why I bumped this version in this PR.

elihart:
A stable release of `androidx.fragment` 1.1.0 is out - can we update to that and then be able to merge this? https://developer.android.com/jetpack/androidx/releases/fragment?authuser=2#1.1.0

It seems strange that a stable release of androidx.fragment would require an alpha version of the lifecycle library

mradzinski:
I agree with @elihart. Here's what it looks to be the `build.gradle` file of the latest stable version (1.1.0) of androidx.fragments: https://android.googlesource.com/platform/frameworks/support/+/refs/heads/androidx-master-dev/fragment/fragment/build.gradle

gpeal:
@mradzinski @elihart I wonder if it was an issue with the code not syncing. I upgraded in [this pr](https://github.com/airbnb/MvRx/pull/285) and this is the compile classpath
https://gist.github.com/gpeal/ef77f349eb90bc69b7518a94390609a0

gpeal:
@mmcwong you can now rebase on https://github.com/airbnb/MvRx/pull/285

mvrx/src/main/kotlin/com/airbnb/mvrx/BaseMvRxFragment.kt
```diff
@@ -11,6 +11,14 @@ import androidx.lifecycle.LifecycleOwner
  */
 abstract class BaseMvRxFragment : Fragment(), MvRxView {
 
+<<<<<<< HEAD
```
gpeal: @mmcwong Looks like you accidentally left some merge conflicts in

mmcwong: whoops!

mvrx/src/main/kotlin/com/airbnb/mvrx/MvRxFactory.kt
```diff
@@ -1,10 +1,114 @@
...
+    if (viewModelClass.constructors.size == 1) {
+        val primaryConstructor = viewModelClass.constructors[0]
+        if (primaryConstructor.parameterTypes.size == 1 && primaryConstructor.parameterTypes[0].isAssignableFrom(state::class.java)) {
+            return primaryConstructor?.newInstance(state) as? VM
+        }
+    }
+    return null
+}
+
+private fun <S : MvRxState> Bundle.toRestoredState(viewModelContext: ViewModelContext): Pair<ViewModelContext, (S) -> S> {
```
elihart: maybe a clearer name would be `toStateRestorer`? Might even be worth creating a data class to hold these return values in instead of a Pair for clarity

mvrx/src/main/kotlin/com/airbnb/mvrx/MvRxViewModelFactory.kt
```diff
@@ -78,23 +83,29 @@ sealed class ViewModelContext {
  * reference is available when an activity scoped ViewModel is first created, during process restoration, activity scoped ViewModels will be created
  * _without_ a fragment reference, so it is only safe to reference the activity.
  */
-class ActivityViewModelContext(
+data class ActivityViewModelContext(
     override val activity: FragmentActivity,
     override val args: Any?
-) : ViewModelContext()
+) : ViewModelContext() {
+    override val owner = activity
```
elihart: you can use `get() = `for all of the assignments you added in this file to avoid object allocations. We don't really need to store the separately, and is nice to minimize memory that our library uses for people

mvrx/src/main/kotlin/com/airbnb/mvrx/MvRxViewModelProvider.kt
```diff
@@ -22,68 +25,48 @@ object MvRxViewModelProvider {
...
-                return primaryConstructor?.newInstance(state) as? VM
+    private fun <VM : BaseMvRxViewModel<S>, S : MvRxState> VM.getSavedStateBundle(viewModelContext: ViewModelContext) =
+        withState(this) { state ->
+            Bundle().apply {
+                putBundle(KEY_MVRX_SAVED_INSTANCE_STATE, state.persistState())
+                viewModelContext.args?.let {
+                    when (it) {
+                        is Parcelable -> putParcelable(KEY_MVRX_SAVED_ARGS, it)
+                        is Serializable -> putSerializable(KEY_MVRX_SAVED_ARGS, it)
+                        else -> throw IllegalStateException("Args must be parcelable or serializable")
```
elihart: can do `error("Args must be parcelable or serializable")` instead to be shorter

mvrx/src/main/kotlin/com/airbnb/mvrx/MvRxExtensions.kt
```diff
@@ -117,36 +126,19 @@ inline fun <T, reified VM : BaseMvRxViewModel<S>, reified S : MvRxState> T.activ
...
- * For internal use only. Public for inline.
- *
- * Looks for [MvRx.KEY_ARG] on the arguments of the fragment receiver.
- *
- * Also adds the fragment's MvRx args to the host Activity's [MvRxViewModelStore] so that they can be used to recreate initial state
- * in a new process.
- */
-@Suppress("FunctionName")
-@RestrictTo(LIBRARY)
-inline fun <T : Fragment> T._activityArgsProvider(keyFactory: () -> String): Any? {
```
elihart: I'm wondering about removing this now, are you sure that the functionality is made up somewhere else? I think we may be missing something.

Below you have 
```
FragmentViewModelContext -> viewModelContext.copy(args = restoredArgs)
```

which never happened before - the restored arguments were only used in activity context. 

The point of this activity args approach is to identify the fragment that created an "activityViewModel", save those arguments, and if another fragment restores that view model it can use the correct arguments from a different fragment.

I don't think this current approach lets us do that. The arguments from the first fragment get overridden, since when saving to the bundle we use `viewModelContext.args` of the last fragment to access that viewmodel and register the saved state registry.

A potential fix might be to add an internal property on BaseMvRxViewModel that stores the arguments of the fragment used to create it. Then in `getSavedStateBundle` we would save those arguments instead of the args from the context

In our app we should be able to test if this works or not by going through a flow to a page that has an "existingViewModel", where the viewmodel was initialized by arguments from the first screen. Then use debug settings to force the process to be killed when you switch apps (set it to keep 0 processes). Coming back to the app should restore the flow correctly, but right now I believe it would crash

build.gradle
```diff
@@ -2,7 +2,9 @@ buildscript {
     ext.kotlinVersion = '1.3.50'
     ext.appCompatVersion = '1.1.0'
     ext.recyclerViewVersion = '1.0.0'
+    ext.constraintLayoutVersion = '1.1.3'
     ext.lifecycleVersion = '2.1.0'
+    ext.coreVersion = '1.1.0'
```
elihart: core is a generic name here. this could probably be `androidXCoreKtx` (lots of libraries have "core" in the module names"

mvrx/src/main/kotlin/com/airbnb/mvrx/MvRxViewModelProvider.kt
```diff
@@ -22,68 +26,81 @@ object MvRxViewModelProvider {
...
+        ).get(key, viewModelClass)
+
+        try {
+            // Save the view model's state to the bundle so that it can be used to recreate
+            // state across system initiated process death.
+            viewModelContext.savedStateRegistry.registerSavedStateProvider(key) {
+                viewModel.getSavedStateBundle(restoredContext.args)
             }
+        } catch (e: IllegalArgumentException) {
+            // The view model was already registered with the context.
```
elihart: can you add some more details here about why this is expected and ok? and about how we need to ensure that only the initial fragment that created the viewmodel is used to save state?

mvrx/src/main/kotlin/com/airbnb/mvrx/MvRxViewModelProvider.kt
```diff
@@ -22,68 +26,81 @@ object MvRxViewModelProvider {
...
     ): VM {
-        // This wraps the fact that ViewModelProvider.of has individual methods for Fragment and FragmentActivity.
-        val factory = MvRxFactory { createViewModel(viewModelClass, stateClass, viewModelContext, initialStateFactory = initialStateFactory) }
-        return when (viewModelContext) {
-            is ActivityViewModelContext -> ViewModelProviders.of(viewModelContext.activity, factory)
-            is FragmentViewModelContext -> ViewModelProviders.of(viewModelContext.fragment, factory)
-        }.get(key, viewModelClass)
-    }
+        val stateRestorer = viewModelContext
+            .savedStateRegistry
```
elihart: did you address the case where this is called before onCreate and savedStateRegistry doesn't have the restored state yet?

mmcwong: I can check if `savedStateRegistroy.isRestored` and if not throw a custom error message (it currently already throws an exception but the error message makes a little less sense to users who don't know about the saved state registry). I also added a test for this case.

mvrx/src/main/kotlin/com/airbnb/mvrx/MvRxViewModelProvider.kt
```diff
@@ -39,8 +39,13 @@ object MvRxViewModelProvider {
         forExistingViewModel: Boolean = false,
         initialStateFactory: MvRxStateFactory<VM, S> = RealMvRxStateFactory()
     ): VM {
-        val stateRestorer = viewModelContext
-            .savedStateRegistry
+        val savedStateRegistry = viewModelContext.savedStateRegistry
+
+        if (!savedStateRegistry.isRestored) {
+            error("You can only access a view model after onCreate of your component has been called.")
```
elihart: maybe  change `component` to `Fragment`? could also specify that only `super.onCreate` needs to be called, viewmodels can be accessed in onCreate after the super call.

@BenSchwab how do you feel about  this new restriction? Previously it would work to access viewmodels after super.onAttach of the fragment, but I believe that doing so is currently broken, since the viewmodel accessed then can be replaced in the viewmodel store by another viewmodel that is created during the state restoration in onCreate.

So in practice  this restriction always existed and this enforces it, and any violations are currently bugs

mvrx/src/test/kotlin/com/airbnb/mvrx/MvRxViewModelProviderTest.kt
```diff
@@ -85,6 +85,12 @@ class MvRxViewModelProviderTest : BaseTest() {
         }
     }
 
+    @Test(expected = IllegalStateException::class)
+    fun failForAccessingViewModelBeforeOnCreate() {
+        val activityWithoutSetup = Robolectric.buildActivity(FragmentActivity::class.java).get()
```
elihart: ideally this would test the error message too, since we know this will  fail for some other reason at least, but we want to make sure it fails with your custom error message

mvrx/src/main/kotlin/com/airbnb/mvrx/MvRxViewModelProvider.kt
mmcwong: This could also apply for `Activities` as well? I update this to be more specific.

mmcwong:
@BenSchwab could you take another look at this before I merge this change? Namely the restriction around accessing the view model after the `super.onCreate` method

mvrx/src/main/kotlin/com/airbnb/mvrx/MvRxViewModelProvider.kt
```diff
@@ -22,68 +26,88 @@ object MvRxViewModelProvider {
...
+                stateRestorer?.toRestoredState,
+                forExistingViewModel,
+                initialStateFactory
+            )
+        ).get(key, viewModelClass)
+
+        try {
+            // Save the view model's state to the bundle so that it can be used to recreate
+            // state across system initiated process death.
+            viewModelContext.savedStateRegistry.registerSavedStateProvider(key) {
```
BenSchwab: Mostly curious, could you first call consume, and then call register to avoid the the try catch? Fine with this approach, though.

mmcwong: Calling consume just tries to retrieve the bundle from the saved state, we would need to call `unregisterSavedStateProvider` instead. Calling unregister each time before calling register could work but would require storing the args which initialized the activity viewModel somewhere. This is because within a flow, using `existingViewModel` would unregister the fragment that initially created the view model and reregister the current fragment instead. We would no longer have access to the args that created the view model when saving to the bundle which is required when recreating the state after process death.

PR #255, opened 2019-07-29
Make minSdk 21 in all samples to fix #199

### LINKED ISSUE #199: Sample App has issue with TLS1.2 on <= Android 4.4
On Android 4.1.x and up until 4.4 TLS1.2 is not switched on by default. So if you run Sample App on Android 4.4 (try it on emulator) and below you can see that url https://icanhazdadjoke.com/ throw error due to TLS1.2 support issue. 

Retrofit2 using OkHttpClient and some suggestions how to resolve it described [here](https://github.com/square/okhttp/issues/2372). The easiest way is to use Google Play Services together with `ProviderInstaller.installIfNeeded` but it will force to apply MultiDex because Sample App already close to 65,536 methods. Another way is to provide custom SocketFactory.

In addition you didn't handle onFail in subscriber for RandomDadJokeState, so there no any errors showing in UI if request is failed.

tasomaniac:
Make minSdk 21 in all samples 

The other samples were already 21. `mvrx` library dependencies have `minSdkVersion = 16`

Also clean-up unneeded declarations and unneeded dependencies in sample gradle configurations for simplicity. Sorry for putting all this together but I think they're good clean-up. Run all the samples and verified that they work well.

Fixes #199

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -11,7 +11,6 @@ buildscript {
     ext.navVersion = '2.0.0'
     ext.roomVersion = '2.0.0'
     ext.buildToolsVersion = '28.0.3'
-    ext.androidXTestVersion = '1.1.0'
     ext.espressoVersion = '3.1.1'
 
     repositories {
--- a/counter/build.gradle
+++ b/counter/build.gradle
@@ -1,6 +1,7 @@
 apply plugin: 'com.android.application'
 apply plugin: 'kotlin-android'
 apply plugin: 'kotlin-android-extensions'
+
 android {
     compileSdkVersion 28
 
@@ -10,12 +11,6 @@ android {
         targetSdkVersion 28
         versionCode 1
         versionName "1.0"
-    }
-
-    buildTypes {
-        release {
-            minifyEnabled false
-        }
     }
 }
 
--- a/sample/build.gradle
+++ b/sample/build.gradle
@@ -7,7 +7,7 @@ android {
     compileSdkVersion 28
     defaultConfig {
         applicationId "com.airbnb.sample"
-        minSdkVersion 16
+        minSdkVersion 21
         targetSdkVersion 28
         versionCode 1
         versionName "0.0.1"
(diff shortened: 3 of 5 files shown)


build.gradle
```diff
@@ -11,7 +11,6 @@ buildscript {
     ext.navVersion = '2.0.0'
     ext.roomVersion = '2.0.0'
     ext.buildToolsVersion = '28.0.3'
-    ext.androidXTestVersion = '1.1.0'
```
tasomaniac: was never used

counter/build.gradle
```diff
@@ -11,12 +12,6 @@ android {
         versionCode 1
         versionName "1.0"
     }
-
-    buildTypes {
-        release {
-            minifyEnabled false
```
tasomaniac: by default, it is false already.

sample/build.gradle
```diff
@@ -32,9 +32,8 @@ dependencies {
     implementation project(':mvrx')
     implementation "org.jetbrains.kotlin:kotlin-stdlib:$kotlinVersion"
     implementation "org.jetbrains.kotlin:kotlin-stdlib-jdk8:$kotlinVersion"
-    // TODO: clean up the support library dependency resolution.
```
tasomaniac: After AndroidX migration, just removing TODO here and exclude works fine.

todomvrx/build.gradle
```diff
@@ -7,26 +7,14 @@ android {
...
         versionCode 1
         versionName "0.0.1"
 
         testInstrumentationRunner "androidx.test.runner.AndroidJUnitRunner"
     }
     buildToolsVersion "$buildToolsVersion"
-
-    android {
-        sourceSets {
-            String sharedTestDir = 'src/sharedTest/java'
```
tasomaniac: Probably a copy paste. There is no `sharedTest` folder.

todomvrx/build.gradle
```diff
@@ -45,12 +33,6 @@ dependencies {
     kapt "com.airbnb.android:epoxy-processor:$epoxyVersion"
     implementation 'io.reactivex.rxjava2:rxandroid:2.1.1'
     implementation "io.reactivex.rxjava2:rxjava:2.2.8"
-    implementation "com.squareup.retrofit2:retrofit:$retrofitVersion"
```
tasomaniac: Also copy paste.

tasomaniac:
Glad to help 🖖

PR #258, opened 2019-08-01
Fix a uniqueId related problem in unit tests

tasomaniac:
### Problem

Using `uniqueOnly()` in a class that is unit tested crashes because of lateinit not being initialized. In unit tests, `onCreate` is never called. That means that the lateinit will never be instantiated. 

### Solution

Instead, make `persistedViewId` only contain the persisted id. That means it is nullable now. Initially there is no persisted id and it makes sense. 

Then when `mvrxId` is accessed for the first time, we generate the id. And later if the user changes orientation, it should be persisted. 

Functionality wise, this change should behave 100% the same. But it fixes a problem with unit testing view layer.

### DIFF
--- a/mvrx/src/main/kotlin/com/airbnb/mvrx/MvRxViewId.kt
+++ b/mvrx/src/main/kotlin/com/airbnb/mvrx/MvRxViewId.kt
@@ -1,0 +1,38 @@
+package com.airbnb.mvrx
+
+import android.os.Bundle
+import java.util.*
+import kotlin.properties.ReadOnlyProperty
+import kotlin.reflect.KProperty
+
+class MvRxViewId : ReadOnlyProperty<MvRxView, String> {
+
+    private var value: String? = null
+
+    override fun getValue(thisRef: MvRxView, property: KProperty<*>): String {
+        val currentValue = value
+        return if (currentValue == null) {
+            val id = generateUniqueId(thisRef)
+            value = id
+            id
+        } else {
+            currentValue
+        }
+    }
+
+    private fun generateUniqueId(thisRef: MvRxView) = thisRef::class.java.simpleName + "_" + UUID.randomUUID().toString()
+
+    fun saveTo(bundle: Bundle) {
+        bundle.putString(PERSISTED_VIEW_ID_KEY, value)
+    }
+
+    fun restoreFrom(bundle: Bundle?) {
+        if (value == null) {
+            value = bundle?.getString(PERSISTED_VIEW_ID_KEY)
+        }
+    }
+
+    companion object {
+        private const val PERSISTED_VIEW_ID_KEY = "mvrx:persisted_view_id"
+    }
+}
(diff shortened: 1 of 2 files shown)


elihart:
Thanks for putting this up. This looks fine, but it also made me think that we're putting a fair amount of complexity into this base class, which people would have to copy if they want a custom base class or for other implementations of MvRxView.

It would be nice instead to encapsulate this as much as possible, both for reuse and to make for easy base class implementations. I came up with something like this
```
abstract class BaseMvRxFragment : Fragment(), MvRxView {

    override val mvrxViewId: ViewId = ViewId()

        mvrxViewId.restoreFrom(savedInstanceState)
    }

    override fun onSaveInstanceState(outState: Bundle) {
        super.onSaveInstanceState(outState)
        mvrxViewId.saveTo(outState)
    }
}


class ViewId : ReadOnlyProperty<Any, String> {

    private var value: String? = null

    override fun getValue(thisRef: Any, property: KProperty<*>): String {
        val currentValue = value

        return if (currentValue == null) {
            val id = generateUniqueId(thisRef)
            value = id
            id
        } else {
            currentValue
        }
    }

    private fun generateUniqueId(thisRef: Any) = thisRef::class.java.simpleName + "_" + UUID.randomUUID().toString()

    fun saveTo(bundle: Bundle) {
        bundle.putString(PERSISTED_VIEW_ID_KEY, value)
    }

    fun restoreFrom(bundle: Bundle?) {
        if (value == null) {
            value = bundle?.getString(PERSISTED_VIEW_ID_KEY)
        }
    }

    companion object {
        private const val PERSISTED_VIEW_ID_KEY = "mvrx:persisted_view_id"
    }
}
```

tasomaniac:
This looks really good. Actually if you do `by ViewId()` I don't think nobody needs to change their base. This can be backward compatible. 

I makes total sense to keep Base implementation smaller. I will update this soon.

tasomaniac:
@elihart please have another look. It was done in a backward compatible way. I think it is still also something people can copy paste from `BaseMvRxFragment`. This also made `BaseMvRxFragment` so much simpler.

mvrx/src/main/kotlin/com/airbnb/mvrx/MvRxViewId.kt
```diff
@@ -0,0 +1,38 @@
...
+import java.util.*
+import kotlin.properties.ReadOnlyProperty
+import kotlin.reflect.KProperty
+
+class MvRxViewId : ReadOnlyProperty<MvRxView, String> {
+
+    private var value: String? = null
+
+    override fun getValue(thisRef: MvRxView, property: KProperty<*>): String {
+        val currentValue = value
```
tasomaniac: I know you all don't like using Kotlin scoping functions that much but this can be actually simplified like below. I was gonna send that but then decided to ask first. Let me know what you think

```
return value ?: generateUniqueId(thisRef).also { value = it }
```

elihart: sure, I like that one liner :)

tasomaniac: Applied.

tasomaniac:
I see what you mean. In certain cases, I can see that it is perfectly viable to have constant (or pre-computed) `viewId` for custom `MvRxView` implementations. That's why the current implementation of `MvRxViewId` (that includes save/restore logic etc) may be overly complex.

The intention of this PR was to fix a small problem with unit tests. 😄 I would be much more comfortable by not changing public API

elihart:
Ah sure, completely understand :D this is great for now, thanks again!

PR #263, opened 2019-08-14
Small fixes on comments

tasomaniac:
Small fixes on comments on functions available in `BaseMvRxViewModel`.


mvrx/src/main/kotlin/com/airbnb/mvrx/BaseMvRxViewModel.kt
```diff
@@ -136,7 +136,7 @@ abstract class BaseMvRxViewModel<S : MvRxState>(
 
     /**
      * Access the current ViewModel state. Takes a block of code that will be run after all current pending state
-     * updates are processed. The `this` inside of the block is the state.
```
tasomaniac: This was just wrong :) The block gets the state in a parameter.

PR #264, opened 2019-08-15
Warm reflection only for fields with @PersistState annotation

### LINKED ISSUE #253: ClassNotFoundException on 1.0.2
Hi there,

We have recently updated mvrx dependency from `1.0.1` to `1.0.2`. Which has messed up lots of our classes on our release builds with proguard. We are basically getting `java.lang.ClassNotFoundException` all over the app(data classes, enums, Async from mvrx, etc..) 

Reverting to 1.0.1 fixes the problem

tasomaniac:
This is to try fix #253 

Basically it runs warming the Kotlin reflect only on the fields with `PersistState` annotation.

I actually haven't verified this. I wonder how we can do that 🤔

### DIFF
--- a/mvrx/src/main/kotlin/com/airbnb/mvrx/BaseMvRxViewModel.kt
+++ b/mvrx/src/main/kotlin/com/airbnb/mvrx/BaseMvRxViewModel.kt
@@ -19,6 +19,7 @@ import java.util.concurrent.ConcurrentHashMap
 import kotlin.reflect.KProperty1
 import kotlin.reflect.KVisibility
 import kotlin.reflect.full.declaredMemberProperties
+import kotlin.reflect.full.findAnnotation
 import kotlin.reflect.full.memberProperties
 import kotlin.reflect.full.primaryConstructor
 import kotlin.reflect.jvm.isAccessible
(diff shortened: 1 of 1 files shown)


MatushkaRossiya:
While the actual issue has been fixed in a different (proper) way, wouldn't this PR be still useful as a small performance optimization, or do I misunderstand what's happening here?

tasomaniac:
I believe this change is still helpful. Although this is done in background thread, it is still a heavy operation. In our project, we don't even use PersistState, so warming reflection for all fields is an overhead for us.

mvrx/src/main/kotlin/com/airbnb/mvrx/BaseMvRxViewModel.kt
```diff
@@ -70,8 +71,10 @@ abstract class BaseMvRxViewModel<S : MvRxState>(
         initialState::class.declaredMemberProperties.asSequence()
             .filter { it.visibility == KVisibility.PUBLIC }
             .forEach { prop ->
-                @Suppress("UNCHECKED_CAST")
-                (prop as? KProperty1<S, Any?>)?.get(initialState)
+                if (prop.findAnnotation<PersistState>() != null) {
```
gpeal: @tasomaniac Does this actually improve performance? Finding annotations might be as slow as just accessing the field.

tasomaniac: This is my naive approach. I didn't know that findAnnotation would be expensive. In our project we don't use PersistState at all so I was wondering how can we avoid warming up reflection since we don't really need it.

gpeal: @tasomaniac You could measure it by sticking in some logs in a class that has a lot of both PersistState and not PersistState properties.

tasomaniac: So I tried to do that but couldn't find a noticable difference.

While doing that I also realized that we load all annotations for all fields. So finding the annotation is instant. 

It's up to you to merge this. I don't think it would make it worse.

PR #266, opened 2019-08-15
Fix shrinking with cache warming

rossbacher:
During reflection cache warming also the types are accessed via reflection. Need to keep them too.

This should fix https://github.com/airbnb/MvRx/issues/253

### DIFF
--- a/mvrx/proguard-rules.pro
+++ b/mvrx/proguard-rules.pro
@@ -34,6 +34,7 @@
 
 # Members of the Kotlin data classes used as the state in MvRx are read via Kotlin reflection which cause trouble
 # with Proguard if they are not kept.
--keepclassmembers class ** implements com.airbnb.mvrx.MvRxState {
+# During reflection cache warming also the types are accessed via reflection. Need to keep them too.
+-keepclassmembers,includedescriptorclasses class ** implements com.airbnb.mvrx.MvRxState {
    *;
 }


mvrx/proguard-rules.pro
```diff
@@ -34,6 +34,7 @@
 
 # Members of the Kotlin data classes used as the state in MvRx are read via Kotlin reflection which cause trouble
 # with Proguard if they are not kept.
--keepclassmembers class ** implements com.airbnb.mvrx.MvRxState {
+# During reflection cache warming also the types are accessed via reflection. Need to keep them too.
+-keepclassmembers,includedescriptorclasses class ** implements com.airbnb.mvrx.MvRxState {
```
tasomaniac: First time I see descriptorclasses modifier. Is that the magic comes from? 😊

rossbacher: Yes, read: 

> **includedescriptorclasses**
> Specifies that any classes in the type descriptors of the methods and fields that the -keep option keeps should be kept as well. This is typically useful when keeping native method names, to make sure that the parameter types of native methods aren't renamed either. Their signatures then remain completely unchanged and compatible with the native libraries.

from: https://www.guardsquare.com/en/products/proguard/manual/usage

and keeping the type is exactly what is needed here.

rossbacher:
Replaced with https://github.com/airbnb/MvRx/pull/267

PR #271, opened 2019-08-17
Update MvRx sample module to Koin 2.0

### LINKED ISSUE #270: Update MvRx sample to use Koin 2.0
The sample application shipped with MvRx uses Koin 0.9.3, and the older API that comes with it. Would it be okay if I create a PR migrating the sample to Koin 2.0?

haroldadmin:
Fixes #270 

- Remove old Koin artifacts, and replace them with Koin 2.0 equivalents.
- Update MvRx sample module to use Koin 2.0
- Remove unused Koin dependency from Todo-MvRx Sample.

### DIFF
--- a/todomvrx/build.gradle
+++ b/todomvrx/build.gradle
@@ -33,7 +33,6 @@ dependencies {
     kapt "com.airbnb.android:epoxy-processor:$epoxyVersion"
     implementation 'io.reactivex.rxjava2:rxandroid:2.1.1'
     implementation "io.reactivex.rxjava2:rxjava:2.2.8"
-    implementation "org.koin:koin-android-architecture:$koinVersion"
     implementation "androidx.navigation:navigation-fragment-ktx:$navVersion"
     implementation "androidx.navigation:navigation-ui-ktx:$navVersion"
     implementation "androidx.room:room-runtime:$roomVersion"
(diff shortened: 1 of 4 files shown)


todomvrx/build.gradle
```diff
@@ -33,7 +33,6 @@ dependencies {
     kapt "com.airbnb.android:epoxy-processor:$epoxyVersion"
     implementation 'io.reactivex.rxjava2:rxandroid:2.1.1'
     implementation "io.reactivex.rxjava2:rxjava:2.2.8"
-    implementation "org.koin:koin-android-architecture:$koinVersion"
```
corjohnson: Does this need to add the implementation of koin-android? or no?

haroldadmin: What's the point of not using Android specific Koin implementation in an Android app?

corjohnson: I was just wondering if something needed to replace the deleted line in the Gradle file. If not, then no big deal 🙂

haroldadmin: I think the Koin folks stopped shipping the `org.koin:koin-android-architecture` artifact, which is why I removed it.

tasomaniac: I believe this is just a copy paste. `todomvrx` project actually does not even use Koin. I'm sure there are more undeeded dependencies here.

tasomaniac: Removed bunch more in my previous PR https://github.com/airbnb/MvRx/pull/255/files#diff-9b2ebc432ba289ec4a9b8f3feb124159L48

PR #273, opened 2019-08-26
Delay warming reflection for 500ms

tasomaniac:
I did some investigation as per #264 regarding the effect of warming reflection. It takes around 1sec on OnePlus 5 (2 years old). It is considerably high, I would say. That is especially because it is done during construction of the ViewModel.

### Suggested solution: 

Delayed warming for 500 milliseconds.

### DIFF
--- a/mvrx/src/main/kotlin/com/airbnb/mvrx/BaseMvRxViewModel.kt
+++ b/mvrx/src/main/kotlin/com/airbnb/mvrx/BaseMvRxViewModel.kt
@@ -16,6 +16,7 @@ import io.reactivex.functions.Consumer
 import io.reactivex.schedulers.Schedulers
 import java.util.Collections
 import java.util.concurrent.ConcurrentHashMap
+import java.util.concurrent.TimeUnit
 import kotlin.reflect.KProperty1
 import kotlin.reflect.KVisibility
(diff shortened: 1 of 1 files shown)


BenSchwab:
@tasomaniac @gpeal I don't think I am following what the consequence/impetus of delaying the reflection warming is.

tasomaniac:
Reflection warming is done so that we don't have to pay that cost in case a configuration change happens. This is done during the constructor of every ViewModel. Since it is useful only when PersistState is being used and configuration change actually happens, I think it is a big cost to pay. Delaying it would help since it will not be done immediately.

gpeal:
@tasomaniac What do you mean by the cost if a configuration change happens?

BenSchwab:
But it's always done on a background thread regardless of the delay. Did you run a profile of this to determine it improves performance? I guess having one of your phone's cores occupied could slow things down, but I would find it surprising.

tasomaniac:
My intention with this PR was also to ask this. I'm not fully sure and it is not easy to measure. I wanted to open it since delaying also does not hurt. 

Recent devices are powerful. Consider a low-end device and multiple operations going on in the device already (like listening music and play store updates on background etc). I was thinking this delay would be useful on such low-end devices. 

I can actually grab a low-end device and try.

PR #282, opened 2019-09-04
Testing and mocking infrastructure

### LINKED ISSUE #265: Recommendation for rendering test
I'd like to know if there is a recommended way to test state rendering.
What I want to achieve is an Espresso test which checks state of all view widgets (visibility, text etc) for a given value of the `ViewState` class.

Here is an example of my current solution:

```
@Test
fun emptyLogin() {
    val scenario = launchFragmentInContainer<LoginFragment>()
    scenario.onFragment { fragment ->
        fragment.model.forceState(LoginViewState(isLoginEmpty = true))
    }
    onView(withText("Login can not be empty!"))
            .check(matches(isDisplayed()))
}
```
```
class LoginFragment : BaseMvRxFragment() {
    // ...
    val model: LoginViewModel by fragmentViewModel()
}
```
```
class LoginViewModel(
    initialState: LoginViewState
) : BaseMvRxViewModel<LoginViewState>(initialState, true) {
    // ...
    @VisibleForTesting(otherwise = VisibleForTesting.NONE)
    fun forceState(state: LoginViewState) = setState { state }
}
```

I find this `forceState` method dirty. Also I don't like the idea of making `model` field in the `Fragment` public. Is there any cleaner/recommended solution to achieve my goal?

elihart:
I am migrating our internal testing/mocking tools to this open source repo.

***Mock Behavior***
At a high level the idea is that a view model and state store can have a `MockBehavior` applied to it when it is created. This allows things such as blocking state from changing, allowing synchronous state changes to be forced, blocking async executions, etc.

I have found that it is necessary for testing infrastructure to be able to have hooks into when viewmodels and state stores are created and disposed, as well as allow mock behavior of existing state stores to be modified dynamically and globally. For example, the test system may need to globally toggle whether state changes are allowed.  The approach I have taken to do that is to have a single `MvRxViewModelConfigProvider` which has a customizable mock behavior, creates state stores based on that behavior, and keeps references to active stores and viewmodels to allow toggling their behavior.

I have created singletons to manage this configuration which are all under the `MvRx` object.

***Mock State***
I've included a system for easily generating mock state via a script. Since these are run from the command line they aren't a part of the library that is shipped, but I have packaged it as an executable so users can easily download and use it.

***MvRx launcher***
This allows loading and screen/state in the app. To ease review, I've set this up as a separate PR onto this branch - https://github.com/airbnb/MvRx/pull/292

***Unit Test Framework***
A separate piece is a unit test framework for view models that builds on this mocking behavior. This work is still to come.

Fixes #265


BenSchwab:
I think this pattern seems reasonable. RxJava has the RxJavaPlugins class that this feels similar to. Maybe we can unify this pattern with the existing debug pattern into a MvRxPlugin's class? It would be nice to define an interface for what is configurable, and then ship as much of the testing/mocking code as a separate artifact.

gpeal:
@BenSchwab We already have a testing artifact we can use for this 😄

mvrx/src/main/kotlin/com/airbnb/mvrx/BaseMvRxViewModel.kt
```diff
@@ -223,6 +267,29 @@ abstract class BaseMvRxViewModel<S : MvRxState>(
...
     }
 
+    /**
+     * Similar to [execute]. This subscribes to an Observable, provides a reducing function to update state when the observable passes a value,
+     * and registers the subscription to be disposed when the ViewModel is cleared.
+     *
+     * This function should be preferred over calling [Observable.subscribe] directly so that disposal is handled for you, and so setting state
+     * is simplified.
+     */
+    fun <T> Observable<T>.executeWithoutAsync(stateReducer: S.(value: T) -> S): Disposable {
```
tasomaniac: What about just calling this `subscribe`? Or something with that terminology. MvRx adds execute terminology and it has been working great. Not sure if that should be used here.

mvrx/src/main/kotlin/com/airbnb/mvrx/mock/MvRxViewModelConfig.kt
```diff
@@ -0,0 +1,223 @@
...
+import android.util.Log
+import com.airbnb.mvrx.BaseMvRxViewModel
+import com.airbnb.mvrx.MvRxState
+import com.airbnb.mvrx.MvRxStateStore
+import com.airbnb.mvrx.MvRxViewModelFactory
+import com.airbnb.mvrx.RealMvRxStateStore
+import java.util.LinkedList
+
+// TODO: Is there a better way to set this?
+var mvrxViewModelConfigProvider = MvRxViewModelConfigProvider()
```
tasomaniac: Apart from changing the debugMode boolean and changing properties, do you expect users to subclass the Provider? I realized that it is open.

Not sure how different it is but I would prefer keeping this Val and provide a function like below to configure further. 

```
mvrxConfigure(configure: MvRxViewModelConfigProvider.() -> Unit) {
   mvrxViewModelConfigProvider.configure()
}
```

mvrx/src/main/kotlin/com/airbnb/mvrx/mock/MockableMvRxView.kt
```diff
@@ -0,0 +1,819 @@
...
+
+
+interface MockableMvRxView : MvRxView {
+    fun provideMocks(): FragmentMocker<out MockableMvRxView, out Parcelable> = EmptyMocks
+}
+
+/**
+ * Used with [MockableMvRxView.provideMocks] for mocking a MvRx fragment that has no view models (eg only static content and fragment arguments).
+ *
+ * See https://airbnb.quip.com/4S6aAV3uzBRP/Testing-MvRx-Screens
```
digitalbuddha: Remove comment or change to public?

mvrx/src/main/kotlin/com/airbnb/mvrx/BaseMvRxViewModel.kt
elihart: The reason I prefixed this with execute was to include it in autocomplete. If people are used to using execute with mvrx, then `executeWithoutAsync` is discoverable

mvrx/src/main/kotlin/com/airbnb/mvrx/mock/MvRxViewModelConfig.kt
elihart: I don't think many people will need to subclass MvRxViewModelConfigProvider, but I left it open so that they could provide a custom implementation of `buildStateStore` if desired.

Any particular reasons you like the function configuration better? It seems like it would make it harder to override functions like `buildStateStore`, although we could have setter functions to override default behavior.

Needing to call a function like `mvrxConfigure` is a little less discoverable than just setting the val though

tasomaniac: I would do the function approach in case you have setters. It would look like a DSL and allow others to configure. I put that idea since some properties are variables already. 

I think if you want to make it open and provide people to replace the whole implementation, then shouldn't it be immutable?

mvrx/src/main/kotlin/com/airbnb/mvrx/mock/MockableMvRxStateStore.kt
```diff
@@ -0,0 +1,113 @@
+package com.airbnb.mvrx.mock
+
+import com.airbnb.mvrx.RealMvRxStateStore
+import com.airbnb.mvrx.ScriptableMvRxStateStore
+import com.airbnb.mvrx.ScriptableStateStore
+import com.airbnb.mvrx.mock.MockBehavior.StateStoreBehavior
+import io.reactivex.Observable
+import io.reactivex.schedulers.Schedulers
+
+interface MockableStateStore<S : Any> : ScriptableStateStore<S> {
```
BenSchwab: Is this scriptable state store needed any more? It seems equivalent to using the block action api and the freeze state call. There seems to be a lot of state stores, and confusing to track all their different purposes.

mvrx/src/main/kotlin/com/airbnb/mvrx/BaseMvRxViewModel.kt
```diff
@@ -223,6 +263,29 @@ abstract class BaseMvRxViewModel<S : MvRxState>(
...
+        val blockExecutions =
+            config.currentMockBehavior?.blockExecutions ?: MockBehavior.BlockExecutions.No
+
+        if (blockExecutions != MockBehavior.BlockExecutions.No) {
```
BenSchwab: Is there anyway we can handle this in the state store? It seems like the concept of blocking etc. is a state store concept, not a view model concept.

mvrx/src/main/kotlin/com/airbnb/mvrx/BaseMvRxViewModel.kt
```diff
@@ -24,28 +29,35 @@ import kotlin.reflect.full.primaryConstructor
...
 
 /**
- * To use MvRx, create your own base MvRxViewModel that extends this one and sets debugMode.
- *
- * All subsequent ViewModels in your app should use that one.
+ * To use MvRx, create view models that extend this base view model.
  */
 abstract class BaseMvRxViewModel<S : MvRxState>(
-    initialState: S,
-    debugMode: Boolean,
```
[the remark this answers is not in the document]
elihart: Yes, will definitely call out the breaking change in the release notes. I don't think debug mode changes for individual viewmodels are needed, but lmk if  any one disagrees

mvrx/src/main/kotlin/com/airbnb/mvrx/MvRx.kt
```diff
@@ -1,9 +1,49 @@
...
 object MvRx {
     /**
      * If your initial state needs to use Fragment arguments, store your arguments
      * as a parcelable class stored at this key.
      */
     const val KEY_ARG = "mvrx:arg"
+
+    /**
+     * Defines configuration for how ViewModels are created and what settings they use.
+     * By default this applies debug settings to all ViewModels and SHOULD be overridden in
```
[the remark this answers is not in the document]
gpeal: @elihart @BenSchwab I like the idea of having it be uninitialized and crash with a good error message. 
What we _really_ don't want is for people to not read the docs (very common) then:
1) Accidentally have debug mode always on
2) Never have debug mode on
3) Discover too late that it was never on, turn it on but have a bunch of random crashes that no one person can fix so they leave it off forever.

launcher/src/main/java/com/airbnb/mvrx/launcher/MvRxLauncherMockActivity.kt
```diff
@@ -0,0 +1,197 @@
...
+            try {
+                activity.showView(view)
+
+                if (activity.intent.getBooleanExtra(FINISH_AFTER_LAUNCH, false)) {
+                    // If we are initializing from arguments then network requests will be
+                    // made, and we want to give enough time for them to complete to make
+                    // sure the parsing and display it tested as well.
+                    val isInitializing =
+                        mock.mock.isDefaultInitialization || mock.mock.isForProcessRecreation
+                    val finishAfterMs: Long = if (isInitializing) 3000 else 500
```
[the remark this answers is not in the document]
BenSchwab: I also wonder if this can leverage some of the TTI concepts we use -- i.e. waiting until async props marked as necessary for load go to a terminal state.

launcher/src/main/java/com/airbnb/mvrx/launcher/ViewModelEnabler.kt
```diff
@@ -0,0 +1,69 @@
...
+        }, delayTime)
+    }
+
+    companion object {
+        /**
+         * Amount of time delay between when the fragment is resumed and when view models
+         * have "execution" enabled so that real loads can occur.
+         *
+         * This is left mutable so users can customize the default. Too short and it may not result
+         * in an accurate mock (as requests can override the mock state). Too long and it may
```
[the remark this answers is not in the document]
elihart: `any requests made during initialization are not fired` - this is it. Basically what I  am concerned  about is  some  delayed behavior in initialization that triggers a state change or execution. Maybe I  am being overly cautious here, since this generally shouldn't happen, but it seems  safer.

launcher/src/main/java/com/airbnb/mvrx/launcher/MvRxLauncherMockActivity.kt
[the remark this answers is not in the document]
elihart: those are nice  ideas for follow up work  :+1:

elihart:
@BenSchwab this is ready for review again. I have updated the mvrx testing module test rule to override mock behavior, and also cleaned up some other settings to decouple things more.

I spent a while trying to make the config factory injectable into the  view model, but in the end I really needed  it to be static because  the mocking system needs to be able to access it _before_ the view model is created which made things messy. This makes the overall interface  not as clean as we would perhaps like, but I  think it is reasonable

elihart:
I've released this branch as a 2.0.0-alpha1 release - https://github.com/airbnb/MvRx/releases/tag/2.0.0-alpha1

I've added documentation for all the new features to the wiki, as well as details on what breaking changes this introduced.

I am keeping this as an alpha to gather feedback on the new API's and breaking changes before we solidify them as a stable release.

I plan to keep all this work in this branch, with updates added as PR's on top of this branch, until we are ready for stable (or possibly beta). This will allow us to easily continue to ship small updates if needed

elihart:
Because of https://github.com/airbnb/MvRx/pull/334 I need to pull the mocking code into a separate module since it relies on kotlin reflect. Will make a new release (including fixes to other reported issues) once that is done

gpeal:
@elihart This can be closed right?

PR #283, opened 2019-09-04
Make searching for the companion object factory more robust

BenSchwab:
Ran into an issue with an engineer today where if a MvRxFactory companion object is named, we would fail to find it. This PR changes the search for a companion object to be more robust and search all declared (nested) classes.

### DIFF
--- a/mvrx/src/main/kotlin/com/airbnb/mvrx/MvRxViewModelProvider.kt
+++ b/mvrx/src/main/kotlin/com/airbnb/mvrx/MvRxViewModelProvider.kt
@@ -87,19 +87,14 @@ object MvRxViewModelProvider {
 }
 
 /**
- * Return the [Class] of the [MvRxViewModelFactory] for a given ViewModel class, if it exists.
+ * Return the [Class] of the companion [MvRxViewModelFactory] for a given ViewModel class, if it exists.
  */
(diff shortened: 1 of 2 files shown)


mvrx/src/main/kotlin/com/airbnb/mvrx/MvRxViewModelProvider.kt
```diff
@@ -87,19 +87,14 @@ object MvRxViewModelProvider {
...
  */
 internal fun <VM : BaseMvRxViewModel<*>> Class<VM>.factoryCompanion(): Class<out MvRxViewModelFactory<VM, *>>? {
-    val companionClass = try {
-        Class.forName("$name\$Companion")
-    } catch (exception: ClassNotFoundException) {
-        return null
-    }
-    return if (MvRxViewModelFactory::class.java.isAssignableFrom(companionClass)) {
+    return declaredClasses.firstOrNull {
+        MvRxViewModelFactory::class.java.isAssignableFrom(it)
```
BenSchwab: afaik something like `filterInstanceOf` won't work because we need to match on the inner generic type of `class`

BenSchwab:
I just profiled with 10+ nested classes, and the performance was sub millisecond.

PR #285, opened 2019-09-06
Upgrade androidx and allow layout resources to be specified in the fragment constructor

gpeal:
https://developer.android.com/jetpack/androidx/versions/stable-channel#september_5_2019

The main change allows specifying a layout resource in the constructor of your Fragment.


mvrx/src/main/kotlin/com/airbnb/mvrx/BaseMvRxFragment.kt
```diff
@@ -9,7 +10,7 @@ import androidx.lifecycle.LifecycleOwner
  *
  * This is necessary for the view model delegates and persistence to work correctly.
  */
-abstract class BaseMvRxFragment : Fragment(), MvRxView {
+abstract class BaseMvRxFragment(@LayoutRes contentLayoutId: Int = 0) : Fragment(contentLayoutId), MvRxView {
```
jonasbark: Judging from the default argument (0) I guess it wouldn't clash with other usages, e.g. headless fragments or databinding usage, wouldn't it?

tasomaniac: It does not matter with abstract classes, subclassing users should either provide the default constructor if they don't provide the contentLayout

gpeal: @jonasbark all it's doing it setting an int which will default to 0 anyway
https://android.googlesource.com/platform/frameworks/support/+/androidx-master-dev/fragment/fragment/src/main/java/androidx/fragment/app/Fragment.java#490

PR #287, opened 2019-09-14
Adding inter view model subscription support

pmecho:
When a view model contains a reference to another view model it'd be nice to allow it to subscribe to property changes. A use case is when a Fragment's view model contains a reference to the Activity's view model. Without this subscription support, the Fragment would have to contain the communication logic between the two view models.

### DIFF
--- a/mvrx/src/main/kotlin/com/airbnb/mvrx/BaseMvRxViewModel.kt
+++ b/mvrx/src/main/kotlin/com/airbnb/mvrx/BaseMvRxViewModel.kt
@@ -50,17 +50,17 @@ abstract class BaseMvRxViewModel<S : MvRxState>(
      * [Lifecycle.State.RESUMED] and when [ViewModel.onCleared] is called the lifecycle will be
      * [Lifecycle.State.DESTROYED].
      *
-     * This is not be publicly accessible as it should only be used to control subscriptions
+     * This is not publicly accessible as it should only be used to control subscriptions
      * between two view models.
      */
     private val lifecycleOwner: LifecycleOwner = LifecycleOwner { lifecycleRegistry }
-    private val lifecycleRegistry: LifecycleRegistry = LifecycleRegistry(lifecycleOwner)
+    private val lifecycleRegistry: LifecycleRegistry = LifecycleRegistry(lifecycleOwner).apply { currentState = Lifecycle.State.RESUMED }
+
 
     internal val state: S
         get() = stateStore.state
 
     init {
-        lifecycleRegistry.currentState = Lifecycle.State.RESUMED
         Completable.fromCallable { warmReflectionCache(initialState) }.subscribeOn(Schedulers.computation()).subscribe()
 
         if (this.debugMode) {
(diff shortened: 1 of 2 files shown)


mvrx/src/main/kotlin/com/airbnb/mvrx/BaseMvRxViewModel.kt
```diff
@@ -41,10 +43,24 @@ abstract class BaseMvRxViewModel<S : MvRxState>(
...
     private val activeSubscriptions = Collections.newSetFromMap(ConcurrentHashMap<String, Boolean>())
 
+    /**
+     * Define a [LifecycleOwner] to control subscriptions between [BaseMvRxViewModel]s. This only
+     * provides two states, [Lifecycle.State.RESUMED] and [Lifecycle.State.DESTROYED] as it follows
+     * the [ViewModel] object lifecycle. That is, when instantiated the lifecycle will be
+     * [Lifecycle.State.RESUMED] and when [ViewModel.onCleared] is called the lifecycle will be
+     * [Lifecycle.State.DESTROYED].
+     *
+     * This is not be publicly accessible as it should only be used to control subscriptions
```
elihart: nit: missed a word `is not be publicly`

mvrx/src/main/kotlin/com/airbnb/mvrx/BaseMvRxViewModel.kt
```diff
@@ -41,10 +43,24 @@ abstract class BaseMvRxViewModel<S : MvRxState>(
...
+     * between two view models.
+     */
+    private val lifecycleOwner: LifecycleOwner = LifecycleOwner { lifecycleRegistry }
+    private val lifecycleRegistry: LifecycleRegistry = LifecycleRegistry(lifecycleOwner)
+
     internal val state: S
         get() = stateStore.state
 
     init {
+        lifecycleRegistry.currentState = Lifecycle.State.RESUMED
```
elihart: could do this inline with the declaration to group it all together
```
LifecycleRegistry(lifecycleOwner).apply { currentState = Lifecycle.State.RESUMED }

File: mvrx/src/main/kotlin/com/airbnb/mvrx/BaseMvRxViewModel.kt [hunk 3]
```diff
@@ -602,6 +656,12 @@ abstract class BaseMvRxViewModel<S : MvRxState>(
         return this
     }
 
+    private fun <S : MvRxState> assertSubscribeToDifferentViewModel(viewModel: BaseMvRxViewModel<S>) {
+        if (this == viewModel) {
```
elihart: could also do
```
require(this != viewModel) {
 "This method is for subscribing to other view models. Please pass a different instance as the argument."
}
```

Still throws `IllegalArgumentException` and is a little shorter and more idiomatic kotlin.

BenSchwab:
I'm not fully understanding why a view model needs to be a lifecycle owner. Would exposing the protected functions, and calling with `null` be sufficient?

If a fragment view model subscribes to an activity view model, when the fragment view model is destroyed, it will call `disposeOnClear` and everything is good.

If an activity view model subscribes to a fragment view model, then this would help as the activity view model would see that the fragment view model has been destroyed and remove it's subscription. _but_ I feel like this case is an error to begin with, so not sure we should build infrastructure that supports it. Thoughts @pmecho @elihart?

pmecho:
> If a fragment view model subscribes to an activity view model, when the fragment view model is destroyed, it will call `disposeOnClear` and everything is good.

@BenSchwab I don't think this case works as is. When the fragment view model subscribes to the activity view model it doesn't keep a reference to the disposable. That'd be created on the activity's view model since we call `activityViewModel.subscribe(...)`. When the fragment is cleared it doesn't know that it subscribed to the activity's view model. I might have missed something though.

PR #295, opened 2019-09-30
feat: Add news app sample to showcase usage of Dagger with MvRx

### LINKED ISSUE #278: Add sample for MvRx usage with Dagger
I've recently spent some time integrating Dagger with an app using MvRx, and I was thinking it would could be useful for people if this repository contained an official sample for this too. A sample application using MvRx with Dagger and AssistedInject.

I could refactor an existing sample app in this repo to use Dagger, or maybe create an entirely new one. Thoughts?

haroldadmin:
Here is a very small app using the [NewsAPI](https://newsapi.org). It showcases the usage of Dagger and AssistedInject with apps using MvRx. I have tried to keep the app extremely simple: it just displays a feed of news articles and launches a Custom Tab intent when an article is clicked.

This fixes #278.


elihart:
Thanks for sharing this sample for everyone. I'm curious about your experience with `AssistedInject` though - it seems to be a fair amount of boilerplate and makes for a tight coupling between your fragment and viewmodel.

The goal is to be able to inject a dagger provided instance into the viewmodel constructor, right? We use an Application wide dagger component system that allows us to easily access a dependency off of a graph when we need it for creating  the viewmodel.

```kotlin
    companion object : MvRxViewModelFactory<HeadlinesViewModel, HeadlinesState> {

        override fun create(viewModelContext: ViewModelContext, state: HeadlinesState): HeadlinesViewModel? {
           val headlinesUseCase by inject(BaseGraph::headlinesUseCase)
            return HeadlinesViewModel(state, headlinesUseCase)
        }
    }
```

In my opinion this is a much simpler pattern - you don't have to add all of the boilerplate of assisted inject, and the fragment doesn't need to be involved.

haroldadmin:
Thanks for taking a look at this Eli.

The goal with AssistedInject is to be able to do constructor injection even for parameters which are available *only at runtime*, such as initial state and fragment args. The other dependencies are satisfied using Dagger.

For this News app, suppose we had a screen which showed the articles related to a particular topic, and we pass this `topic` parameter as a fragment arg from another screen. This `topic` parameter would be considered a parameter available only at runtime since we can not know it at compile time. However, it would also be great if we could have it available to our ViewModel in its constructor. Something like this:

```kotlin
class TopicsViewModel @Inject constructor(
  initialState: TopicState, val topic: String, ... // a gazillion other dependencies
): MvRxViewModel(initialState) {
  init {
    loadArticlesForTopic(topic)
  }
  ...
}
```

This would not be ordinarily possible with Dagger, since it does not know how to satisfy this dependency on the `topic` and `initialState` parameters. This is where AssistedInject comes in, as it would create an implementation of the `Factory` class for our ViewModel automatically. This factory would take in the runtime params, and give us back an instance of our ViewModel. Something like this:

```kotlin
class TopicsViewModel @AssistedInject constructor(
  @Assisted initialState: TopicState,
  @Assisted val topic: String, 
  ... // a gazillion other dependencies
): MvRxViewModel(initialState) {
  init {
    loadArticlesForTopic(topic)
  }
  ...

  @AssistedInject.Factory
  interface Factory {
    fun create(topic: String, initialState: TopicsState): TopicsViewModel
  }
}
```

Using this approach, our `companion object` for this ViewModel and the fragment can be implemented as:

```kotlin
class TopicsViewModel ... {
  companion object: MvRxViewModelFactory<TopicsViewModel, TopicsState> {
    override fun create(...) {
      val fragment = (context as FragmentViewModelContext).fragment<TopicsFragment>()
      val topic = fragment.arguments.getString("KEY_TOPIC")
      return fragment.viewModelFactory.create(topic, state)
}

class TopicsFragment: MvRxFragment() {
  @Inject lateinit var viewModelFactory: TopicsViewModel.Factory
  ...
}
```

The advantages of this method are:
* We don't have to manually provide the other gazillion dependencies from our Dagger component. 
* Our implementation of this companion object would need no change even if we add other dependencies in our ViewModel's constructor, if they can be satisfied with dagger.

Of course, this can be done without AssistedInject as well. However, for people that prefer this style, this sample may be of help.

Finally, maybe I should actually add this topics screen to the sample to make it better. Would you agree?

gpeal:
If we are going to make a standalone dagger example (which I think we should), let's make it as minimal as possible. It's neat that this is a fully fledged app that does something but I think it actually distracts from the purpose of demonstrating dagger. Let's make this as simple as possible.

@haroldadmin @elihart @BenSchwab

haroldadmin:
@gpeal What do you have in mind? Demonstrating Dagger and AssistedInject would require at least some functionality in an app. Should I make the news app simpler (maybe remove the pagination?) or create something new?

gpeal:
@haroldadmin It can be a simple repository that has a single observable function that returns something like `Observable.just("Hello World").delay(2000, TimeUnit.Seconds)` or something like that.

haroldadmin:
Okay, I guess it will be cleaner to discard this pull request and create a new one for the simpler app then. I shall close this one when the new one is ready.

BenSchwab:
Thanks for creating this @haroldadmin! I agree that we should do something as simple as possible.

But most of all, it's great that this is getting the ball ruling on amping up MvRx Dagger support.

I think the approach here, of injecting the factory on the fragment, would make the most sense if you were going to use something from the fragment, such as binding the fragment argument into a dagger subcomponent.

In this simple case, instead we could be add a provision method to our dagger component:
```
interface AppComponent { 
     fun headlinesViewModelFactory(): HeadlinesViewModel.Factory
}
```
And then access the factory in the MvRxViewModelFactory from the component. This only couples the MvRxViewModelFactory with the component, instead of the MvRxViewModelFactory with the component through the fragment.

knordbo:
I've used similar approach in a sample except constructor injected the view model factory in the fragment (https://github.com/knordbo/satis/blob/master/app/src/main/java/com/satis/app/feature/account/AccountFragment.kt#L21). If multiple fragments use the same view model those fragments could simply conform to some interface.

I do think though that if you don't constructor inject it, your implementation is still doing stuff like `(requireActivity().application as NewsApplication).appComponent` so you might as well use a provisioning method as suggested above.

I would suggest writing a dummy UI test as well to see if your setup makes sense for testing. One benefit of constructor injecting everything is no need for dagger in your tests, though you can of course still do it if you don't want to pass in all the args manually. Dummy test without dagger: https://github.com/knordbo/satis/blob/master/app/src/androidTest/java/com/satis/app/feature/account/AccountFragmentTest.kt https://github.com/knordbo/satis/blob/master/app/src/androidTest/java/com/satis/app/feature/account/AccountFakes.kt

haroldadmin:
@knordbo Constructor injection is Fragments is not specifically a feature of Dagger, but rather of `FragmentFactory`, right? It would be fun to try it out, but it would again tie the ViewModel to the fragment. I think adding the provisioning method in a Component is the way to go here.

knordbo:
@haroldadmin `IntoMap` is a feature of dagger (https://github.com/knordbo/satis/blob/master/app/src/main/java/com/satis/app/feature/account/AccountModule.kt#L20). You are right that `FragmentFactory` is used, but constructor injection is still done using Dagger. I don't see any difference between constructor injecting fragments, view models or any other classes from that point of view.

You don't have to tie the ViewModel to the fragment, just as how you want `appComponent` to be an interface providing this factory, you can make the fragment/activity implement the factory interface which your view model will call. Something along the lines of (imagine traverseFor is a simple extension function which depending on fragment/activity view model context finds the closest of type T):

```
override fun create(viewModelContext: ViewModelContext, state: HeadlinesState): HeadlinesViewModel? {
    val factory = viewModelContext.traverseFor<ImagesViewModel.Factory>()
    return factory.create(state)
}
```

Again, this is mostly a personal preference as I found it easier to test (espresso or any kind of instrumentation test) without putting up test components etc in dagger. Would be interested to see how your test setup looks like! :)

haroldadmin:
I have created another PR #307 using the suggestions everyone shared here. Closing this one as planned.

PR #300, opened 2019-10-03
Fix reproducible crash on minified applications using MvRx.

aderington:
Crash is triggered in `PersistState.persistState` when it looks for the `primaryConstructor` of the `MvRxState`.  The metadata that Kotlin reflection uses to indicate the `primaryConstructor` is stripped by Proguard without this rule.

Fixes Issue: https://github.com/airbnb/MvRx/issues/299

### DIFF
--- a/mvrx/proguard-rules.pro
+++ b/mvrx/proguard-rules.pro
@@ -43,3 +43,6 @@
 # kept as they are accessed via reflection.
 -keepnames class com.airbnb.mvrx.MvRxState
 -keepnames class * implements com.airbnb.mvrx.MvRxState
+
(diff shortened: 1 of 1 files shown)


haroldadmin:
The 'primaryConstructor()` call does not play well with Proguard/R8. Another approach to solving this problem would be to replace the call to `primaryContructor()` with `constructors.first()`. This way we can avoid keeping the Kotlin metadata for reflection.

I was able to fix this problem in my own library with this approach.

rossbacher:
@haroldadmin There are many more issues that require us to keep the `@Metadata` annotation. Just not in the lib.

rossbacher:
Replaced by #304 .

PR #302, opened 2019-10-03
Fix infinite thread leaks while flushing queues in state store

xMickeymikex:


elihart:
If https://github.com/airbnb/MvRx/pull/277 is working then you should be able to get this in a 1.2.0-SNAPSHOT soon

xMickeymikex:
No problem ;)

PR #307, opened 2019-10-12
feat: Add a sample app to demonstrate usage of Dagger and AssistedInject with MvRx

### LINKED ISSUE #278: Add sample for MvRx usage with Dagger

haroldadmin:
Here is a very small app which showcases the usage of Dagger and AssistedInject with apps using MvRx. The app simply shows a "Hello, world!" message after two seconds.

This fixes #278. This PR was made keeping the suggestions in #295 in mind.


haroldadmin:
@BenSchwab @elihart @gpeal Do you think adding constructor injection in Fragments (as suggested by @knordbo) would be a good idea for this sample? I have left it out in the interest of simplicity, but would be happy to add it if you think so.

hellodagger/src/main/java/com/airbnb/mvrx/helloDagger/HelloFragment.kt
```diff
@@ -0,0 +1,41 @@
...
+                    messageTextView.text = state.message()
+                    helloButton.isEnabled = true
+                }
+                is Fail -> {
+                    messageTextView.text = getString(R.string.helloFragmentFailureText)
+                    helloButton.isEnabled = true
+                }
+            }
+        }
+    }
```
gpeal: This can be simplified as
```
override fun invalidate() = withState(viewModel) { state ->
    helloButton.isEnabled = state.message !is Loading
    messageView.text = when (state.message) {
        is Uninitialized, is Loading ->  getString(R.string.helloFragmentLoadingText)
        is Success -> state.message()
        is Fail ->  getString(R.string.helloFragmentFailureText)
    }
}
```

hellodagger/src/main/res/values/strings.xml
```diff
@@ -0,0 +1,6 @@
+<resources>
+    <string name="app_name">Hello Dagger</string>
+    <string name="helloButtonText">Say Hello</string>
```
gpeal: String resources should be undescored

hellodagger/src/main/java/com/airbnb/mvrx/helloDagger/HelloFragment.kt
haroldadmin: Fixed, thanks for the suggestion.

hellodagger/src/main/res/values/strings.xml
haroldadmin: Fixed.

haroldadmin:
@gpeal I have implemented the method you suggested for ViewModel injection. I think it is great. There is some initial Boilerplate that needs to be written (ViewModelKey, Interface for Assisted factories, Map injection), but it would scale better. Thanks for the idea!

hellodagger/src/main/java/com/airbnb/mvrx/helloDagger/HelloFragment.kt
```diff
@@ -0,0 +1,31 @@
...
+import android.view.ViewGroup
+import com.airbnb.mvrx.*
+import kotlinx.android.synthetic.main.fragment_hello.*
+
+class HelloFragment: BaseMvRxFragment() {
+
+    private val viewModel: HelloViewModel by fragmentViewModel()
+
+    override fun onCreateView(inflater: LayoutInflater, container: ViewGroup?, savedInstanceState: Bundle?): View? {
+        return inflater.inflate(R.layout.fragment_hello, container, false)
```
gpeal: You can put this layout in the constructor of BaseMvRxFragment now :)

hellodagger/src/main/java/com/airbnb/mvrx/helloDagger/HelloFragment.kt
```diff
@@ -0,0 +1,31 @@
...
+    }
+
+    override fun onViewCreated(view: View, savedInstanceState: Bundle?) {
+        super.onViewCreated(view, savedInstanceState)
```
gpeal: You don't need super for onViewCreated

hellodagger/src/main/java/com/airbnb/mvrx/helloDagger/di/AssistedViewModelFactory.kt
```diff
@@ -0,0 +1,13 @@
+package com.airbnb.mvrx.helloDagger.di
+
+import com.airbnb.mvrx.MvRxState
+import com.airbnb.mvrx.helloDagger.base.MvRxViewModel
+
+/**
+ * Creates and returns instances of the requested ViewModel.
+ *
+ * This interface serves as a supertype for AssistedInject factories in ViewModels.
```
gpeal: Can you add a code snippet? It's not clear what this does with somebody who has no context.

hellodagger/src/main/java/com/airbnb/mvrx/helloDagger/HelloFragment.kt
haroldadmin: Thanks, good idea!

hellodagger/src/main/java/com/airbnb/mvrx/helloDagger/HelloFragment.kt
haroldadmin: Right, it's an empty method. Thanks, I've fixed this.

hellodagger/src/main/java/com/airbnb/mvrx/helloDagger/di/AssistedViewModelFactory.kt
haroldadmin: How does this look?
```
/**
 * Creates and returns instances of the requested ViewModel.
 *
 * This interface serves as a supertype for AssistedInject factories in ViewModels.
 * We can use this as a marker interface in a Multibinding Dagger setup to populate a map
 * of ViewModel classes with their AssistedInject factories.
 *
 * Here's an example for such a setup:
 *
 * class MyViewModel @AssistedInject constructor(...): BaseMvRxViewModel<MyState>(...) {
 *   @AssistedInject.Factory
 *   interface Factory : AssistedViewModelFactory<MyViewModel, MyState> {
 *     override fun create(state: MyState): MyViewModel
 *   }
 * }
 *
 * @AssistedModule(includes = [AssistedInject_MyAppModule::class])
 * @Module
 * interface MyAppModule {
 *   @Binds
 *   @IntoMap
 *   @ViewModelKey(MyViewModel::class)
 *   fun myViewModelFactory(factory: MyViewModel.Factory): AssistedViewModelFactory<*, *>
 * }
 *
 * Now we can request Dagger for an instance of Map<Class<out MvRxViewModel<*>>, AssistedViewModelFactory<*, *>>,
 * and use it to access our ViewModel Factories.
 */
interface AssistedViewModelFactory<VM: MvRxViewModel<S>, S: MvRxState> {
    fun create(state: S): VM
}
```

hellodagger/src/main/java/com/airbnb/mvrx/helloDagger/di/AppComponent.kt
```diff
@@ -0,0 +1,14 @@
...
+import com.airbnb.mvrx.helloDagger.HelloViewModel
+import com.airbnb.mvrx.helloDagger.base.MvRxActivity
+import dagger.Component
+
+@Component(modules = [AppModule::class])
+interface AppComponent {
+
+    fun inject(activity: MvRxActivity)
+
+    fun helloViewModelFactory(): HelloViewModel.Factory
```
knordbo: I believe this is unused

hellodagger/src/main/res/layout/activity_main.xml
```diff
@@ -0,0 +1,7 @@
+<?xml version="1.0" encoding="utf-8"?>
+<FrameLayout xmlns:android="http://schemas.android.com/apk/res/android"
```
knordbo: nitpick: you could use [FragmentContainerView](https://developer.android.com/reference/androidx/fragment/app/FragmentContainerView)

hellodagger/src/androidTest/java/com/airbnb/mvrx/helloDagger/HelloFragmentTest.kt
```diff
@@ -0,0 +1,19 @@
...
+import androidx.test.espresso.assertion.ViewAssertions.matches
+import androidx.test.espresso.matcher.ViewMatchers.*
+import org.hamcrest.CoreMatchers.not
+import org.junit.Test
+
+class HelloFragmentTest {
+
+    @Test
+    fun showsLoadingMessageWhenCreated() {
+        val scenario = launchFragmentInContainer<HelloFragment>()
```
knordbo: I'm surprised this works :O `launchFragmentInContainer` should launch an empty activity hosting your fragment, this activity which is not a `MvRxActivity` which is a hard requirement from the view model factory setup if I understand the code correctly.

haroldadmin: This test does indeed pass, and upon further investigation I think this is because of the lazy nature of 'by fragmentViewModels()` delegate. I think `invalidate()` is never called, and that's why the ViewModel is never accessed within the fragment. Since it is lazily initialized, it also means that it is never created, and thus the underlying activity is never used for its creation using the multibinding map.

Thanks for pointing this out. I'm replacing the usage of `FragmentScenario` with `ActivityScenario` and re writing the test using it.

hellodagger/src/main/res/layout/activity_main.xml
haroldadmin: Okay, changed it to use FragmentContainerView.

hellodagger/src/main/java/com/airbnb/mvrx/helloDagger/di/AppComponent.kt
haroldadmin: Yep, it is. Fixed now. Thanks!

hellodagger/src/androidTest/java/com/airbnb/mvrx/helloDagger/HelloFragmentTest.kt
haroldadmin: I'm running into a problem here. This is what my test currently looks like:

```kotlin
@Test
    fun fragmentIsInLoadingStateWhenCreated() {
        val scenario: ActivityScenario<MainActivity> = ActivityScenario.launch(MainActivity::class.java)

        scenario.onActivity { activity: MainActivity ->
            val fragment = activity.findFragmentById<HelloFragment>(R.id.rootContainer)

            withState(fragment.viewModel) {
                assert(it.message is Loading)
            }

            onView(withId(R.id.messageTextView)).check(matches(withText(R.string.hello_fragment_loading_text)))
            onView(withId(R.id.helloButton)).check(matches(not(isEnabled())))
        }
    }
```

The problem is that the test just never completes. I'm not sure what's the problem here. On adding some log statements, I can see that the call to the first `onView(...)` never completes.

Is there anything I'm doing wrong here?

gpeal: It's not exactly lazy. It's a "lifecycle away aware lazy" and auto instantiates itself in onStart. `postInvalidate()` is also called in onStart() so that MvRx works with static screens and for cases like this.

hellodagger/src/main/java/com/airbnb/mvrx/helloDagger/Extensions.kt
```diff
@@ -0,0 +1,11 @@
+package com.airbnb.mvrx.helloDagger
+
+import androidx.annotation.IdRes
+import androidx.fragment.app.Fragment
+import com.airbnb.mvrx.helloDagger.base.MvRxActivity
+
+// Useful for getting access to a fragment in tests
+fun <A: Fragment> MvRxActivity.findFragmentById(@IdRes id: Int): A {
```
gpeal: Can you move this to your test class then?

hellodagger/src/main/java/com/airbnb/mvrx/helloDagger/base/MvRxActivity.kt
```diff
@@ -0,0 +1,26 @@
...
+import com.airbnb.mvrx.helloDagger.di.AssistedViewModelFactory
+import javax.inject.Inject
+
+/**
+ * Base class for Activities.
+ *
+ * This class contains an injected map of [AssistedViewModelFactory]s to make it easier to create
+ * and instantiate requested ViewModels.
+ */
+abstract class MvRxActivity(@LayoutRes layoutResId: Int = 0) : AppCompatActivity(layoutResId) {
```
gpeal: I could call this BaseActivity as to not confuse it with a class provided by the core library

hellodagger/src/main/java/com/airbnb/mvrx/helloDagger/di/AssistedViewModelFactory.kt
gpeal: Better, for this example, I think more documentation and less code makes sense.

hellodagger/src/main/java/com/airbnb/mvrx/helloDagger/Extensions.kt
haroldadmin: For some reason Android Studio says that the `supportFragmentManager.findFragmentById()` method is [inaccessible](https://user-images.githubusercontent.com/24315306/66856438-d5125680-efa2-11e9-8d31-c58c05cfa100.png) when I place this file is not in the main source set. It compiles and the test runs just fine though, so I think this could be an IDE bug. I have moved it to the `androidTest` source set now.

hellodagger/src/main/java/com/airbnb/mvrx/helloDagger/base/MvRxActivity.kt
haroldadmin: Okay, going to rename it. I think I'll go ahead and do the same with the base ViewModel class as well then.

hellodagger/src/main/java/com/airbnb/mvrx/helloDagger/di/AssistedViewModelFactory.kt
haroldadmin: I've revised the documentation for this interface, and have added more explanations to all the parts. It is longer, but I think it conveys the message a lot better now:

```kotlin
/*
 * Serves as a supertype for AssistedInject factories in ViewModels.
 *
 * We can use this interface as a marker in a Multibinding Dagger setup to populate a Map
 * of ViewModel classes with their AssistedInject factories.
 *
 * This setup is needed because AssistedInject factories are isolated from each other, so this
 * interface provides a common parent type to facilitate grouping these factories into a collection.
 * When we use it correctly with Dagger, a Map of ViewModels and their factories can be created for us
 * automatically. This Map can then be used to retrieve for a ViewModel's factory using its class.
 *
 * Here's an example for such a setup:
 *
 * First we define our ViewModel with an @AssistedInject annotated constructor, and a Factory interface
 * implementing AssistedViewModelFactory.
 *
 * class MyViewModel @AssistedInject constructor(...): BaseMvRxViewModel<MyState>(...) {
 *   @AssistedInject.Factory
 *   interface Factory : AssistedViewModelFactory<MyViewModel, MyState> {
 *     override fun create(state: MyState): MyViewModel
 *   }
 * }
 *
 * Then we need to create a Dagger Module, which contains methods that @Binds @IntoMap all of our
 * AssistedViewModelFactories using a [ViewModelKey]. Notice that the input to these methods is
 * the exact type of our AssistedInject factory, but the return type is an AssistedViewModelFactory.
 *
 * @AssistedModule(includes = [AssistedInject_MyAppModule::class])
 * @Module
 * interface MyAppModule {
 *   @Binds
 *   @IntoMap
 *   @ViewModelKey(MyViewModel::class)
 *   fun myViewModelFactory(factory: MyViewModel.Factory): AssistedViewModelFactory<*, *>
 * }
 *
 * This Module tells Dagger to include MyViewModel.Factory class in the Multibinding map using
 * MyViewModel::class as the key. Such a method should be added for **every ViewModel Factory**
 * so that they can be identified by Dagger and used for populating the Map.
 *
 * The generated map can then be injected wherever it is required.
 *
 * class BaseActivity: BaseMvRxActivity() {
 *   @Inject
 *   lateinit var factories: @JvmSuppressWildcards Map<Class<out BaseViewModel<*>>, AssistedViewModelFactory<*, *>>
 *
 *   onCreate(...) {
 *     appComponent().inject(this)
 *   }
 * }
 */
interface AssistedViewModelFactory<VM: BaseViewModel<S>, S: MvRxState> {
    fun create(state: S): VM
}
```

hellodagger/src/main/java/com/airbnb/mvrx/helloDagger/base/BaseActivity.kt
```diff
@@ -0,0 +1,26 @@
...
+/**
+ * Base class for Activities.
+ *
+ * This class contains an injected map of [AssistedViewModelFactory]s to make it easier to create
+ * and instantiate requested ViewModels.
+ */
+abstract class BaseActivity(@LayoutRes layoutResId: Int = 0) : AppCompatActivity(layoutResId) {
+
+    @Inject
+    lateinit var viewModelFactoryMap: @JvmSuppressWildcards Map<Class<out BaseViewModel<*>>, AssistedViewModelFactory<*, *>>
```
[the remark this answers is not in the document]
haroldadmin: Yep. It would be simpler to simply add a provision for this dependency in the AppComponent.
I honestly don't have a better reason for why I injected in the activity other than "I've always done it this way"

Will change to dependency provision in the AppComponent and update the PR soon.

hellodagger/src/main/java/com/airbnb/mvrx/helloDagger/HelloViewModel.kt
```diff
@@ -0,0 +1,34 @@
...
+    fun sayHello() {
+        repo.sayHello().execute { copy(message = it) }
+    }
+
+    @AssistedInject.Factory
+    interface Factory : AssistedViewModelFactory<HelloViewModel, HelloState> {
+        override fun create(state: HelloState): HelloViewModel
+    }
+
+    companion object : MvRxViewModelFactory<HelloViewModel, HelloState> {
```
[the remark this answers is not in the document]
haroldadmin: Sure, that would simplify things.

hellodagger/src/main/java/com/airbnb/mvrx/helloDagger/base/BaseViewModel.kt
```diff
@@ -0,0 +1,32 @@
...
+ * [AssistedViewModelFactory] interface.**
+ */
+abstract class BaseViewModel<S : MvRxState>(initialState: S) : BaseMvRxViewModel<S>(initialState, BuildConfig.DEBUG) {
+
+    companion object {
+        inline fun <reified VM : BaseViewModel<S>, reified S : MvRxState> createViewModel(
+                fragmentActivity: FragmentActivity,
+                state: S
+        ): VM {
+            val activity = fragmentActivity as BaseActivity
```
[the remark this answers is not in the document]
haroldadmin: You're right, getting rid of the Activity from the DI picture would be great. I'll make this change.

hellodagger/src/main/java/com/airbnb/mvrx/helloDagger/base/BaseViewModel.kt
```diff
@@ -0,0 +1,32 @@
...
+            val viewModelFactory = activity.viewModelFactoryMap[VM::class.java]
```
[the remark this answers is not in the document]
haroldadmin: The smaller lifecycle only extends to state persistence across configuration changes, and I'm not sure if that would ever bring about a change in the state, unless your state is dependent upon configuration properties. So in most cases I think it is pretty safe to assume that accessing an Activity this way is not a problem.

That being said, I would still be happy if we can get rid of the Activity from this picture completely. A DaggerMvRxViewModelFactory and a factory map dependency provision in AppComponent are great suggestions. Thanks!

I'll update the PR with these changes.

hellodagger/src/main/java/com/airbnb/mvrx/helloDagger/HelloViewModel.kt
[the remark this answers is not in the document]
BenSchwab: Okay, either way. Will leave this up to you.

haroldadmin:
@BenSchwab @gpeal I've made the following changes:
- Remove BaseActivity
- Remove BaseViewModel
- Add a DaggerMvRxViewModelFactory
- Provision for AssistedInject factories map in AppComponent

Thanks for the great suggestions.

I'm not so sure about making DaggerMvRxViewModelFactory an abstract class though. I could not figure out a way to keep it an interface and still use an inline function with reified types in it. Any feedback would be great.

haroldadmin:
@gpeal Yes, definitely.

haroldadmin:
@gpeal I have added the README file. It contains a section for why `AssistedViewModelFactory` and `DaggerMvRxViewModelFactory` were added, a small usage example, as well as a short paragraph on how things work. Do you think I should add more details to it?

gpeal:
@haroldadmin nvm, found you Twitter

haroldadmin:
@gpeal Thanks! My twitter handle is the same as my Github username: @haroldadmin.

haroldadmin:
@gpeal @BenSchwab Do you think we should look into making this a multi-module app in the future? The only thing which changes is each module gets its own dagger component and dagger module. All the different components can be brought together using dependent components in Dagger. I have tried this in my own app which uses dynamic feature modules and it works well.

BenSchwab:
@haroldadmin I'm personally very curious to learn more about your dagger set up (specifically, how you are augmenting the multibound map in the subcomponents?).

I think in terms of a sample, multi-module seems a bit out of scope for a sample. There are lots of ways to set up dagger in a multi-module project, so canonicalizing one as the preferred way seems outside of MvRx's domain.

But also sounds like great blog post material!

kdreamix:
@haroldadmin definitely interested in the multi-module setup, would you mind to share it in a separate repo?

PR #327, opened 2019-12-02
Separated PersistState into its own module and removed kotlin-reflect

gpeal:
In this PR, I was able to replace all non-`@PersistState` usages of kotlin-reflect with java reflection. I then moved `@PersistState` into its own new library `mvrx-persiststate` and removed kotlin-reflect from the main mvrx module.

The main library uses java reflection to try and do the state persistence and just noops if it doesn't exist so all you need to do is add the library for it to work.

# TODO
- [ ] Figure out what to do with tests. Some of the core tests use `@PersistState`. I could move them but then it will make jacoco more complicated and make it more confusing to know where to write future tests. Alternatively, we could make a new `unittests` module and move all mvrx tests there. Then there will be one module with all unit tests and it'll be clear. If we do that, should we just add them to the `testing` module or make a new one?
- [ ] Rename `testing` to `mvrx-testing` to make it clear that it is a library, not a sample and make it consistent with `mvrx-persiststate`. If we do that, maybe we should move all sample modules into a subfolder. This also opens up the ability to make `mvrx-coroutines` or maybe finally convert the internals to coroutines and make `mvrx-rxjava`. What do you think?


mvrx-persiststate/src/main/kotlin/com/airbnb/mvrx/persiststate/StatePersistorImpl.kt
```diff
@@ -0,0 +1,114 @@
...
+import kotlin.reflect.KProperty1
+import kotlin.reflect.KVisibility
+import kotlin.reflect.full.declaredMemberProperties
+import kotlin.reflect.full.instanceParameter
+import kotlin.reflect.full.memberFunctions
+import kotlin.reflect.full.primaryConstructor
+
+object StatePersistorImpl : StatePersistor {
+    override fun <S : MvRxState> persistState(obj: S, assertCollectionPersistability: Boolean): Bundle {
+        val persistStateArgs = obj::class.primaryConstructor
```
gpeal: The implementation here was copied directly.

mvrx/src/main/kotlin/com/airbnb/mvrx/BaseMvRxViewModel.kt
```diff
@@ -60,7 +58,7 @@ abstract class BaseMvRxViewModel<S : MvRxState>(
         get() = stateStore.state
 
     init {
-        Completable.fromCallable { warmReflectionCache(initialState) }.subscribeOn(Schedulers.computation()).subscribe()
+        Completable.fromCallable { statePersistorOrNull()?.warmReflectionCache(initialState) }.subscribeOn(Schedulers.computation()).subscribe({}, {})
```
gpeal: Errors here are now silently caught.

mvrx/src/main/kotlin/com/airbnb/mvrx/MvRxFactory.kt
```diff
@@ -58,7 +56,7 @@ private fun <VM : BaseMvRxViewModel<S>, S : MvRxState> createViewModel(
     return requireNotNull(viewModel) {
         // If null, use Kotlin reflect for best error message. We will crash anyway, so performance
         // doesn't matter.
-        if (viewModelClass.kotlin.primaryConstructor?.parameters?.size?.let { it > 1 } == true) {
+        if (viewModelClass.constructors.firstOrNull()?.parameters?.size?.let { it > 1 } == true) {
```
gpeal: This seems to work fine and all tests pass.

elihart: do we have a test where a State class has a secondary constructor?

PR #331, opened 2019-12-13
Work with Kotlin reflection support in shrinker

rossbacher:
Modify rules to remove the ones from default that are not required if Kotlin reflection is supported by the code shrinker.
This way we can work with Dexguard 8.6+ and have some additional code removed.

I also added this to the Wiki here: https://github.com/airbnb/MvRx/wiki#kotlin-reflection-with-proguarddexguardr8. I'm open to not have the change in the README, but as that this is a change that almost everybody using the lib will have to do when updating I though I put up there too.


MatushkaRossiya:
Proguard 6.3 (currently released in beta) does support processing of Kotlin metadata according to release notes, so maybe it's worth noting in the README and wiki.

rossbacher:
This has been made obsolete by https://github.com/airbnb/MvRx/pull/334

PR #334, opened 2019-12-22
Remove Kotlin-Reflect entirely

gpeal:
I was able to fully remove kotlin-reflect without changing any functionality, annotation processors, or separating `@PersistState` to its own module. It also doesn't rely on Kotlin's `@Metadata` annotation at all.

These were all trivial except for `@PersistState`. I tried a few methods including using the [kotlinx-metadata-jvm](https://github.com/JetBrains/kotlin/tree/master/libraries/kotlinx-metadata/jvm) library which is significantly smaller than `kotlin-reflect`. I got it to work but was able to remove the need for even that.

The comments in `PersistState.kt` should be clear enough to understand what it is doing. If it is not, let me know and I can add clarifying details.

@rossbacher This seems to work with proguard. I added a keystore to the counter app and ran proguard on it and it still worked. You are the expert so please let me know if there are any concerns here. The only thing this relies on is the `copy$default` static function and `componentN` functions to exist and have their original name. I think this will also make https://github.com/airbnb/MvRx/pull/331 redundant.

App | Size (kb)
------------ | -------------
App with AppCompat | 2240
w/ RxJava + RxAndroid | 2912
w/ MvRx + kotlin-reflect (1.3.0) | 3848
w/ MvRx + kotlin-metadata-jvm | 3340
w/ MvRx + this change | 3000

Thanks to @ZacSweers for pointing me in the direction of `copy$default` and the property mask.

Fixes https://github.com/airbnb/MvRx/issues/274


mvrx/src/main/kotlin/com/airbnb/mvrx/BaseMvRxViewModel.kt
```diff
@@ -70,26 +64,6 @@ abstract class BaseMvRxViewModel<S : MvRxState>(
...
-     * Kotlin reflection has a large overhead the first time you run it
-     * but then is pretty fast on subsequent times. Running these methods now will
-     * initialize kotlin reflect and warm the cache so that when persistState() gets
-     * called synchronously in onSaveInstanceState() on the main thread, it will be much faster.
-     * This improved performance 10-100x for a state with 100 @PersistState properties.
-     *
-     * This is also @Synchronized to prevent a ConcurrentModificationException in kotlin-reflect: https://gist.github.com/gpeal/27a5747b3c351d4bd592a8d2d58f134a
-     */
-    @Synchronized
-    fun warmReflectionCache(initialState: S) {
```
gpeal: @tasomaniac You'll appreciate this.

mvrx/src/main/kotlin/com/airbnb/mvrx/PersistState.kt
```diff
@@ -73,49 +72,71 @@ private fun assertCollectionPersistability(value: Any?) {
...
-        .filter { persistedConstructorParamNames.contains(it.name) }
-        .fold(mutableMapOf<KParameter, Any?>()) { map, param ->
-            // We do the containsKey check to differentiate between a missing key and one that is explicitly null.
-            if (containsKey(param.name)) map[param] = this[param.name]
-            map
+    // There is 1 bitmask for each block of 32 parameters.
+    val parameterBitMasks = IntArray(fieldCount / 32 + 1) { 0 }
+    val parameters = arrayOfNulls<Any?>(fieldCount)
+    parameters[0] = initialState
+    for (i in 0..(fieldCount - 1)) {
```
elihart: iterate  from 1? the first parameter is the initial state

mvrx/src/main/kotlin/com/airbnb/mvrx/PersistState.kt
```diff
@@ -73,49 +72,71 @@ private fun assertCollectionPersistability(value: Any?) {
...
+        val bundleKey = i.toString()
+        if (containsKey(bundleKey)) {
```
elihart: should we log a warning if it doesn't exist? seems like that shouldn't happen

mvrx/src/main/kotlin/com/airbnb/mvrx/PersistState.kt
gpeal: @elihart The first parameter of the copy function is the initial state but this is iterating over the constructor parameters

mvrx/src/main/kotlin/com/airbnb/mvrx/MvRxMutabilityHelper.kt
```diff
@@ -27,31 +25,52 @@ private const val IMMUTABLE_MAP_MESSAGE =
...
+        }
 }
 
+/**
+ * Since we can only use java reflection, this basically duck types a data class.
+ * componentN methods are also used for @PersistState.
+ */
+private val Class<*>.isData: Boolean
+    get() {
+        declaredMethods.firstOrNull { it.name == "component1" } ?: return false
```
tasomaniac: Very small detail: `find` can be used instead of `firstOrNull`. They do the same. `find` is just read better in this case.

mvrx/src/main/kotlin/com/airbnb/mvrx/PersistState.kt
```diff
@@ -31,30 +28,31 @@ annotation class PersistState
...
-        return Bundle()
-    }
+internal fun <T : MvRxState> T.persistState(validation: Boolean = false): Bundle {
+    val jvmClass = this::class.java
+    // Find the first constructor annotated with @PersistState or return.
+    val constructor = jvmClass.constructors.firstOrNull { it.parameterAnnotations.any { it.any { it is PersistState } } } ?: return Bundle()
+
+    val bundle = Bundle()
+    constructor.parameterAnnotations.forEachIndexed { i, p ->
+        if (!p.any { it is PersistState }) return@forEachIndexed
```
tasomaniac: `!p.any` can be replaced by `p.none { }`

mvrx/src/main/kotlin/com/airbnb/mvrx/MvRxMutabilityHelper.kt
gpeal: @tasomaniac IMO, it's just a matter of whichever convention the codebase uses. I wish there was only 1 in the stdlib since they are the same. In the MvRx codebase, firstOrNull is used so I went for consistency.

mvrx/src/main/kotlin/com/airbnb/mvrx/MvRxMutabilityHelper.kt
```diff
@@ -27,31 +25,52 @@ private const val IMMUTABLE_MAP_MESSAGE =
...
-            prop.isSubtype(SparseArrayCompat::class) -> "You cannot use SparseArrayCompat for ${prop.name}.\n$IMMUTABLE_LIST_MESSAGE"
-            prop.isSubtype(ArrayMap::class) -> "You cannot use ArrayMap for ${prop.name}.\n$IMMUTABLE_MAP_MESSAGE"
-            Build.VERSION.SDK_INT >= Build.VERSION_CODES.KITKAT &&
-                prop.isSubtype(android.util.ArrayMap::class) -> "You cannot use ArrayMap for ${prop.name}.\n$IMMUTABLE_MAP_MESSAGE"
-            prop.isSubtype(HashMap::class) -> "You cannot use HashMap for ${prop.name}.\n$IMMUTABLE_MAP_MESSAGE"
-            prop.isSubtype(Function::class, KCallable::class) -> {
-                "You cannot use functions inside MvRx state. Only pure data should be represented: ${prop.name}"
+    fun Field.isSubtype(vararg classes: KClass<*>): Boolean {
+        return classes.any { klass ->
+            return when (val returnType = this.type) {
```
ZacSweers: Any concern about type variables here?

mvrx/src/main/kotlin/com/airbnb/mvrx/MvRxMutabilityHelper.kt
```diff
@@ -27,31 +25,52 @@ private const val IMMUTABLE_MAP_MESSAGE =
...
+                else -> false
             }
-            else -> null
-        }?.let { throw IllegalArgumentException(it) }
+        }
     }
+
+    java.declaredFields
+        // During tests, jacoco can add a transient field called jacocoData.
+        .filter { !Modifier.isTransient(it.modifiers) }
```
ZacSweers: nit: `filterNot { Modifier.isTransient(it.modifiers) }`

mvrx/src/main/kotlin/com/airbnb/mvrx/MvRxMutabilityHelper.kt
```diff
@@ -27,31 +25,52 @@ private const val IMMUTABLE_MAP_MESSAGE =
...
+                else -> null
+            }?.let { throw IllegalArgumentException(it) }
+        }
 }
 
+/**
+ * Since we can only use java reflection, this basically duck types a data class.
+ * componentN methods are also used for @PersistState.
+ */
+private val Class<*>.isData: Boolean
```
ZacSweers: I think a better thing to check for here would be the `copy` method

mvrx/src/main/kotlin/com/airbnb/mvrx/PersistState.kt
```diff
@@ -31,30 +28,31 @@ annotation class PersistState
...
-    val persistStateArgs = klass.primaryConstructor
-            ?.parameters
-            ?.filter { it.annotations.any { it.annotationClass == PersistState::class } }
-    if (persistStateArgs?.isEmpty() != false) {
-        return Bundle()
-    }
+internal fun <T : MvRxState> T.persistState(validation: Boolean = false): Bundle {
+    val jvmClass = this::class.java
+    // Find the first constructor annotated with @PersistState or return.
+    val constructor = jvmClass.constructors.firstOrNull { it.parameterAnnotations.any { it.any { it is PersistState } } } ?: return Bundle()
```
ZacSweers: The comment here says it's looking for constructors annotated with `@PersistState`, but the implementation appears to be looking for the first constructor with any _parameters_ with that annotation?

mvrx/src/main/kotlin/com/airbnb/mvrx/PersistState.kt
ZacSweers: Could this check be skipped since you've already done it above?

mvrx/src/main/kotlin/com/airbnb/mvrx/PersistState.kt
```diff
@@ -31,30 +28,31 @@ annotation class PersistState
...
+    constructor.parameterAnnotations.forEachIndexed { i, p ->
+        if (!p.any { it is PersistState }) return@forEachIndexed
+        // For each parameter in the constructor, there is a componentN function becasuse state is a data class.
+        // We can rely on this to be true because the MvRxMutabilityHelpers asserts that the state class is a data class.
+        val getter = try {
+            jvmClass.getDeclaredMethod("component${i + 1}").also { it.isAccessible = true }
+        } catch (e: NoSuchMethodException) {
+            if (validation) throw e
+            return Bundle()
+        } catch (e: SecurityException) {
```
ZacSweers: Can this happen on android? 🤔

mvrx/src/main/kotlin/com/airbnb/mvrx/PersistState.kt
```diff
@@ -31,30 +28,31 @@ annotation class PersistState
...
+    val constructor = jvmClass.constructors.firstOrNull { it.parameterAnnotations.any { it.any { it is PersistState } } } ?: return Bundle()
+
+    val bundle = Bundle()
+    constructor.parameterAnnotations.forEachIndexed { i, p ->
+        if (!p.any { it is PersistState }) return@forEachIndexed
+        // For each parameter in the constructor, there is a componentN function becasuse state is a data class.
+        // We can rely on this to be true because the MvRxMutabilityHelpers asserts that the state class is a data class.
+        val getter = try {
+            jvmClass.getDeclaredMethod("component${i + 1}").also { it.isAccessible = true }
+        } catch (e: NoSuchMethodException) {
```
ZacSweers: Might be better to just iterate over the methods and find it yourself rather than a full try/catch?

mvrx/src/main/kotlin/com/airbnb/mvrx/PersistState.kt
```diff
@@ -73,49 +71,75 @@ private fun assertCollectionPersistability(value: Any?) {
...
+
+    // For data classes, Kotlin generates a static functionc called copy$default.
+    // The first parameter is the object to copy from.
+    // The next parameters are all of parameters to copy (it's jvm bytecode/java so there are no optional parameters in the generated method).
+    // The next parameter(s) are a bitmask. Each parameter index corresponds to one bit in the int.
+    //     If the bitmask is 1 for a given parameter then the new object will have the original object's value and the parameter value will be ignored.
+    //     If the bitmask is 0 then it will use the value from the parameter.
+    //     There is 1 bitmask for every 32 parameters. If there are 48 parameters, there will be 2 bitmasks. Parameter 33 will be the first bit of the 2nd bitmask.
+    // The last parameter is ignored. It can be null.
+    val copyFunction = jvmClass.declaredMethods.first { it.name == "copy\$default" }
```
ZacSweers: Probably worth adding a proguard rule for this. Combine it with the persisted state base class

```
-if class * extends path.to.MvRxState
-keepclassmembers @path.to.PersistedState class <1> {
    synthetic <1> copy$default(...);
}
```

mvrx/src/main/kotlin/com/airbnb/mvrx/PersistState.kt
```diff
@@ -124,6 +148,25 @@ internal fun <T : Any> Bundle.restorePersistedState(initialState: T): T {
...
+    fun <T : MvRxState> restorePersistedState(bundle: Bundle, initialState: T, validation: Boolean = false) = bundle.restorePersistedState(initialState, validation)
+}
+
+/**
+ * Throws [NoSuchElementException] if there is no method.
+ */
+@Suppress("UNCHECKED_CAST")
+private fun <T : Any> Class<T>.copyMethod(): Method = this.declaredMethods.first { it.name == "copy\$default" }
+
+private val Class<*>.defaultParameterValue: Any?
```
ZacSweers: Do you need to handle boxed vs not boxed here? We had to handle that in Moshi because it would fail at runtime otherwise.

https://github.com/square/moshi/blob/master/kotlin/codegen/src/main/java/com/squareup/moshi/kotlin/codegen/api/kotlintypes.kt#L84-L91

mvrx/src/main/kotlin/com/airbnb/mvrx/BaseMvRxViewModel.kt
```diff
@@ -164,12 +137,13 @@ abstract class BaseMvRxViewModel<S : MvRxState>(
      * a fair amount of reflection.
      */
     private fun validateState(initialState: S) {
-        if (state::class.visibility != KVisibility.PUBLIC) {
-            throw IllegalStateException("Your state class ${state::class.qualifiedName} must be public.")
+        if (!Modifier.isPublic(state::class.java.modifiers)) {
```
ZacSweers: Slight semantic change here - `internal` would not have returned true before. Maybe that's fine, but worth noting in release notes.

rossbacher:
@gpeal Yes, I'll have a deeper look in a bit into this but you are right that this will make #331 obsolete.

rossbacher:
@gpeal I think we can also get rid of some of the config that is now no needed anymore, I already made some progress today with that (which made the sample app a bit smaller as fewer things actually need to be kept).
But I ran into a few instances where I don't understand yet, why the rules are still needed (e.g. for the cache prewarming)

I need to dig a bit deeper on Monday.

gpeal:
@rossbacher The cache warning shouldn't be needed anymore and was removed in this PR
https://github.com/airbnb/MvRx/pull/334/files#diff-d6da0afe1caeee8562010718cd2c4257L83

rossbacher:
@gpeal Exactly that confused me. Looking into that.

rossbacher:
@gpeal I updated the proguard config of sample and the lib itself to reflect the removal of Kotlin reflection This should again help to make apps smaller. Though very little in the demo apps as they only have very few States and Models.

gpeal:
Thanks @rossbacher!

@BenSchwab @elihart want to also review this?

mvrx/src/main/kotlin/com/airbnb/mvrx/PersistState.kt
rossbacher: @gpeal @ZacSweers

The old rule would have worked but it is not optimal and I also had updated it for this change already. 
I retested and updated again to:
```
# The constructor as well as the copy$default() method and the component*() methods of the Kotlin data classes used as
# the state in MvRx are read via reflection which cause trouble with Proguard if they are not kept.
-keepclassmembers class ** implements com.airbnb.mvrx.MvRxState {
   public <init>(...);
   synthetic *** copy$default(...);
   public *** component*();
}
```
this should take care of all of those.

rossbacher:
> Thanks @rossbacher!
> 
> @BenSchwab @elihart want to also review this?

Actually I has missed @ZacSweers comment, just addressed it and updated the config.

jess-anders:
This is amazing @gpeal !  Thanks!  We are just starting to use MvRx, and so far it's been great.  Waiting on this change to get released before we can merge and push out in a release though.  Any idea on the timeline for this getting merged and released?

Issue #305, opened 2019-10-09
Nav Graph Scoped View Models?


nealsanche:
MvRx has activity, fragment, parentFragment, and targetFragment scoped viewModel delegate functions. Not sure what the last one does, as I haven't seen an example.

Navigation Component has the ability to get a ViewModel that is scoped to a particular nested navigation graph. This is super handy when executing a flow that you need to do again and again, and don't want the viewModel to be reused between flows.

I know that using fragmentViewModel on the start fragment in the flow, and existingViewModel in the sub-parts of the flow can also do something similar to nav graph scoped view models.

But, where it gets fun is when you're trying to test fragments in isolation using something like FragmentScenario. existingViewModel will actually crash in this case because you might be testing the second fragment in a flow. Which actually brings to mind two specific questions.

How could a ViewModel be created while doing an Espresso test so that existingViewModel would succeed instead of crashing -- remember I'm testing fragments in isolation?

Is it possible to integrate nav graph scoped view models with MvRx in a simple way?

elihart:
The testing system I'm working on with https://github.com/airbnb/MvRx/pull/282 would help to test fragments with existing view models, but I see how nav graph scope would still be helpful.

We don't use the navigation architecture component, so it's a little hard to comment on it. Based on https://developer.android.com/reference/kotlin/androidx/navigation/package-summary#navgraphviewmodels what you are doing seems reasonable.

we could maybe either expose an internal version of `mvrxFactory` to make this easier for people to use, or add a navigation module that provides this function.

It seems simple enough to use directly like this for now, but it is good to think about providing this from the library at some point. Thanks for sharing!

marukami:
> It seems simple enough to use directly like this for now, but it is good to think about providing this from the library at some point. Thanks for sharing!

I spiked out a [prototype](https://github.com/airbnb/MvRx/compare/master...marukami:navigation-viewmodel-store) and it should be possible without changing any fundamentals to have something like this. 
```kotlin
private val viewModel: MyDetailViewModel by navGraphViewModel(R.id.my_graph_id)
``` 
 
@elihart would this approach make sense when navigation to the library?

iMvBremen:
[...]

PR #337, opened 2020-01-05
Add Jetpack navigation module

marukami:
Adding support for the Navigation Graph #305 is a little tricky. The issue you run into is cause we need the navigation controller, and, so this means when restoring from process death, the fragments are restored in the activity#onCreate super before the activity has inflated its content. So, the `NavigationController` is not always accessible in Fragment#onCreate. When not restoring from process death, the existing ViewModel delegates method of creating the ViewModel like `fragmentViewModel` works fine.

For now, my current workaround is to remove the lifecycle aware lazy and use Lazy. I think it should be possible to make navigation ViewModel delegate auto-subscribe by delaying the subscription until `onStart` when the `NavigationController` is not yet accessible.

I still need to add some CI tests, but, I thought I would open the PR now to get some feedback on my current approach and if it makes sense to have navigation as an add-on module.


nealsanche:
Personally, I like this direction. We were definitely having difficulties with restoring after process death, and if this solves that issue, it'd be very welcome.

mvrx/src/main/kotlin/com/airbnb/mvrx/MvRxViewModelFactory.kt
```diff
@@ -101,11 +101,13 @@ data class FragmentViewModelContext(
     /**
      * The fragment owner of the ViewModel.
      */
-    val fragment: Fragment
+    val fragment: Fragment,
+    private val ownerProvider: ViewModelStoreOwner = fragment,
+    private val savedStateRegistryProvider: SavedStateRegistry = fragment.savedStateRegistry
```
mradzinski: Shouldn't this be `private val savedStateRegistryProvider: SavedStateRegistry = ownerProvider.savedStateRegistry`? Just wondering what would happen If you assign an `ownerProvider` but not a `savedStateRegistryProvider` .

kdreamix:
Illegal state exception will still be thrown in this approach on configuration changed

marukami:
> Illegal state exception will still be thrown in this approach on configuration changed

@kdreamix Good to know thanks for the post.

kdreamix:
Any idea on how to work around that?

marukami:
> Any idea on how to work around that?

At the time you asked no. I think I have a working prototype now. 

Sorry I Accidentally deleted this PR 😅 

The MvRx view needs to be subscribed to the ViewModel and there seems to be two ways to trigger this safely. 
1. `try/catch` in the `lifecyleOwner` onCreate, and fallback to onStart if `onCreate` fails
2. Access the ViewModel from the Fragment in or after and only after the `onActivityCreated`, forcing the lazy initializer to fire.

I opted to use both as some fragments might only ever use invalidate.

PR #345, opened 2020-01-24
Catch class `ClassNotFoundException` when loading classes.

carlosmuvi:
Fixes https://github.com/airbnb/MvRx/issues/335.


carlosmuvi:
Definitely! on it.

carlosmuvi:
Ready for another look.

Issue #130, opened 2018-11-01
[Open Question] BaseMvRxViewModel first class coroutine support


gpeal:
We've already made MvRx work great with rxjava by adding things like `disposeOnClear` and `execute`.
I think we could add first class support for coroutines by:
* Making BaseMvRxViewModel a CoroutineScope
* Making an `execute` on `Deferred<T>` which would look something like (untested code):
```
    protected suspend fun <T> Deferred<T>.execute(reducer: S.(Async<T>) -> S) {
        setState { reducer(Loading()) }
        try {
            val value = await()
            setState { reducer(Success(value)) }
        } catch (e: Exception) {
            setState { reducer(Fail(e)) }
        }
    }
```

@hellohuanlin @elihart @BenSchwab @ATizik

ATizik:
It's preferable to avoid Deferred, maybe it can look like this?

`protected suspend fun <T> (suspend ()->T).execute(reducer: S.(Async<T>) -> S) {
    setState { reducer(Loading()) }
    try {
        val value = this()
        setState { reducer(Success(value)) }
    } catch (e: Exception) {
        setState { reducer(Fail(e)) }
    }
}`

Also when [this issue](https://github.com/Kotlin/kotlinx.coroutines/issues/254) is resolved, additional `execute` extensions can be added

gpeal:
@ATizik Why is it preferable to avoid `Deferred`?

ATizik:
@gpeal 
In idiomatic coroutine code instead of
`fun foo(): Deferred<Int> = TODO()`
it's better to use
`suspend fun foo(): Int = TODO()`

[Here](https://github.com/Kotlin/kotlinx.coroutines/issues/342#issuecomment-383841713) is more info

gpeal:
@ATizik Makes sense. Some things like the [retrofit coroutines adapter](https://github.com/JakeWharton/retrofit2-kotlin-coroutines-adapter) will return `Deferred<T>` though.

razilsh:
@gpeal Retrofit now support suspend functions 😄

elihart:
it would be great to do this once the new Jetpack provided ViewModel coroutine scope is out of beta.

It seems reasonable to provide an extension to both `Deferred<T>` and (suspend ()->T)`.

I think we should also have the function return a `Job` so the user can cancel it if needed. In the try/catch block do we need to handle CancellationException any differently if the job is canceled, and we don't want it to trigger the fail state?

[...]

PR #347, opened 2020-01-27
Replaced internal Rx with Coroutines

### LINKED ISSUE #79: MvRxStateStore based on kotlin coroutines
Since MvRxStateStore can be a performance bottleneck on screens with state changing very frequently, maybe it would be beneficial to implement it with coroutines instead of Rx? 
I can provide performance comparisons to determine actual performance gain in case it's needed

### LINKED ISSUE #296: Wrong comment in real mvrx state store
https://github.com/airbnb/MvRx/blob/8375363d29cdab3561b5259118c2c2e4190cc562/mvrx/src/main/kotlin/com/airbnb/mvrx/RealMvRxStateStore.kt#L24 the comment refers to an observer while the next line is a disposable

gpeal:
This PR converts the internal implementation from RxJava to Coroutines. This was not designed to be a breaking change so no external APIs still use Disposable.

However, a new API was added: `ViewModel.stateFlow` This property allows consumers to easily combine MvRx state with other operations or chain it to other Flow operators. Thoughts on this API? It doesn't violate the primary concern of providing synchronous state access from within a ViewModel.

Once this is landed, I think it is time to create the `mvrx-rxjava2` artifact and start splitting off RxJava for good. Thoughts?

Fixes #296
Fixes #79 
#130


mvrx/src/main/kotlin/com/airbnb/mvrx/CoroutinesStateStore.kt
```diff
@@ -0,0 +1,70 @@
...
+        scope.coroutineContext.cancel()
+        withStateChannel.cancel()
+        flushQueuesChannel.cancel()
+        stateChannel.cancel()
+    }
+
+    companion object {
+        var scopeFactory: (() -> CoroutineScope)? = null
+
+        fun coroutineScope() = scopeFactory?.invoke() ?: CoroutineScope(newSingleThreadContext("CoroutineStateStore") + SupervisorJob())
```
elihart: there is already a standard lib function named `coroutineScope` so it is a little confusing to overload that term here. consider making this more explicitly named

mvrx/src/main/kotlin/com/airbnb/mvrx/CoroutinesStateStore.kt
```diff
@@ -0,0 +1,70 @@
...
+
+    override fun set(stateReducer: S.() -> S) {
+        setStateChannel.offer(stateReducer)
+        flushQueuesChannel.offer(Unit)
+    }
+
+    override val flow get() = stateChannel.asFlow().distinctUntilChanged()
+
+    override fun cancel() {
+        scope.coroutineContext.cancel()
```
elihart: does the scope need to be cleaned up so the thread is released?

mvrx/src/main/kotlin/com/airbnb/mvrx/CoroutinesStateStore.kt
elihart: will this create a new thread for each state store? is that overkill?

mvrx/src/main/kotlin/com/airbnb/mvrx/CoroutinesStateStore.kt
```diff
@@ -0,0 +1,70 @@
...
+        withStateChannel.offer(block)
+        flushQueuesChannel.offer(Unit)
+    }
+
+    override fun set(stateReducer: S.() -> S) {
+        setStateChannel.offer(stateReducer)
+        flushQueuesChannel.offer(Unit)
+    }
+
+    override val flow get() = stateChannel.asFlow().distinctUntilChanged()
```
elihart: prefer explicit types for this public API

mvrx/src/main/kotlin/com/airbnb/mvrx/CoroutinesStateStore.kt
gpeal: @elihart That is by design and what we do today. Each RealMvRxStateStore uses a `Schedulers.newThread()`. This 1) ensures correct ordering from within a ViewModel and 2) prevents one slow ViewModel from slowing down others.

gpeal:
@elihart This is ready for another look. I still can't get the exceptions to properly propagate for tests. Any eyes there are welcome.

jpventura:
Noob suggestion: instead of creating `mvrx-rxjava2`

  - Create `mvrk` (or `mvrc`).
  - Refactor `mvrx` build over `mvrk` layer and `kotlinx-coroutines-rx2`

One could still infers MvRx pun _semantically_ suggests RxJava dependencies. MvRk could still be read Maveri**K** and also be a pun to its **K**otlin coroutines dependency :thinking:.

mvrx/src/main/kotlin/com/airbnb/mvrx/CoroutinesStateStore.kt
```diff
@@ -0,0 +1,85 @@
...
+    private val withStateChannel = Channel<(S) -> Unit>(capacity = Channel.UNLIMITED)
+    private val flushQueuesChannel = Channel<Unit>(capacity = Channel.CONFLATED)
+
+    private val stateChannel = ConflatedBroadcastChannel<S>(initialState)
+    override val state: S get() = stateChannel.valueOrNull ?: error("There is no state yet! This should never happen.")
+
+    init {
+        scope.launch {
+            flushQueuesChannel.consumeEach {
+                while (!setStateChannel.isEmpty || !withStateChannel.isEmpty) {
```
elihart: prefer `isNotEmpty()`?

mvrx/src/main/kotlin/com/airbnb/mvrx/MvRxLifecycleAwareFlow.kt
```diff
@@ -0,0 +1,70 @@
...
+ * This function suspends until the next time the owner is stopped.
+ */
+private suspend fun LifecycleOwner.launchWhenStartedAndCancelWhenStopped(block: suspend () -> Unit) {
+    var parentJob = SupervisorJob()
+    lifecycle.addObserver(object : LifecycleEventObserver {
+        override fun onStateChanged(source: LifecycleOwner, event: Lifecycle.Event) {
+            when (event) {
+                Lifecycle.Event.ON_START -> lifecycleScope.launch(parentJob) { block() }
+                Lifecycle.Event.ON_STOP -> {
+                    lifecycle.removeObserver(this)
```
elihart: if added after onStop this is never removed, not sure if that is a problem

mvrx/src/main/kotlin/com/airbnb/mvrx/MvRxLifecycleAwareFlow.kt
```diff
@@ -0,0 +1,70 @@
...
+import kotlinx.coroutines.channels.Channel
+import kotlinx.coroutines.flow.Flow
+import kotlinx.coroutines.flow.channelFlow
+import kotlinx.coroutines.flow.launchIn
+import kotlinx.coroutines.flow.onEach
+import kotlinx.coroutines.launch
+
+/**
+ * Similar to LifecycleScope.launchWhenStarted but cancels the block when the owner is stopped.
+ * This function suspends until the next time the owner is stopped.
```
elihart: `suspends until the next time the owner is stopped` is this right? should this be "started"?

mvrx/src/main/kotlin/com/airbnb/mvrx/CoroutinesStateStore.kt
```diff
@@ -0,0 +1,85 @@
...
+    private val setStateChannel = Channel<S.() -> S>(capacity = Channel.UNLIMITED)
+    private val withStateChannel = Channel<(S) -> Unit>(capacity = Channel.UNLIMITED)
+    private val flushQueuesChannel = Channel<Unit>(capacity = Channel.CONFLATED)
+
+    private val stateChannel = ConflatedBroadcastChannel<S>(initialState)
+    override val state: S get() = stateChannel.valueOrNull ?: error("There is no state yet! This should never happen.")
+
+    init {
+        scope.launch {
+            flushQueuesChannel.consumeEach {
```
elihart: I was thinking about how we could simplify this. I think we could do something like this
```
sealed class StateAction
class SetState(reducer) : StateAction()
class GetState(block) : StateAction()

private val stateActionChannel = PriorityChannel<StateAction>(comparator = { // state actions get priority }
```

A bit of psuedo code, but any way, leverage sealed classes to unify into a single channel and then use an idea like PriorityQueue to poll and make sure we get all SetState actions first.

Main problem is coroutines don't seem to provide something out of the box like "PriorityChannel" although this thread gave me some ideas https://github.com/Kotlin/kotlinx.coroutines/issues/238 and this library has an implementation 
https://github.com/kerubistan/kroki/blob/master/kroki-coroutines/src/main/kotlin/io/github/kerubistan/kroki/coroutines/Channels.kt

Since we probably don't want to implement our own, I think going with the `select` approach here would be nice https://github.com/Kotlin/kotlinx.coroutines/issues/238#issuecomment-368080421

Then we still need a separate channel for get/set, but we can get rid of the flush channel and just select between get/set

mvrx/src/main/kotlin/com/airbnb/mvrx/CoroutinesStateStore.kt
```diff
@@ -0,0 +1,85 @@
...
+    }
+
+    override fun get(block: (S) -> Unit) {
+        if (!scope.isActive) return
+        withStateChannel.offer(block)
+        flushQueuesChannel.offer(Unit)
+    }
+
+    override fun set(stateReducer: S.() -> S) {
+        if (!scope.isActive) return
```
elihart: is this necessary? would it be better to fail if this is called after the viewmodel is cleared/canceled?

mvrx/src/main/kotlin/com/airbnb/mvrx/MvRxStateStore.kt
```diff
@@ -1,11 +1,11 @@
...
-import io.reactivex.disposables.Disposable
+import kotlinx.coroutines.flow.Flow
 
-interface MvRxStateStore<S : Any> : Disposable {
+interface MvRxStateStore<S : Any> {
     val state: S
     fun get(block: (S) -> Unit)
     fun set(stateReducer: S.() -> S)
-    val observable: Observable<S>
+    fun cancel()
```
elihart: it seems antithetical to coroutines to have an explicit cancel function

mvrx/src/test/kotlin/com/airbnb/mvrx/BaseTest.kt
```diff
@@ -1,62 +1,42 @@
...
-            RxJavaPlugins.setInitIoSchedulerHandler { immediate }
-            RxJavaPlugins.setSingleSchedulerHandler { immediate }
-            // This is necessary to prevent rxjava from swallowing errors
-            // https://github.com/ReactiveX/RxJava/issues/5234
-            Thread.setDefaultUncaughtExceptionHandler { _, e ->
-                if (e is CompositeException && e.exceptions.size == 1) throw e.exceptions[0]
-                throw e
-            }
-            RxAndroidPlugins.setInitMainThreadSchedulerHandler { immediate }
+            CoroutinesStateStore.scopeFactory = { CoroutineScope(TestCoroutineDispatcher()) }
```
elihart: this is nicer :D

mvrx/src/main/kotlin/com/airbnb/mvrx/BaseMvRxViewModel.kt
```diff
@@ -58,16 +65,17 @@ abstract class BaseMvRxViewModel<S : MvRxState>(
...
+            }
         }
     }
 
     @CallSuper
     override fun onCleared() {
         super.onCleared()
         disposables.dispose()
-        stateStore.dispose()
+        stateStore.cancel()
```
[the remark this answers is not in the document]
gpeal: @elihart Maybe you can think of a better way but there is an initialization issue trying to do that.
The state store gets constructor injected into the view model so its scope/parent job isn't accessible yet.
What do you want to do here?

mvrx/src/main/kotlin/com/airbnb/mvrx/BaseMvRxViewModel.kt
```diff
@@ -664,56 +672,55 @@ abstract class BaseMvRxViewModel<S : MvRxState>(
...
+    ): Job = if (lifecycleOwner == null || FORCE_DISABLE_LIFECYCLE_AWARE_OBSERVER) {
+        onEach { subscriber(it) }.launchIn(viewModelScope)
+    } else if (deliveryMode is UniqueOnly) {
+        flowWhenStarted<T>(
+            lifecycleOwner,
+            deliveryMode,
+            lastDeliveredValue(deliveryMode),
+            onStart = {
+                if (activeSubscriptions.contains(deliveryMode.subscriptionId)) error(duplicateSubscriptionMessage(deliveryMode))
+                activeSubscriptions += deliveryMode.subscriptionId
```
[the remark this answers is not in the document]
gpeal: It's the same thing (from the stdlib)
```
/**
 * Adds the specified [element] to this mutable collection.
 */
@kotlin.internal.InlineOnly
public inline operator fun <T> MutableCollection<in T>.plusAssign(element: T) {
    this.add(element)
}
```

mvrx/src/main/kotlin/com/airbnb/mvrx/CoroutinesStateStore.kt
gpeal: Channels don't have that

gpeal: I could do `poll() != null` but looking at the impl, I think this is lighter weight

mvrx/src/main/kotlin/com/airbnb/mvrx/CoroutinesStateStore.kt
gpeal: @elihart That seems slightly less optimized because now we'd have to create a container object for each action and the queue would have to be sorted. I'm also not sure it would be easier to read. What do you think?

mvrx/src/main/kotlin/com/airbnb/mvrx/MvRxLifecycleAwareFlow.kt
gpeal: @elihart I moved the remove to ON_DESTROY where I think it should have been

mvrx/src/main/kotlin/com/airbnb/mvrx/MvRxStateStore.kt
gpeal: See above but I also changed how this works in the latest commit

gpeal:
@elihart Pushed a new commit. Could you pull this and see if you can help me fix the one failing test? It's `duplicateUniqueOnlySubscribeThrowIllegalStateException` which correctly throws the expected exception but it doesn't bubble up correctly.

mvrx/src/main/kotlin/com/airbnb/mvrx/CoroutinesStateStore.kt
```diff
@@ -0,0 +1,114 @@
...
+
+@Suppress("EXPERIMENTAL_API_USAGE")
+class CoroutinesStateStore<S : MvRxState>(initialState: S) : MvRxStateStore<S> {
+
+    /** Channel that serves as a trigger to flush the setState and withState queues. */
+    private val flushQueuesChannel = Channel<Unit>(capacity = Channel.CONFLATED)
+    private val setStateChannel = Channel<S.() -> S>(capacity = Channel.UNLIMITED)
+    private val withStateChannel = Channel<(S) -> Unit>(capacity = Channel.UNLIMITED)
+
+    private val stateChannel = ConflatedBroadcastChannel<S>(initialState)
```
rougsig, line 25: ConflatedBroadcastChannel can drop events. Is it expected behavior for CoroutinesStateStore?
Real example of dropped events can be state updates for `Async`. 

```kotlin
    @Test
    fun test_ConflatedBroadcastChannel_runBlocking() {
        runBlocking {
            val store = CoroutinesStateStore(MvRxStateStoreTestState())
            async {
                store.flow.collect {
                    println("received count=${it.count}")
                }
            }
            async {
                (1..5).forEach {
                    println("send count=$it")
                    store.set { copy(count = it) }
                }
                store.cancel()
            }
        }
        // console out
        // received count=1
        // send count=1
        // send count=2
        // send count=3
        // send count=4
        // send count=5
        // received count=5
    }
```

gpeal: @rougsig great catch. It is fixed with a buffer in the latest code.

mvrx/src/main/kotlin/com/airbnb/mvrx/BaseMvRxViewModel.kt
```diff
@@ -664,56 +692,66 @@ abstract class BaseMvRxViewModel<S : MvRxState>(
...
+    private fun <T> Flow<T>.assertOneActiveSubscription(owner: LifecycleOwner, deliveryMode: UniqueOnly) {
+        owner.lifecycle.addObserver(object : LifecycleObserver {
+            @OnLifecycleEvent(Lifecycle.Event.ON_START)
+            fun onStart() {
+                if (activeSubscriptions.contains(deliveryMode.subscriptionId)) error(duplicateSubscriptionMessage(deliveryMode))
+                activeSubscriptions += deliveryMode.subscriptionId
+            }
+
+            @OnLifecycleEvent(Lifecycle.Event.ON_STOP)
+            fun onStop() {
```
BenSchwab, line 737: This seems a bit different from the old behavior, none seem like a huge deal, but want to document them:

The code used to remove the subscription in dispose - this handled the case the the observable was terminated manually. You could in theory then subscribe again with the same subscription with no issue. We might be able to get the same behavior with `invokeOnCompletion` and checking for the `cancel` flag

The old code would track on subscription immediately, and would clear in destroy. The new code leaves a period where `Lifecycle < on_start`, where you can subscribe with the same id multiple times, but won't receive the error until the lifecycle has been triggered. Maybe it would make more sense to use `ON_CREATE` and `ON_DESTROY` lifecycle events?

mvrx/src/main/kotlin/com/airbnb/mvrx/BaseMvRxViewModel.kt
```diff
@@ -664,56 +692,66 @@ abstract class BaseMvRxViewModel<S : MvRxState>(
...
+                .dropWhile { it == lastDeliveredValue }
+                .flowWhenStarted(lifecycleOwner)
+                .distinctUntilChanged()
+                .onEach { lastDeliveredStates[deliveryMode.subscriptionId] = it }
+        } else {
+            flowWhenStarted(lifecycleOwner)
+        }
+        return flow
+            .onEach { subscriber(it) }
+            .launchIn(viewModelScope)
```
BenSchwab, line 725: Doesn't this leak the fragment? This flow is going to be retained for the life of the view model but references a subscriber (fragment) with a smaller lifecycle. Should this be `lifecycleOwner.lifecycleScope`?

mvrx/src/main/kotlin/com/airbnb/mvrx/CoroutinesStateStore.kt
```diff
@@ -0,0 +1,112 @@
...
+
+@Suppress("EXPERIMENTAL_API_USAGE")
+class CoroutinesStateStore<S : MvRxState>(
+    initialState: S,
+    scope: CoroutineScope
+) : MvRxStateStore<S> {
+
+    /** Channel that serves as a trigger to flush the setState and withState queues. */
+    private val flushQueuesChannel = Channel<Unit>(capacity = Channel.CONFLATED)
+    private val setStateChannel = Channel<S.() -> S>(capacity = Channel.UNLIMITED)
```
BenSchwab, line 23: What benefit do we get making these channels vs. queues? Seems like are justing using poll/offer

mvrx/src/main/kotlin/com/airbnb/mvrx/CoroutinesStateStore.kt
```diff
@@ -0,0 +1,112 @@
...
+    private val withStateChannel = Channel<(S) -> Unit>(capacity = Channel.UNLIMITED)
+
+    private val stateChannel = BroadcastChannel<S>(capacity = Channel.BUFFERED)
+    override var state = initialState
```
elihart, line 28: does this need to be a var? can we not get the last value off of the flow or channel?

if a var, should at least make it `private set`?

mvrx/src/main/kotlin/com/airbnb/mvrx/BaseMvRxViewModel.kt
```diff
@@ -28,9 +39,10 @@ import kotlin.reflect.KProperty1
 abstract class BaseMvRxViewModel<S : MvRxState>(
     initialState: S,
     debugMode: Boolean,
-    private val stateStore: MvRxStateStore<S> = RealMvRxStateStore(initialState)
+    stateStoreOveride: MvRxStateStore<S>? = null
```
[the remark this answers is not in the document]
gpeal: @elihart I can make the scope injectable but it should be made clear that the ViewModel owns the scope and it will be cancelled in onClear so you shouldn't pass in another scope like a Fragment's scope

mvrx/src/main/kotlin/com/airbnb/mvrx/BaseMvRxViewModel.kt
```diff
@@ -222,7 +250,7 @@ abstract class BaseMvRxViewModel<S : MvRxState>(
      * For ViewModels that want to subscribe to itself.
      */
     protected fun subscribe(subscriber: (S) -> Unit) =
```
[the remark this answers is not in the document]
gpeal: @elihart There will be a number of follow up PRs to add these but wanted to start with just the internal impl 😄

mvrx/src/main/kotlin/com/airbnb/mvrx/CoroutinesStateStore.kt
```diff
@@ -0,0 +1,112 @@
...
+    override val flow: Flow<S> get() = channelFlow {
```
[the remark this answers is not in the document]
gpeal: @elihart StateFlow conflates sets so if it gets set very quickly, an intermediate state can get conflated before consumers pick it up. An earlier commit here has that impl but I was able to create a failing test that demonstrated that case.

mvrx/src/main/kotlin/com/airbnb/mvrx/CoroutinesStateStore.kt
```diff
@@ -0,0 +1,112 @@
...
+
+            withStateChannel.poll()?.let { withStateBlock ->
+                withStateBlock(state)
+            }
+        }
+    }
+
+    override fun get(block: (S) -> Unit) {
+        withStateChannel.offer(block)
+        if (MvRxTestOverrides.FORCE_SYNCHRONOUS_STATE_STORES) {
```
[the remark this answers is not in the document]
gpeal: @elihart Sweet! Should we leave this here for now then you can clean it up when we land that branch?

mvrx/src/main/kotlin/com/airbnb/mvrx/CoroutinesStateStore.kt
gpeal: @elihart It does. ArrayBroadcastChannel is not a ConflatedBroadcastChannel nor StateFlow which provides access to the most recent value.
This was one of the trickiest parts or the PR because it needs to:
1) Provide the most recent value
2) Be resilient against slow consumers (aka be buffered)
3) Multicast

This was the only solution I could come up with that achieves all 3.

mvrx/src/main/kotlin/com/airbnb/mvrx/BaseMvRxViewModel.kt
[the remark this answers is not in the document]
elihart: @gpeal that sounds good, what do you think @BenSchwab ?

This is what we do
```
    private val customCoroutineContext: CoroutineContext = inject(MvRxDagger.AppGraph::viewModelCoroutineContext).value
) : CoroutineScope by CoroutineScope(customCoroutineContext + SupervisorJob()), BaseMvRxViewModel<S>(
```

mvrx/src/main/kotlin/com/airbnb/mvrx/CoroutinesStateStore.kt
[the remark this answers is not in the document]
elihart: @gpeal good to know. may be good to document that here so we know why we chose this implementation

mvrx/src/main/kotlin/com/airbnb/mvrx/CoroutinesStateStore.kt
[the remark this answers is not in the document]
elihart: yep

mvrx/src/main/kotlin/com/airbnb/mvrx/CoroutinesStateStore.kt
elihart: ok. would just request some code comments so we have a history documented of the reasoning

mvrx/src/main/kotlin/com/airbnb/mvrx/BaseMvRxViewModel.kt
```diff
@@ -51,23 +63,39 @@ abstract class BaseMvRxViewModel<S : MvRxState>(
...
+     * Return the current state as a Flow. For certain situations, this may be more convenient
+     * than subscribe and selectSubscribe because it can easily be composed with other
+     * coroutines operations and chained with operators.
+     */
+    val stateFlow: Flow<S>
+        get() = stateStore.flow
+
+    @VisibleForTesting
+    internal val bufferedStateFlow: Flow<S>
+        get() = stateStore.flow.buffer(10)
```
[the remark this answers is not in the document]
gpeal: @BenSchwab I had a comment in a previous pace I had this to explain that it's an arbitrary number. Thoughts?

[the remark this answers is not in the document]
gpeal: I changed this to use the jvm default value

mvrx/src/main/kotlin/com/airbnb/mvrx/CoroutinesStateStore.kt
denis-bezrukov: queues aren't thread safe, while `channel.offer` is thread safe.

mvrx/src/main/kotlin/com/airbnb/mvrx/CoroutinesStateStore.kt
```diff
@@ -0,0 +1,128 @@
...
+     *
+     * StateFlow can't be used because it conflates new sets. If you set its value in a tight
+     * loop and listen to changes on a flow, you will miss values. ConflatedBroadcastChannel
+     * alone has the same issue.
+     *
+     * A normal Channel can't be used because it isn't multicast.
+     */
+    private val stateChannel = BroadcastChannel<S>(capacity = Channel.BUFFERED)
+    override var state = initialState
+    override val flow: Flow<S> get() = channelFlow {
```
denis-bezrukov, line 43: from performance perspective will be better to use `flow { }` builder here, because `channelFlow {}` provides thread-safe ProducerScope, but you don't have concurrency here, and your code is just sequential

mvrx/src/main/kotlin/com/airbnb/mvrx/BaseMvRxViewModel.kt
```diff
@@ -51,23 +83,39 @@ abstract class BaseMvRxViewModel<S : MvRxState>(
...
+     * Return the current state as a Flow. For certain situations, this may be more convenient
+     * than subscribe and selectSubscribe because it can easily be composed with other
+     * coroutines operations and chained with operators.
+     */
+    val stateFlow: Flow<S>
+        get() = stateStore.flow
+
+    @VisibleForTesting
+    internal val bufferedStateFlow: Flow<S>
+        get() = stateStore.flow.buffer()
```
denis-bezrukov, line 103: actually, you don't need to buffer state flow after switching from StateFlow to BroadcastChannel, because you're already have buffer in `CoroutineStateStore` and producer supports backpressure correctly (in [this ](https://github.com/airbnb/MvRx/pull/385/commits/acf024a2c414436fbb131df2e39272b02bd17c26)commit in another PR). 

Buffering state here is actually just increases buffer size twice from 64 to 128 - 64 per subscriber, and 64 common buffer. 
Also, as one more side effect, a 64-element buffer will be created for each subscriber which may not be memory efficient

mvrx/src/main/kotlin/com/airbnb/mvrx/CoroutinesStateStore.kt
```diff
@@ -0,0 +1,128 @@
...
+        send(state)
```
denis-bezrukov, line 44: I'm still concerned the case we discussed when subscriber subscribes to updates in the middle of VM lifecycle and then receives current state, and then few states from the past. 

I think I can create a unit test for that

mvrx/src/main/kotlin/com/airbnb/mvrx/CoroutinesStateStore.kt
```diff
@@ -0,0 +1,128 @@
...
+        stateChannel.consumeEach { send(it) }
+    }.distinctUntilChanged()
```
denis-bezrukov, line 46: If you remove `buffer()` from BaseMvRxViewModel.kt according my comment, you probably need to add buffer(1) here before disticntUntilChanged to consume `send(state)` and subscribe to stateChannel immediately.

mvrx/src/main/kotlin/com/airbnb/mvrx/CoroutinesStateStore.kt
```diff
@@ -0,0 +1,128 @@
...
+                val newState = state.reducer()
+                stateChannel.offer(newState)
+                state = newState
+                reducer = setStateChannel.poll()
+            }
+
+            withStateChannel.poll()?.let { withStateBlock ->
+                withStateBlock(state)
+            }
+        }
```
denis-bezrukov, line 106: the whole thing with prioritization setState over withState can be done via `select` instead of two queues + flush queue or priority queue as were suggested before. 
With `select` no flushQueue necessary, and get/set funs will be just offering to withStateChannel / setStateChannel. 

It could be done after removing FORCE_SYNCHRONOUS_STATE_STORES from this store though.

mvrx/src/main/kotlin/com/airbnb/mvrx/ScriptableMvRxStateStore.kt
```diff
@@ -1,22 +1,23 @@
...
  * A [MvRxStateStore] which ignores standard calls to [set]. Instead it can be scripted via calls to
  * [next]. This is intended to be used for tests only, and in particular UI tests where you wish to test
  * how your UI code reacts to different ViewModel states. This is not as useful for unit testing your view model,
  * as business logic in state reducers will not be used.
  */
+@Suppress("EXPERIMENTAL_API_USAGE")
 class ScriptableMvRxStateStore<S : Any>(initialState: S) : MvRxStateStore<S> {
 
-    private val subject: BehaviorSubject<S> = BehaviorSubject.createDefault(initialState)
+    private val stateFlow = MutableStateFlow(initialState)
```
denis-bezrukov, line 15: Should you switch to BroadcastChannel as well?

mvrx/src/main/kotlin/com/airbnb/mvrx/BaseMvRxViewModel.kt
gpeal: @denis-bezrukov I was thinking that the first buffer would be for fast sets and this one could account for slow consumers. Wouldn't they serve two different purposes?
Without this, what would happen to other consumers if there was one slow one?

mvrx/src/main/kotlin/com/airbnb/mvrx/CoroutinesStateStore.kt
gpeal: @denis-bezrukov `channelFlow` is backed by a buffered channel so this helps against slow consumers. You can change it and see the failing test.

mvrx/src/main/kotlin/com/airbnb/mvrx/BaseMvRxViewModel.kt
gpeal: @denis-bezrukov. I was able to remove this (but kept channelFlow above) and all tests still passed so I think I am okay removing this and saving the memory.

mvrx/src/main/kotlin/com/airbnb/mvrx/CoroutinesStateStore.kt
gpeal: @denis-bezrukov A test would be nice. What do you think an ideal solution would be?

mvrx/src/main/kotlin/com/airbnb/mvrx/BaseMvRxViewModel.kt
denis-bezrukov: @gpeal ArrayBroadcastChannel's buffer behaves the same way - slow consumer won't prevent fast from collecting everything in the buffer. Slow consumer only prevents sending buffer+1 element to it. So adding one more buffer here is just increases the total buffer size, but less effective than making it 128 at BroadcastChannel. 

Sample: total buffer size 10, send 20 items to it:
Channel's buffer only: https://pl.kotl.in/SoU1QXYxH
Channel's buffer + flow buffer (5 + 5): https://pl.kotl.in/8YWk_V25V

In both cases fast consumer quickly consumes first 10 items (total buffer size) and only then it slows down by the slow consumer the same way in both cases.

mvrx/src/main/kotlin/com/airbnb/mvrx/CoroutinesStateStore.kt
```diff
@@ -0,0 +1,120 @@
...
+     *     setState { ... }
+     *     withState { ... }
+     * }
+     */
+    private suspend fun flushQueuesOnce() {
+        select<Unit> {
+            setStateChannel.onReceive { reducer ->
+                val newState = state.reducer()
+                stateChannel.offer(newState)
+                state = newState
```
denis-bezrukov, line 99: I'm thinking about checking that `state != newState` here as previously, to use buffer effectively

mvrx/src/main/kotlin/com/airbnb/mvrx/CoroutinesStateStore.kt
```diff
@@ -0,0 +1,120 @@
...
+     *
+     * A normal Channel can't be used because it isn't multicast.
+     */
+    private val stateChannel = BroadcastChannel<S>(capacity = Channel.BUFFERED)
+    override var state = initialState
+    override val flow: Flow<S>
+        get() = flow {
+            emit(state)
+            stateChannel.consumeEach { emit(it) }
+        }.buffer(1).distinctUntilChanged()
```
denis-bezrukov, line 48: probably worth commenting that's 1 is intentionally and non-random

mvrx/src/main/kotlin/com/airbnb/mvrx/CoroutinesStateStore.kt
```diff
@@ -0,0 +1,120 @@
...
+     * withState {
+     *     setState { ... }
+     *     withState { ... }
+     * }
+     */
+    private suspend fun flushQueuesOnce() {
+        select<Unit> {
+            setStateChannel.onReceive { reducer ->
+                val newState = state.reducer()
+                stateChannel.offer(newState)
```
denis-bezrukov, line 98: offer -> send

mvrx/src/main/kotlin/com/airbnb/mvrx/CoroutinesStateStore.kt
```diff
@@ -0,0 +1,120 @@
...
+class CoroutinesStateStore<S : MvRxState>(
+    initialState: S,
+    scope: CoroutineScope
+) : MvRxStateStore<S> {
+
+    private val setStateChannel = Channel<S.() -> S>(capacity = Channel.UNLIMITED)
+    private val withStateChannel = Channel<(S) -> Unit>(capacity = Channel.UNLIMITED)
+
+    /**
+     * This combination of BroadcastChannel, mutable state property and channelFlow was arrived
```
denis-bezrukov, line 29:
```suggestion
     * This combination of BroadcastChannel, mutable state property and flow was arrived
```

mvrx/src/main/kotlin/com/airbnb/mvrx/CoroutinesStateStore.kt
gpeal: @denis-bezrukov By correctly do you mean that it allows ~64 unique elements to be in the buffer at any given time? Some people's state classes are very large such that hashing it may take a non-trivial amount of time so if it can be limited to a single time (the distinctUntilChanged() on the Flow) that would be ideal.

denis-bezrukov: >64 unique elements to be in the buffer at any given time?

yes

>Some people's state classes are very large such that hashing it may take a non-trivial amount of time so if it can be limited to a single time (the distinctUntilChanged() on the Flow) that would be ideal.

since `flow {}` is cold, `distinctUntilChanged()` on it will be called for every consumer

mvrx/src/main/kotlin/com/airbnb/mvrx/BaseMvRxViewModel.kt
```diff
@@ -28,9 +42,23 @@ import kotlin.reflect.KProperty1
...
+    /**
+     * Provide a default context for viewModelScope. It will be added after [SupervisorJob]
+     * and [Dispatchers.Main.immediate]. This should only be used for tests and should only
+     * be exposed via a shared base class within your app. If your features extend this
+     * directly, do not override this in the primary constructor of your feature ViewModel.
+     */
+    contextOverride: CoroutineContext? = null
 ) : ViewModel() {
     private val debugMode = if (MvRxTestOverrides.FORCE_DEBUG == null) debugMode else MvRxTestOverrides.FORCE_DEBUG
+    protected val viewModelScope = CoroutineScope(SupervisorJob() + Dispatchers.Main.immediate + contextOverride)
```
wlara, line 60: Why not use `androidx.lifecycle.viewModelScope`?

mvrx/src/main/kotlin/com/airbnb/mvrx/CoroutinesStateStore.kt
gpeal: Great point!

mvrx/src/main/kotlin/com/airbnb/mvrx/BaseMvRxViewModel.kt
gpeal: @wlara because in the next PR this won't implement Jetpack ViewModel

CHANGELOG.md
```diff
@@ -1,5 +1,31 @@
...
+            }
+        }
+    }
+}
+```
+Previously, setState would only be prioritized at that nesting level so it would run:
+[W1, S1, W2, S2, S3]
+Now, it will run:
+[W1, S1, S2, S3, W2]
+
```
gpeal, line 28: @BenSchwab @elihart FYI

mvrx/src/main/kotlin/com/airbnb/mvrx/BaseMvRxViewModel.kt
```diff
@@ -664,56 +704,72 @@ abstract class BaseMvRxViewModel<S : MvRxState>(
...
+        val flow = if (lifecycleOwner == null || FORCE_DISABLE_LIFECYCLE_AWARE_OBSERVER) {
+            this
+        } else if (deliveryMode is UniqueOnly) {
+            val lastDeliveredValue: T? = lastDeliveredValue(deliveryMode)
+            this
+                .assertOneActiveSubscription(lifecycleOwner, deliveryMode)
+                .dropWhile { it == lastDeliveredValue }
+                .flowWhenStarted(lifecycleOwner)
+                .distinctUntilChanged()
+                .onEach { lastDeliveredStates[deliveryMode.subscriptionId] = it }
```
denis-bezrukov, line 731: It introduces a behavior change comparing with previous version.
here is a test demonstrates an issue (Put it into FragmentSubscriberTest):

```
    @Test
    fun testUniqueOnly() {
        val (controller, fragment) = createFragmentInTestActivity<ViewSubscriberFragment>()
        assertEquals(0, fragment.selectSubscribeUniqueOnlyValue)

        fragment.setFoo(1)

        assertEquals(1, fragment.selectSubscribeUniqueOnlyValue)
        assertEquals(2, fragment.selectSubscribeUniqueOnlyCallCount)

        controller.pause()
        controller.stop()
        fragment.setFoo(2)
        fragment.setFoo(1)
        controller.start()
        controller.resume()

        assertEquals(1, fragment.selectSubscribeUniqueOnlyValue)
        assertEquals(3, fragment.selectSubscribeUniqueOnlyCallCount)
    }
```

it passed in master, but fails in current branch

mvrx/src/main/kotlin/com/airbnb/mvrx/MvRxLifecycleAwareFlow.kt
```diff
@@ -0,0 +1,33 @@
...
+import kotlinx.coroutines.flow.Flow
+import kotlinx.coroutines.flow.MutableStateFlow
+import kotlinx.coroutines.flow.combineTransform
+
+/**
+ * Emits values from the source flow only when the owner is started.
+ * When the owner transitions to started, the most recent value will be emitted.
+ */
+fun <T : Any> Flow<T>.flowWhenStarted(owner: LifecycleOwner): Flow<T> {
+    val startedFlow = MutableStateFlow(false)
```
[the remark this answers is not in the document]
gpeal: @denis-bezrukov I didn't do this because addObserver should bring it up to date according to the docs
https://developer.android.com/reference/kotlin/androidx/lifecycle/Lifecycle#addobserver

mvrx/src/main/kotlin/com/airbnb/mvrx/BaseMvRxViewModel.kt
gpeal: @denis-bezrukov This is an interesting case. @elihart @BenSchwab What do you think the expected value here should be? In words, if a user subscribes with uniqueOnly(), if the value changes and changes back while the Fragment is stopped, should it redeliver? Previously, it would but now it wouldn't. My hunch says that the new behavior is better. What do you think?

BenSchwab: Ha - this is interesting.. I could go either way..

The old behavior may be useful in some cases, say you selectSubscribeUniqueOnly. And the toast shown is set to `true`, then we go to stop, and then we reset the toast, then something else set's it back to `true`. The expectation may be that the toast is shown again.

gpeal: @BenSchwab I don't know. My gut says that the toast shouldn't be shown because the trigger didn't happen when the toast was shown, it could have happened long before.

If we think this is the last open question in this PR, I think we should land it so subsequent work can begin and consider this case more carefully.

elihart: @gpeal i could also go either way. not sure there is a correct answer here but:
1) we need to make sure to document this very clearly
2) possibly provide a setting or name change besides `uniqueOnly` so users can control this

mvrx/src/main/kotlin/com/airbnb/mvrx/CoroutinesStateStore.kt
elihart: @gpeal is it not possible or necessary to add `private set`?

mvrx/src/main/kotlin/com/airbnb/mvrx/CoroutinesStateStore.kt
```diff
@@ -0,0 +1,128 @@
...
+     *     setState { ... }
+     *     withState { ... }
+     * }
+     */
+    private suspend fun flushQueuesOnce() {
+        select<Unit> {
+            setStateChannel.onReceive { reducer ->
+                val newState = state.reducer()
+                if (newState != state) {
+                    stateChannel.send(newState)
```
elihart, line 105: @gpeal should we maybe set `state` before sending on the channel, so that any channel subscribers can then query `state` and have it match?

mvrx/src/main/kotlin/com/airbnb/mvrx/CoroutinesStateStore.kt
```diff
@@ -0,0 +1,128 @@
...
+     *
+     * Sitaution 2
+     *
+     * withState {
+     *     setState { ... }
+     *     withState { ... }
+     * }
+     */
+    private suspend fun flushQueuesOnce() {
+        select<Unit> {
```
elihart, line 101: I like this, it's what I had in mind when I suggested something like this a while back

mvrx/src/main/kotlin/com/airbnb/mvrx/CoroutinesStateStore.kt
gpeal: I did it this way to avoid ever getting the same state twice on lines 52 and 53

buildSrc/build.gradle.kts
```diff
@@ -0,0 +1,18 @@
...
+}
+
+plugins {
+    java
+    `kotlin-dsl`
+}
+
+dependencies {
+    implementation(gradleApi())
+    implementation("com.android.tools.build:gradle:3.6.3")
```
jpventura, line 15: implementation Libraries.gradlePlugin
    implementation Libraries.jacoco
    implementation Libraries.kotlin

Strange... it should be an entry to `Libraries.gradlePlugin` on `dependencies.gradle` :thinking:

mvrx/src/main/kotlin/com/airbnb/mvrx/MvRxLifecycleAwareFlow.kt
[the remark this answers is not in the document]
jpventura: @gpeal Why not using [`owner.whenStarted()`](https://developer.android.com/reference/kotlin/androidx/lifecycle/package-summary#whenstarted_1)?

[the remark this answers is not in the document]
gpeal: @jpventura Because that launches a new coroutine every time it is started.

mvrx/src/main/kotlin/com/airbnb/mvrx/ScriptableMvRxStateStore.kt
```diff
@@ -1,22 +1,29 @@
...
+@Suppress("EXPERIMENTAL_API_USAGE")
 class ScriptableMvRxStateStore<S : Any>(initialState: S) : MvRxStateStore<S> {
 
-    private val subject: BehaviorSubject<S> = BehaviorSubject.createDefault(initialState)
-
-    override val observable: Observable<S> = subject.distinctUntilChanged()
-
-    override val state: S
-        get() = subject.value!!
+    private val stateChannel = BroadcastChannel<S>(capacity = Channel.BUFFERED)
```
jpventura, line 20: @gpeal I think you can achieve the [`BehaviorSubject`](http://reactivex.io/RxJava/javadoc/io/reactivex/subjects/BehaviorSubject.html) behavior using [`ConflatedBroadcastChannel`](https://stackoverflow.com/questions/57345311/how-can-i-send-items-to-a-kotlin-flow-like-a-behaviorsubject).

The [`BroadcastChannel`](https://stackoverflow.com/questions/57345311/how-can-i-send-items-to-a-kotlin-flow-like-a-behaviorsubject) was used to achieve the [`PublishSubject`](http://reactivex.io/RxJava/javadoc/io/reactivex/subjects/PublishSubject.html).

gpeal: @jpventura `ConflatedBroadcastChannel` conflates inputs. It's actually possible to set the state faster than it can be delivered to consumers, even with a buffer. I was able to demonstrate that in some of the added tests. Tricky stuff ;)

mvrx/src/main/kotlin/com/airbnb/mvrx/CoroutinesStateStore.kt
```diff
@@ -0,0 +1,128 @@
...
+            }
+        }
+    }
+
+    /**
+     * Flush the setState and withState queues.
+     * All pending setState reducers will be run prior to every single withState lambda.
+     * This ensures that situations such as the following will work correctly:
+     *
+     * Sitaution 1
```
jpventura, line 88: `s/Sitaution/Situation/g`

mvrx/src/main/kotlin/com/airbnb/mvrx/BaseMvRxViewModel.kt
```diff
@@ -51,23 +79,35 @@ abstract class BaseMvRxViewModel<S : MvRxState>(
...
-                .subscribeOn(Schedulers.computation()).subscribe()
+            viewModelScope.launch(Dispatchers.Default) {
+                validateState(initialState)
+            }
         }
     }
 
     @CallSuper
     override fun onCleared() {
-        super.onCleared()
```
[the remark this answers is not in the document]
gpeal: This was actually not intentional and probably happened with a bad merge from that branch. Good catch!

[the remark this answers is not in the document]
gpeal: Fixing in the BaseFragment PR

mvrx/src/main/kotlin/com/airbnb/mvrx/BaseMvRxViewModel.kt
```diff
@@ -28,9 +42,23 @@ import kotlin.reflect.KProperty1
...
+     * Provide a default context for viewModelScope. It will be added after [SupervisorJob]
+     * and [Dispatchers.Main.immediate]. This should only be used for tests and should only
+     * be exposed via a shared base class within your app. If your features extend this
+     * directly, do not override this in the primary constructor of your feature ViewModel.
+     */
+    contextOverride: CoroutineContext? = null
 ) : ViewModel() {
     private val debugMode = if (MvRxTestOverrides.FORCE_DEBUG == null) debugMode else MvRxTestOverrides.FORCE_DEBUG
+    protected val viewModelScope = CoroutineScope(SupervisorJob() + Dispatchers.Main.immediate + contextOverride)
+    private val stateStore = stateStoreOverride ?: CoroutinesStateStore(initialState, viewModelScope)
```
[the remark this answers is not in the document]
gpeal: @sanogueralorenzo Yup, that's exactly the reason :)

Issue #366, opened 2020-03-26
Update package dependencies to latest stable version


jpventura:
Latest stable versions of dependencies:
    
  - adapter-rxjava2 v2.7.2
  - assisted-inject-annotations-dagger2 v0.5.2
  - assisted-inject-processor-dagger2 v0.5.2
  - converter-moshi v2.7.2
  - core-ktx v1.2.0
  - dagger v2.27
  - dagger-compiler v2.27
  - epoxy v3.9.0
  - fragment-ktx v1.2.3
  - junit v4.13
  - kotlin-stdlib-jdk8 v1.3.50
  - lifecycle-extensions v2.2.0
  - mockito-kotlin v2.2.0
  - moshi v1.9.2
  - moshi-kotlin v1.9.2
  - moshi-kotlin-codegen v1.8.2
  - navigation-fragment-ktx v2.2.1
  - navigation-ui-ktx v2.2.1
  - picasso v2.71828
  - recyclerview v1.1.0
  - retrofit v2.7.2
  - robolectric v4.3
  - room-compiler v2.2.0
  - room-runtime v2.2.0
  - room-rxjava2 v2.2.0
  - rxjava v2.2.9

Dugin123:
is supporting RxJava3 not yet?

elihart:
We have not started work to support rxjava3, but would like to in the future

jpventura:
@elihart the problem is if/when JetBrains will support officially `kotlinx-coroutines-rx3` as they already do for [`kotlinx-coroutines-rx2`](https://github.com/Kotlin/kotlinx.coroutines/tree/master/reactive/kotlinx-coroutines-rx2).

Otherwise, instead of building it over coroutines (PR #347), it will be required maintain two distinct code bases.

Issue #363, opened 2020-03-26
Centralize Gradle dependencies file


jpventura:
About
--------
As presented during [Google I/O 2019][google-io], it is recommended to create extra configurations for versioning dependencies, improving visibility and simplifying libraries versioning across multiple project modules.

References
---------------

  - [Android Jetpack Versions][jetpack-versions]
  - [Caster.IO: Android Clean Architecture][caster-io]
  - [Google Developers Experts: 19 tips for Gradle in Android projects][gde]
  - [Google I/O 2019 - Build Bigger, Better: Gradle for Large Projects (Google I/O'19)][google-io]
  - [Google Samples - Architecture Blueprints v2][blue-prints]

[jetpack-versions]: https://developer.android.com/jetpack/androidx/versions
[blue-prints]: https://github.com/android/architecture-samples
[caster-io]: https://caster.io/courses/android-clean-architecture
[gde]: https://medium.com/google-developer-experts/19-tips-for-gradle-in-android-projects-2019-edition-11af704eb06e
[github]: https://github.com/hitherejoe/GithubTrending/blob/master/dependencies.gradle
[google-io]: https://www.youtube.com/watch?v=sQC9-Rj2yLI

jpventura:
Fixed by #364.

PR #364, opened 2020-03-26
build(all): centralized dependencies file

jpventura:
About
--------
Create [`dependencies.gradle`][github] file visible by all modules, centralizing libraries versioning, configuration and packages updates.

Build tools version was also updated as follows:

  - Build tools v29.0.3
  - Compile SDK v29
  - Minimum API level v16
  - Target API level v29 (Android Q)
  - Libraries were _ceiled_ to highest version across project modules.

In order to ensure that no behavior was broken due the refactor, libraries updates were moved to pull request #367. Android API was [forced to v28 on Roboeletric tests][roboeletric].

Blocks
---------
  - #366 Update package dependencies to latest stable version

Fixes
-------

  - #363 Centralize Gradle dependencies file

References
---------------

  - [Android Jetpack Versions][jetpack-versions]
  - [Caster.IO: Android Clean Architecture][caster-io]
  - [GitHub: Android Clean Architecture][github]
  - [Google Android - Java 8 Support][google-android]
  - [Google Developers Experts: 19 tips for Gradle in Android projects][gde]
  - [Google I/O 2019 - Build Bigger, Better: Gradle for Large Projects (Google I/O'19)][google-io]
  - [Google Samples - Architecture Blueprints v2][blue-prints]
  - [Stack Overflow - Does Roboeletric require Java 9?][roboeletric]

[jetpack-versions]: https://developer.android.com/jetpack/androidx/versions
[blue-prints]: https://github.com/android/architecture-samples
[caster-io]: https://caster.io/courses/android-clean-architecture
[gde]: https://medium.com/google-developer-experts/19-tips-for-gradle-in-android-projects-2019-edition-11af704eb06e
[github]: https://github.com/hitherejoe/GithubTrending/blob/master/dependencies.gradle
[google-android]: https://developer.android.com/studio/write/java8-support
[google-io]: https://www.youtube.com/watch?v=sQC9-Rj2yLI
[roboeletric]: https://stackoverflow.com/questions/56821193/does-robolectric-require-java-9

PR #367, opened 2020-03-26
chore(all): update dependencies

jpventura:
About
--------

Following libraries were upgraded:
    
  - adapter-rxjava2 v2.7.2
  - assisted-inject-annotations-dagger2 v0.5.2
  - assisted-inject-processor-dagger2 v0.5.2
  - converter-moshi v2.7.2
  - core-ktx v1.2.0
  - dagger v2.27
  - dagger-compiler v2.27
  - epoxy v3.9.0
  - fragment-ktx v1.2.3
  - junit v4.13
  - kotlin-stdlib-jdk8 v1.3.50
  - lifecycle-extensions v2.2.0
  - mockito-kotlin v2.2.0
  - moshi v1.9.2
  - moshi-kotlin v1.9.2
  - moshi-kotlin-codegen v1.8.2
  - navigation-fragment-ktx v2.2.1
  - navigation-ui-ktx v2.2.1
  - picasso v2.71828
  - recyclerview v1.1.0
  - retrofit v2.7.2
  - robolectric v4.3
  - room-compiler v2.2.0
  - room-runtime v2.2.0
  - room-rxjava2 v2.2.0
  - rxjava v2.2.9

Kotlin was kept at v1.3.61 to keep compatibility with `kotlin-gradle-plugin`.

Fixes
-------
  - #366 Update package dependencies to latest stable version

Requires
------------
  - #363 Centralize Gradle dependencies file

References
---------------

  - [Google Android - Java 8 Support][google-android]

[google-android]: https://developer.android.com/studio/write/java8-support


hellodagger/build.gradle
```diff
@@ -1,24 +1,25 @@
-apply plugin: "com.android.application"
-apply plugin: "kotlin-android"
-apply plugin: "kotlin-android-extensions"
-apply plugin: "kotlin-kapt"
+apply plugin: 'com.android.application'
```
elihart: I personally prefer double quotes for all strings. the look is consistent and if we ever want to insert arguments we don't have to change from single to double quotes

mvrx/build.gradle
```diff
@@ -2,19 +2,20 @@ apply plugin: 'com.android.library'
...
 
     defaultConfig {
-        minSdkVersion 16
-        targetSdkVersion 28
+        minSdkVersion versions.minSdk
+        targetSdkVersion versions.targetSdk
         consumerProguardFiles 'proguard-rules.pro'
     }
-    buildToolsVersion "$buildToolsVersion"
+
```
elihart: do we need
```
  kotlinOptions {
    jvmTarget = "1.8"
  }
```

in here too, since we are targeting 1.8 in compileOptions?

hellodagger/build.gradle
jpventura: I have no personal taste about it, I've just modified in order to keep string references consistent across Gradle files.

You have Minerva's vote on this :wink:

mvrx/build.gradle
jpventura: As far as I could understand from [Google Android documentation](https://developer.android.com/studio/write/java8-support), this is required bye Kotlin compiler:
```Gradle
android {

  // Configure only for each module that uses Java 8
  // language features (either in its source code or
  // through dependencies).

  compileOptions {
    sourceCompatibility JavaVersion.VERSION_1_8
    targetCompatibility JavaVersion.VERSION_1_8
  }

  // For Kotlin projects
  kotlinOptions {
    jvmTarget = '1.8'
  }

}
```

jpventura: @elihart Found it! [Cannot inline bytecode built with Java 8 into code built with targed Java 6](https://stackoverflow.com/questions/48988778/cannot-inline-bytecode-built-with-jvm-target-1-8-into-bytecode-that-is-being-bui/56996020#56996020)

settings.gradle
```diff
@@ -1,7 +1,7 @@
+include ':counter'
```
tasomaniac: Since I saw categorization in other `gradle` files, I will recommend categorizing samples vs library modules here.

jpventura: @tasomaniac [Great idea!](https://jeroenmols.com/blog/2019/04/02/modularizationexample/) 🎉 

```Groovy
{
    include ':libraries:mvrx'
    include ':libraries:testing'

    include ':samples:counter'
    include ':samples:dogs'
    include ':samples:hellodagger'
    include ':samples:sample'
    include ':samples:todomvrx'
}
```

But I rather do this in a sole PR only for it because lots of files would be just renamed and make a lot of noise at the review. Also it would require [`git log --follow`](https://stackoverflow.com/questions/4393527/why-might-git-log-not-show-history-for-a-moved-file-and-what-can-i-do-about-it) to track code changes.

Does it make sense @elihart @tasomaniac?

tasomaniac: What I meant was to simply put header comments like `// Samples` and separate them. 😊

hellodagger/build.gradle
sanogueralorenzo: 👋 I would keep the double quotes, so you don't need to go from single to double again when you migrate to `.kts` files 😄

jpventura: @sanogueralorenzo agreed, pushed to #364, and rebase #367 over it

settings.gradle
jpventura: @tasomaniac agreed, resolved at #364, and rebased #367 over it.

jpventura:
@elihart rebased and ready to merge 🎉

dependencies.gradle
```diff
@@ -3,46 +3,48 @@ ext {
             // Build tools and SDK
             buildTools      : "29.0.3",
             compileSdk      : 29,
+            gradle          : "3.6.1",
```
elihart, line 6: I think a better name would be "agp" or something more explicit. "gradle" isn't android specific and is a bit vague

dependencies.gradle
```diff
@@ -3,46 +3,48 @@ ext {
...
             jetpack         : "1.1.0",
-            kotlin          : "1.3.50",
+            kotlin          : "1.3.61",
```
elihart, line 8: why not 71?

jpventura: The was a compilation warning because Gradle plugin v3.6.1 was compiled with Kotlin v1.3.61.

dependencies.gradle
jpventura: @elihart both `agp` and `gradlePlugin` would be more descriptive.

jpventura:
@elihart `.idea` modification was re-enabled at e66c2996 in order to ensure that both XML and Kotlin style were correctly being applied by the IDE.

jpventura:
@elihart @gpeal what you consider more imperative to be solved next: AAC Navigation or Coroutines migration?

elihart:
@jpventura I think coroutines migration would be a good thing to tackle first!

PR #351, opened 2020-02-15
Add Jetpack navigation module

marukami:
* Navigation 2.2.0 requires Java8. So, everything was updated to Java8
* Add a `navGraphViewModel` for navigation graph view models
* `navigationLifecycleAwareLazy` a slightly forked `lifecycleAwareLazy` that subscribes the MvRx view when possible and if not defers the lazy until the fragment access the view model
* `MvRxNavDirections` a NavDirections that makes it a little easier to pass an argument Parcelable and set the navigation label if provided
* Added very basic example navigation repo.
* Added Test Fragment scenario for navigation module. 

**NOTES**
* The mvrx view needs to wait for until at least the `onViewCreated` callback. During lifecycle flows like re-configuration the `NavHostFragment` is not accessible until after the activity onCreate has finished.
* Oops accidentally deleted branch for #337 and it closed the PR 😅

PR #372, opened 2020-04-04
build: lifecycle-common instead of lifecycle-extensions

jpventura:


jpventura:
Modifications were cherry-picked from @marukami #351 pull request.

PR #373, opened 2020-04-07
use plugin to remove build.gradle duplication

marukami:
* Groovy and Kotlin `buildSrc` don't play nicely so I moved the `dependencies` into buildSrc and made it a kt file.
* Any globally shared `build.gradle` config was moved into `AndroidMetaModulePlugin`
* #351 needs to share the Jacoco coverage task, so, a custom task was made `JacocoReportTask`


buildSrc/build.gradle.kts
```diff
@@ -0,0 +1,19 @@
...
+}
+
+plugins {
+    java
+    `kotlin-dsl`
+}
+
+dependencies {
+    implementation(gradleApi())
+    // Gradle version can be a lower minior version compared to the main lib/app Gradle project
```
elihart, line 15:
```suggestion
    // Gradle version can be a lower minor version compared to the main lib/app Gradle project
```

elihart: It would be nice to keep the gradle and kotlin plugin versions the same as the rest of the app just for ease of updating

marukami: I was able to share the values using a properties file https://github.com/airbnb/MvRx/blob/5969f3c8425229e8a607eab28acf0539c6d1626d/versions.properties#L1-L2

The only drawback with this method is a little duplication with loading the file and a few @JvmStatic so Groovy can find the function.

buildSrc/build.gradle.kts
```diff
@@ -0,0 +1,24 @@
...
+
+val properties = Properties().apply {
+    load(rootProject.file("../versions.properties").inputStream())
+}
+
+dependencies {
+    implementation(gradleApi())
+    implementation("com.android.tools.build:gradle:${properties["gradlePlugin"] as String}")
+    implementation("org.jacoco:org.jacoco.core:0.8.5")
+    implementation("org.jetbrains.kotlin:kotlin-gradle-plugin:1.3.71")
```
elihart, line 23: use kotlin version here too?

versions.properties
```diff
@@ -0,0 +1,2 @@
+gradlePlugin=3.6.2
```
elihart, line 1: is this so that the buildsrc build.gradle can access these as well? i feel like the amount of complexity (and overhead from continually loading this file) added for that isn't worth it and we should just take the small messiness hit to define these versions in two places instead

marukami: Yep, totally agree. The buildSrc only need to keep up with the major Android Gradle plugin version; even then unless the Tasks or Plugins are using removed APIs it will keep working, the buildSrc can drift quite a bit before it even starts to become an issue at all.

buildSrc/build.gradle.kts
marukami: The same problem as the AGP with access scope is not been shared with any easy. The Kotlin version that Gradle used for the `.gradle.kts` also trails by quite a bit, so, not always having this version in sync should be fine.

PR #382, opened 2020-05-08
Added a DelegateMvRxView 

tunjid:
A simple Delegate that uses the saved state artifact, for cases where inheriting from `BaseMvRxFragment` is unfeasible, undesirable, or for composing with a custom view. The custom view may choose to grab the `SavedStateRegistryOwner` from either an `Activity` or `Fragment`.

I was able to test this with DogsFragment fine, even across process death. I'm not sure how to contribute this change with tests for this project however, pointers are much appreciated.

There's also the function delegates that will need to be added for `activityViewModel` and the like should this initial change be worthwhile.

Usage was as follows:

```kotlin
class DogsFragment : Fragment(), DogsFragmentHandler {

    private val view = DelegateMvRxView(this, ::invalidate)
    private val viewModel: DogViewModel by activityViewModel(view)
    private lateinit var bindings: DogsFragmentBinding
    private val adapter = DogAdapter(this)

    ...

    private fun invalidate() = withState(viewModel) { state ->
        bindings.state = state
    }
}
```

### DIFF
--- a/mvrx/src/main/kotlin/com/airbnb/mvrx/BaseMvRxFragment.kt
+++ b/mvrx/src/main/kotlin/com/airbnb/mvrx/BaseMvRxFragment.kt
@@ -12,30 +12,21 @@ import androidx.lifecycle.LifecycleOwner
  */
 abstract class BaseMvRxFragment(@LayoutRes contentLayoutId: Int = 0) : Fragment(contentLayoutId), MvRxView {
 
-    private val mvrxViewIdProperty = MvRxViewId()
-    final override val mvrxViewId: String by mvrxViewIdProperty
+    private lateinit var viewDelegate: DelegateMvRxView
+
+    final override val mvrxViewId: String get() = viewDelegate.mvrxViewId
 
     override fun onCreate(savedInstanceState: Bundle?) {
-        mvrxViewIdProperty.restoreFrom(savedInstanceState)
         super.onCreate(savedInstanceState)
+        viewDelegate = DelegateMvRxView(this, ::invalidate)
     }
 
-    /**
-     * Fragments should override the subscriptionLifecycle owner so that subscriptions made after onCreate
-     * are properly disposed as fragments are moved from/to the backstack.
-     */
     override val subscriptionLifecycleOwner: LifecycleOwner
         get() = this.viewLifecycleOwnerLiveData.value ?: this
 
-    override fun onSaveInstanceState(outState: Bundle) {
-        super.onSaveInstanceState(outState)
-        mvrxViewIdProperty.saveTo(outState)
-    }
+    override fun uniqueOnly(customId: String?): UniqueOnly = viewDelegate.uniqueOnly(customId)
 
-    override fun onStart() {
-        super.onStart()
-        // This ensures that invalidate() is called for static screens that don't
-        // subscribe to a ViewModel.
-        postInvalidate()
-    }
+    override fun invalidate() = Unit
+
+    override fun postInvalidate() = viewDelegate.postInvalidate()
 }
--- a/mvrx/src/main/kotlin/com/airbnb/mvrx/BaseMvRxFragment.kt
+++ b/mvrx/src/main/kotlin/com/airbnb/mvrx/BaseMvRxFragment.kt
@@ -18,7 +18,11 @@ abstract class BaseMvRxFragment(@LayoutRes contentLayoutId: Int = 0) : Fragment(
 
     override fun onCreate(savedInstanceState: Bundle?) {
         super.onCreate(savedInstanceState)
-        viewDelegate = DelegateMvRxView(this, ::invalidate)
+        viewDelegate = DelegateMvRxView(
+                savedStateRegistryOwner = this,
+                subscriptionLifecycleOwner = this,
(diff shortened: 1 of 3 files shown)


mvrx/src/main/kotlin/com/airbnb/mvrx/DelegateMvRxView.kt
```diff
@@ -0,0 +1,39 @@
+package com.airbnb.mvrx
+
+import android.os.Bundle
+import androidx.lifecycle.Lifecycle
+import androidx.lifecycle.LifecycleEventObserver
+import androidx.lifecycle.LifecycleOwner
+import androidx.savedstate.SavedStateRegistryOwner
+
+class DelegateMvRxView(
+        private val owner: SavedStateRegistryOwner,
```
tunjid, line 10: This can be either an `Activity` or `Fragment` Should Google. The type could be more open to allow. for a different method of `mvrxViewId` saving and restoration.

elihart: this looks pretty reasonable to me. can you add some kdoc to this class to describe the purpose and when it may be used?

tunjid:
I added a `by activityViewModel(view)` when I was testing, but I removed it from the PR as it was a copy paste of the existing `activityViewModel` extension on a `Fragment`; I wanted to make sure the PR had legs before adding more methods and generally present a small diff at first.

I'll go ahead and use the Delegate in the `BaseMvRxFragment`.

mvrx/src/main/kotlin/com/airbnb/mvrx/BaseMvRxFragment.kt
```diff
@@ -12,30 +12,25 @@ import androidx.lifecycle.LifecycleOwner
...
-    }
+    override fun uniqueOnly(customId: String?): UniqueOnly = viewDelegate.uniqueOnly(customId)
 
-    override fun onStart() {
-        super.onStart()
-        // This ensures that invalidate() is called for static screens that don't
-        // subscribe to a ViewModel.
-        postInvalidate()
-    }
+    override fun invalidate() = Unit
```
elihart, line 33: shouldn't we leave this unimplemented to force base classes to implement?

mvrx/src/main/kotlin/com/airbnb/mvrx/BaseMvRxFragment.kt
```diff
@@ -12,30 +12,25 @@ import androidx.lifecycle.LifecycleOwner
...
-    private val mvrxViewIdProperty = MvRxViewId()
-    final override val mvrxViewId: String by mvrxViewIdProperty
+    private lateinit var viewDelegate: DelegateMvRxView
+
+    final override val mvrxViewId: String get() = viewDelegate.mvrxViewId
 
     override fun onCreate(savedInstanceState: Bundle?) {
-        mvrxViewIdProperty.restoreFrom(savedInstanceState)
         super.onCreate(savedInstanceState)
+        viewDelegate = DelegateMvRxView(
```
elihart, line 21: might be nice to provide a fragment specific implementation of this to make it easier to drop in for the common case

tunjid: Sounds good, will add an extension for it.

mvrx/src/main/kotlin/com/airbnb/mvrx/BaseMvRxFragment.kt
```diff
@@ -12,30 +12,22 @@ import androidx.lifecycle.LifecycleOwner
...
-    final override val mvrxViewId: String by mvrxViewIdProperty
+    private lateinit var viewDelegate: DelegateMvRxView
+
+    final override val mvrxViewId: String get() = viewDelegate.mvrxViewId
 
     override fun onCreate(savedInstanceState: Bundle?) {
-        mvrxViewIdProperty.restoreFrom(savedInstanceState)
         super.onCreate(savedInstanceState)
+        viewDelegate = delegateMvRxView(
+                subscriptionLifecycleOwner = subscriptionLifecycleOwner,
```
elihart, line 22: I think `subscriptionLifecycleOwner` is not necessary to explicitly define now, as the default is the same?

mvrx/src/main/kotlin/com/airbnb/mvrx/DelegateMvRxView.kt
```diff
@@ -0,0 +1,66 @@
...
+
+    override fun invalidate() = onInvalidated()
+
+    override fun getLifecycle(): Lifecycle = savedStateRegistryOwner.lifecycle
+}
+
+/**
+ * Convenience for providing a [DelegateMvRxView] to a [Fragment]
+ */
+fun Fragment.delegateMvRxView(
```
elihart, line 58: nice :)

gpeal:
@tunjid This should no longer be necessary as of https://github.com/airbnb/MvRx/pull/396 :)

Thanks for suggesting we take another look at this! It requried some new APIs that didn't exist when MvRx was first built (ViewModel SavedStateHandle).

PR #383, opened 2020-05-09
Add value to all Async (load more data use case) 

eboudrant:
For "load more data" we currently can't have a `Loading` or `Fail` with some existing data to render. The workaround to make it works is to keep the data separated from the `Async` property (see https://gist.github.com/gpeal/145202421a4e6a1e79c6afeb58b50134).

`Async` already have a `invoke()` method so we can easily solve the issue by adding a `value` in `Loading` and `Fail` classes. 

Example :

Initial load : 
```
setState { 
  copy(
    data = Loading()
  ) 
}
```
Load more page : 
```
setState {
  val currentData = this.data.invoke()
  copy(
    data = Loading(currentData)
  ) 
}
```

### DIFF
--- a/mvrx/src/main/kotlin/com/airbnb/mvrx/Async.kt
+++ b/mvrx/src/main/kotlin/com/airbnb/mvrx/Async.kt
@@ -38,13 +38,9 @@ sealed class Async<out T>(val complete: Boolean, val shouldLoad: Boolean, privat
 
 object Uninitialized : Async<Nothing>(complete = false, shouldLoad = true, value = null), Incomplete
 
-class Loading<out T>(value: T? = null) : Async<T>(complete = false, shouldLoad = false, value = value), Incomplete {
-    override fun equals(other: Any?) = other is Loading<*>
+data class Loading<out T>(private val value: T? = null) : Async<T>(complete = false, shouldLoad = false, value = value), Incomplete
 
-    override fun hashCode() = "Loading".hashCode()
-}
-
-class Success<out T>(value: T) : Async<T>(complete = true, shouldLoad = false, value = value) {
+data class Success<out T>(private val value: T) : Async<T>(complete = true, shouldLoad = false, value = value) {
 
     override fun invoke(): T = super.invoke() as T
 
--- a/mvrx/src/main/kotlin/com/airbnb/mvrx/Async.kt
+++ b/mvrx/src/main/kotlin/com/airbnb/mvrx/Async.kt
@@ -9,14 +9,14 @@ import java.util.Arrays
  * Complete: Success, Fail
  * ShouldLoad: Uninitialized, Fail
  */
-sealed class Async<out T>(val complete: Boolean, val shouldLoad: Boolean) {
+sealed class Async<out T>(val complete: Boolean, val shouldLoad: Boolean, private val invokeValue: T?) {
 
     /**
      * Returns the Success value or null.
      *
      * Can be invoked as an operator like: `yourProp()`
      */
-    open operator fun invoke(): T? = null
+    open operator fun invoke(): T? = invokeValue
 
     companion object {
         /**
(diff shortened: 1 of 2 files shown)


mvrx/src/main/kotlin/com/airbnb/mvrx/Async.kt
```diff
@@ -36,17 +36,13 @@ sealed class Async<out T>(val complete: Boolean, val shouldLoad: Boolean) {
...
+data class Loading<out T>(private val value: T? = null) : Async<T>(complete = false, shouldLoad = false, value = value), Incomplete
 
-    override fun hashCode() = "Loading".hashCode()
-}
-
-data class Success<out T>(private val value: T) : Async<T>(complete = true, shouldLoad = false) {
+data class Success<out T>(private val value: T) : Async<T>(complete = true, shouldLoad = false, value = value) {
 
-    override operator fun invoke(): T = value
+    override fun invoke(): T = super.invoke() as T
```
gpeal, line 45: You can leave this as is. Returning value is still correct here

mvrx/src/main/kotlin/com/airbnb/mvrx/Async.kt
```diff
@@ -36,17 +36,13 @@ sealed class Async<out T>(val complete: Boolean, val shouldLoad: Boolean) {
     }
 }
 
-object Uninitialized : Async<Nothing>(complete = false, shouldLoad = true), Incomplete
+object Uninitialized : Async<Nothing>(complete = false, shouldLoad = true, value = null), Incomplete
 
-class Loading<out T> : Async<T>(complete = false, shouldLoad = false), Incomplete {
-    override fun equals(other: Any?) = other is Loading<*>
+data class Loading<out T>(private val value: T? = null) : Async<T>(complete = false, shouldLoad = false, value = value), Incomplete
```
gpeal, line 41: Can you update the javadoc to include information as to why you would use value?

mvrx/src/main/kotlin/com/airbnb/mvrx/Async.kt
eboudrant: Sure, it was a result of my testing of making `Success` not a data class which did not worked.

mvrx/src/main/kotlin/com/airbnb/mvrx/Async.kt
eboudrant: Sure, I updated the doc from `Async.invoke()` ... do you think it is enough?

elihart:
@gpeal your property reference thing looks like a prototype I've been sitting on for a while.  I wanted something to auto cancel requests made for the same property

```kotlin
// In BaseViewModel
 private val propertyToDisposable = mutableMapOf<KProperty1<S, *>, Disposable>()

    fun <T> Observable<T>.execute(
        asyncProperty: KProperty1<S, Async<T>>,
        reducer: S.(Async<T>) -> S = { updateProperty(asyncProperty, it) }
    ): Disposable {
        cancelExecution(asyncProperty)
        return execute(reducer).saveTo(asyncProperty)
    }

    private fun <T> S.updateProperty(stateProperty: KProperty1<S, T>, valueToSet: T): S {
        return callCopy(stateProperty.name to valueToSet)
    }

    protected fun <T> cancelExecution(stateProperty: KProperty1<S, T>, newValue: T? = null) {
        propertyToDisposable[stateProperty]?.dispose()
        propertyToJob[stateProperty]?.cancel()

        newValue?.let { valueToSet ->
            setState {
                updateProperty(stateProperty, valueToSet)
            }
        }
    }

    @JvmName("cancelExecutionAsyncProperty")
    protected fun <T> cancelExecution(stateProperty: KProperty1<S, Async<T>>, newValue: Async<T>? = Uninitialized) {
        // TODO (eli_hart 2019-08-09): what is a good default for newValue?
        cancelExecution(stateProperty, newValue as Any?)
    }

    fun <K : Any, T : Any> Observable<T>.execute(disposableMap: DisposableMap<T>, key: T, stateReducer: S.(Async<T>) -> S) {
        execute(stateReducer).let { disposable ->
            disposableMap.put(key, disposable)
        }
    }
```

lwj1994:
I had a bad experience earlier that when refresh data, the Loading value will be null , so the listpage will have a blank flash.

so i solve it by :  ` if (request is Loading && feeds.isNotEmpty()) return copy(feeds = feeds)
`，

```kotlin
data class MyState(
  val feeds: List<Int> = emptyList(),
  val cursor: String? = null,
  val request: Async<Pair<List<Int>, String?>> = Uninitialized) {

   fun generateNewFetchState(request: Async<Pair<List<FlowFeed>, String?>>, clear: Boolean): MyState {
     if (request is Loading && feeds.isNotEmpty()) return copy(feeds = feeds)
     // ...
   }

}
```

gpeal:
@lwj1994 Did this PR not fix it for you? In the current version of MvRx, you have to make sure you reducer copies over the value. In a future version of MvRx, we will enable it to copy over automatically.

lwj1994:
@gpeal yes it fixed ，I mean before this PR I used the method above to fix it.
Looking forward to the automatic version.

PR #385, opened 2020-05-11
Make BaseMvRxViewModel not extend ViewModel

gpeal:
It would be amazing if we can get BaseMvRxViewModel to be a pojo. That would enable testing on the jvm without robolectric and a potential path to Kotlin Multiplatform.

I was just poking around at what this would look like and it turned out to be fairly simple.

The next hurdle is replacing the inter-vm subscriptions which use Lifecycle with something else but that should be doable.

This PR builds on https://github.com/airbnb/MvRx/pull/347

@ZacSweers

### DIFF
--- a/mvrx/src/main/kotlin/com/airbnb/mvrx/BaseMvRxViewModel.kt
+++ b/mvrx/src/main/kotlin/com/airbnb/mvrx/BaseMvRxViewModel.kt
@@ -42,9 +42,8 @@ abstract class BaseMvRxViewModel<S : MvRxState>(
     stateStoreOveride: MvRxStateStore<S>? = null
 ) {
     private val debugMode = if (MvRxTestOverrides.FORCE_DEBUG == null) debugMode else MvRxTestOverrides.FORCE_DEBUG
-    @VisibleForTesting
-    internal val scope: CoroutineScope = CoroutineScope(SupervisorJob() + Dispatchers.Main.immediate)
-    private val stateStore = stateStoreOveride ?: CoroutinesStateStore(initialState, scope)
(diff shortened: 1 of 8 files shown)


tasomaniac:
What is holding unit testing on jvm? We don't use Robolectric and unit test both BasvRxViewModel and MvRxView layer.

mvrx/src/main/kotlin/com/airbnb/mvrx/BaseMvRxViewModel.kt
```diff
@@ -86,15 +86,13 @@ abstract class BaseMvRxViewModel<S : MvRxState>(
...
+            scope.launch(Dispatchers.Default) {
                 validateState(initialState)
             }
         }
     }
 
-    @CallSuper
-    override fun onCleared() {
-        super.onCleared()
+    fun onCleared() {
```
denis-bezrukov, line 95: you need to cancel `scope` here

gpeal: Doh! Thanks!

mvrx/src/main/kotlin/com/airbnb/mvrx/CoroutinesStateStore.kt
```diff
@@ -76,12 +77,12 @@ class CoroutinesStateStore<S : MvRxState>(
      * }
      */
     @Synchronized
```
denis-bezrukov, line 79: `@Synchronized` on suspend fun will not work properly, plus since it's necessary only for sync case, probably would be better to move it to only blocking call

gpeal:
Superceeded by https://github.com/airbnb/MvRx/pull/394

PR #386, opened 2020-05-11
Use shared dispatcher for flushing queues instead of single thread per VM

denis-bezrukov:
This PR replaces single thread dispatcher per VM with shared thread pool. 
It will reuse existing threads if possible or will spawn one more thread if necessary. As a result, less thread will be used, and threads from cleared VMs will be reused. 

`newCachedThreadPool` docs: 

>Creates a thread pool that creates new threads as needed, but
     will reuse previously constructed threads when they are
     available.  **These pools will typically improve the performance
     of programs that execute many short-lived asynchronous tasks.**
     Calls to {@code execute} will reuse previously constructed
     threads if available. If no existing thread is available, a new
     thread will be created and added to the pool. Threads that have
     not been used for sixty seconds are terminated and removed from
     the cache. Thus, a pool that remains idle for long enough will
     not consume any resources.

I could add some benchmarks for the different scenarios. Is there a common pattern/example in this repo for such benchmarks?

Also cached thread pool can be configured with non-default params:
minThreadCount (currently 0)
maxThreadCount (currently inf)
keepAliveTime (currently 60s)

based on #347

### DIFF
--- a/mvrx/src/main/kotlin/com/airbnb/mvrx/CoroutinesStateStore.kt
+++ b/mvrx/src/main/kotlin/com/airbnb/mvrx/CoroutinesStateStore.kt
@@ -68,12 +68,7 @@ class CoroutinesStateStore<S : MvRxState>(
     private fun setupTriggerFlushQueues(scope: CoroutineScope) {
         if (MvRxTestOverrides.FORCE_SYNCHRONOUS_STATE_STORES) return
 
-        val executor = Executors.newSingleThreadExecutor()
-        scope.coroutineContext[Job]!!.invokeOnCompletion {
-            executor.shutdownNow()
-        }
-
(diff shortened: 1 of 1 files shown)


denis-bezrukov:
@BenSchwab added this PR after reading @elihart's comment https://github.com/airbnb/MvRx/pull/347#discussion_r379120088 . I thought what can be improved here without breaking the current logic @gpeal mentioned [here](https://github.com/airbnb/MvRx/pull/347#discussion_r379193205). 
Sharing threadpool between viewmodels should allow to use less threads than before. Less threads means less memory consumption, less thread context switches, less thread parks/unparks. It also means threads will be reused, so after navigating from one screen to another typically won't produce new threads. 

in our project one screen may contain 30+ active VMs (and some more inactive in fragment backstack), which means previously we had 30+ threads for them.

PR #387, opened 2020-05-12
Allow for custom factories for VM Store

marukami:
Pulled out of #351 since it should allow library users to manually add Navigation support.

### DIFF
--- a/mvrx/src/main/kotlin/com/airbnb/mvrx/MvRxViewModelFactory.kt
+++ b/mvrx/src/main/kotlin/com/airbnb/mvrx/MvRxViewModelFactory.kt
@@ -94,6 +94,10 @@ data class ActivityViewModelContext(
 /**
  * The [ViewModelContext] for a ViewModel created with a
  * fragment scope (`val viewModel by fragmentViewModel<MyViewModel>`).
+ *
+ * The [owner] and [savedStateRegistry] default to the Fragment as the provider, if you need to use an alternative
+ * source for the [ViewModelStoreOwner] and/or [SavedStateRegistry]; such in the case of JetPack Navigation Component destinations with ViewModels,
+ * then you can provided them via this context.
  */
 data class FragmentViewModelContext(
     override val activity: FragmentActivity,
@@ -101,11 +105,11 @@ data class FragmentViewModelContext(
     /**
      * The fragment owner of the ViewModel.
      */
-    val fragment: Fragment
+    val fragment: Fragment,
+    override val owner: ViewModelStoreOwner = fragment,
+    override val savedStateRegistry: SavedStateRegistry = fragment.savedStateRegistry
 ) : ViewModelContext() {
 
-    override val owner get() = fragment
-    override val savedStateRegistry get() = fragment.savedStateRegistry
 
     /**
      * Convenience method to type [fragment].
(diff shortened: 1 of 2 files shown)


mvrx/src/main/kotlin/com/airbnb/mvrx/MvRxViewModelFactory.kt
```diff
@@ -101,11 +101,13 @@ data class FragmentViewModelContext(
     /**
      * The fragment owner of the ViewModel.
      */
-    val fragment: Fragment
+    val fragment: Fragment,
+    private val ownerProvider: ViewModelStoreOwner = fragment,
+    private val savedStateRegistryProvider: SavedStateRegistry = fragment.savedStateRegistry
```
gpeal, line 106: Why not just override the vals in the constructor? Can you also add some docs for why somebody would want to use this?

marukami: > Why not just override the vals in the constructor?

The `FragmentViewModelContext` constructor? They're not exposed tho https://github.com/airbnb/MvRx/blob/66469dc2e66e03a129fc30766f458e9648cbc68c/mvrx/src/main/kotlin/com/airbnb/mvrx/MvRxViewModelFactory.kt#L98-L105

Sure I can add some docs. For my current use case, when you use Jetpack Navigation Component you can use the navigation graph ViewModel stores instead of the fragments store. I need a different store then just the fragment I need something like this 
https://github.com/airbnb/MvRx/blob/d8141ff9812a6f00cccf48d89c181a2c13e261df/mvrx-navigation/src/main/kotlin/com/airbnb/mvrx/navigation/MvRxExtensions.kt#L18-L37

gpeal: @marukami Makes sense but it's best to make the docs ultra clear so other people can understand too :)

marukami: @gpeal No worries, updated class docs.

mvrx/src/main/kotlin/com/airbnb/mvrx/MvRxViewModelFactory.kt
```diff
@@ -94,18 +94,24 @@ data class ActivityViewModelContext(
...
 data class FragmentViewModelContext(
     override val activity: FragmentActivity,
     override val args: Any?,
     /**
      * The fragment owner of the ViewModel.
      */
-    val fragment: Fragment
+    val fragment: Fragment,
+    private val ownerProvider: ViewModelStoreOwner = fragment,
+    private val savedStateRegistryProvider: SavedStateRegistry = fragment.savedStateRegistry
```
BenSchwab, line 110: Can we  remove the member properties and just make the the constructor params non-private?

I find it confusing that the params are named `ownerProvider` but the value is the owner itself, not a provider of the owners.

marukami: Ah, yes, thank you. Yer, it is a bit redundant. Updated.

gpeal:
@marukami Could you retarget this to `release/2.0.0` please?

marukami:
> @marukami Could you retarget this to `release/2.0.0` please?

Was on leave and only getting back to things now.

@gpeal done.

marukami:
@gpeal done let me know if the copy looks good or not.

marukami:
@gpeal can you take another look when you're free?

marukami:
Closing as  #443 included the changes here.

PR #390, opened 2020-05-12
Release 1.5.0

elihart:


CHANGELOG.md
```diff
@@ -1,4 +1,17 @@
 # Change log
+## Version 1.5.0
+- Add an optional nullable value to all Async classes (#383)
+- Update various dependencies
+
+Note: MvRx now targets 1.8 for Java/Kotlin, so you may need to update your projects to also target
+1.8
+```
+android {
```
gpeal, line 9: @elihart Did we determine if this is necessary?

elihart: it’s not necessary, but i think the kotlin code it generates will be more efficient if it targets 1.8, and since that is the recommended best practice from android docs and doesn’t have an api minimum i feel fine with it

eboudrant: For our project it was necessary to set the kotlin jvm target. We would get `Cannot inline bytecode built with JVM target 1.8 into bytecode that is being built with JVM target 1.6. Please specify proper '-jvm-target' option` without.

gpeal: @eboudrant Is there a reason this wasn't already set or can't be set for your project?

eboudrant: No, I can't think of a good reason, it is better with it.

PR #394, opened 2020-05-14
[2.0] Make BaseMvRxViewModel not extend ViewModel

gpeal:
https://github.com/airbnb/MvRx/pull/385 but rebased and retargeted to `release/2.0.0`

It would be amazing if we can get BaseMvRxViewModel to be a pojo. That would enable testing on the jvm without robolectric and a potential path to Kotlin Multiplatform.

I was just poking around at what this would look like and it turned out to be fairly simple.

The next hurdle is replacing the inter-vm subscriptions which use Lifecycle with something else but that should be doable.


gpeal:
@denis-bezrukov

PR #396, opened 2020-05-15
[2.0] Removed the need for a base Fragment class

gpeal:
I was able to get the core MvRx functionality working with just the MvRxView interface, no base class needed! This has been one of the most requested features and is a blocker for some people to use MvRx entirely.

https://github.com/airbnb/MvRx/pull/382 should no longer be necessary after this.

##Breaking Change##
Since MvRxView is just an interface, if no ViewModels are used and no MvRx code is run, we can't call `postInvalidate()` to cause invalidate to be called. Simply using a view model delegate is enough to trigger that behavior, however. I think this breaking change is worth it and I made a note of it in the changelog.

@denis-bezrukov


mvrx/src/main/kotlin/com/airbnb/mvrx/MvRxView.kt
```diff
@@ -29,10 +32,12 @@ interface MvRxView : LifecycleOwner {
      * a lifecycle event (e.g. rotation) you should assign a consistent id. Likely this means you should save the id
      * in onSaveInstance state. The viewId will not be accessed until a subscribe method is called.
      * Accessing mvrxViewId before calling super.onCreate() will cause a crash.
-     *
-     * For an example implementation see [BaseMvRxFragment].
      */
-    val mvrxViewId: String
+    val mvrxViewId: String get() = when (this) {
+        is ViewModelStoreOwner -> ViewModelProvider(this).get(MvRxViewIdViewModel::class.java).mvrxViewId
```
gpeal, line 37: This is the key bit of the PR. It stores the view id in a ViewModel with a SavedStateHandle to maintain the id across process restarts.

elihart: @gpeal i guess we're not worried about the extra overhead this adds to lookup the id every time we need to reference it?

mvrx/src/main/kotlin/com/airbnb/mvrx/MvRxView.kt
```diff
@@ -29,10 +32,12 @@ interface MvRxView : LifecycleOwner {
...
+        else -> error("If your MvRxView is not a ViewModelStoreOwner, you must implement mvrxViewId " +
+                "and return a string that is unique to this view and persistant across its entire lifecycle.")
```
elihart, line 39: for this case might it be nice to keep the `MvRxViewId.kt ` class to make it easier to manually save/restore the id?

CHANGELOG.md
```diff
@@ -25,6 +25,7 @@ Previously, setState would only be prioritized at that nesting level so it would
 [W1, S1, W2, S2, S3]
 Now, it will run:
 [W1, S1, S2, S3, W2]
+- If your MvRxView/Fragment does not use any ViewModels, invalidate() will NOT be called in onStart(). In MvRx 1.x, invalidate would be called even if MvRx was not used at all. If you would like to maintain the original behavior, call `postInvalidate()` from onStart in your base Fragment class.
```
elihart, line 28: I _think_ that if the view is dynamically created, like with a RecyclerView, then it needs to be populated (eg via invalidate) before `onViewStateRestored` otherwise scroll position and saved state for items in the list won't be correctly restored.

`invalidate` must be used instead of `postInvalidate` to be synchronous. At airbnb we do it in onActivityCreated in our base fragment. So even if viewmodels are present and MvRx calls postinvalidate for us I think people may still want to manually invoke invalidate earlier

gpeal: @elihart We can recommend something different, this represents the existing behavior though

mvrx/src/main/kotlin/com/airbnb/mvrx/MvRxView.kt
gpeal: @elihart It should just be a map lookup so I think it's okay.

mvrx/src/main/kotlin/com/airbnb/mvrx/MvRxView.kt
gpeal: @elihart That wouldn't be enough on its own though. It needs the state restoration.

PR #398, opened 2020-05-16
[2.0] Split mvrx and mvrx-rxjava2 and remove rxjava dependency from mvrx

gpeal:
This PR is built on https://github.com/airbnb/MvRx/pull/394 and splits `mvrx-rxjava2` into its own artifact.

`BaseMvRxViewModel` is now in the rxjava2 artifact. The original `BaseMvRxViewModel` is now `BaseMavericksViewModel`. All rxjava extensions have been moved into `BaseMvRxViewModel`. Coroutine equivalents will be added to `BaseMvRxViewModel` in a subsequent PR.

No Mavericks logic had to be changed, all of this is just moving code around so yo udon't need to review the implementation of functions for changes.

cc @denis-bezrukov


mvrx-rxjava2/build.gradle
```diff
@@ -0,0 +1,46 @@
...
+apply plugin: "com.android.library"
+apply plugin: "kotlin-android"
+apply plugin: "kotlin-android-extensions"
+apply plugin: "jacoco"
+apply plugin: "com.vanniktech.maven.publish"
+
+android {
+    resourcePrefix "mvrx_"
+
+    testOptions {
```
elihart, line 12: shouldn't this stuff be applied in the buildsrc plugin?

mvrx-rxjava2/src/test/kotlin/com/airbnb/mvrx/DummyJavaFileTest.java
```diff
@@ -0,0 +1,7 @@
+package com.airbnb.mvrx;
+
+/**
+ * This dummy java file is needed to make Buck and Buck Project happy.
```
elihart, line 4: we don't use buck, why is this included?

mvrx/build.gradle
```diff
@@ -31,7 +31,6 @@ jacoco {
 
 dependencies {
     api Libraries.rxAndroid
-    api Libraries.rxJava
```
elihart, line 34: 🎉

mvrx/src/main/kotlin/com/airbnb/mvrx/BaseMavericksViewModel.kt
```diff
@@ -0,0 +1,191 @@
...
+ * abstract class MavericksViewModel<S : MvRxState>(state: S) : BaseMavericksViewModel<S>(state, debugMode = BuildConfig.DEBUG)
+ * ```
+ */
+abstract class BaseMavericksViewModel<S : MvRxState>(
+    initialState: S,
+    debugMode: Boolean,
+    /**
+     * Provide an overridden state store. This should only be used for tests and should only
+     * be exposed via a shared base class within your app. If your features extend this
+     * directly, do not override this in the primary constructor of your feature ViewModel.
```
elihart, line 31: debug mode and store override params will be changing with the mocking framework fyi

mvrx/src/main/kotlin/com/airbnb/mvrx/MvRxViewModelWrapper.kt
```diff
@@ -0,0 +1,10 @@
+package com.airbnb.mvrx
+
+import androidx.lifecycle.ViewModel
+
+class MvRxViewModelWrapper<VM : BaseMavericksViewModel<S>, S : MvRxState>(val viewModel: VM) : ViewModel() {
```
elihart, line 5: as  part of 2.0 release do we want to separate jetpack dependencies from `BaseMavericksViewModel` in separate modules just so if we ever want to support multiplatform we don't have to have breaking changes again? would be nice to get all our module names solidified for that

mvrx-rxjava2/src/test/kotlin/com/airbnb/mvrx/DummyJavaFileTest.java
gpeal: @elihart It was copied over from the mvrx module. I can definitely delete it.

mvrx-rxjava2/build.gradle
gpeal: @elihart This is all still in the main mvrx gradle file. These aren't set by the plugin. We could add it if we want.

mvrx/src/main/kotlin/com/airbnb/mvrx/BaseMavericksViewModel.kt
gpeal: @elihart Yup! No problem :)

mvrx/src/main/kotlin/com/airbnb/mvrx/MvRxViewModelWrapper.kt
gpeal: @elihart Agreed!

PR #399, opened 2020-05-17
[2.0] Move all subscribe functionality into BaseMavericksViewModel as onEach

gpeal:
This PR is rebased on https://github.com/airbnb/MvRx/pull/398

onEach/onAsync is the Mavericks equivalent on subscribe/selectSubscribe/asyncSubscribe.

This PR moves all of the functionality into BaseMavericksViewModel and makes BaseMvRxViewModel a proxy.

The next step after this is creating coroutines versions of execute.

cc @denis-bezrukov


mvrx/src/main/kotlin/com/airbnb/mvrx/DeliveryMode.kt
```diff
@@ -0,0 +1,39 @@
...
+ * Defines what updates a subscription should receive.
+ * See: [RedeliverOnStart], [UniqueOnly].
+ */
+sealed class DeliveryMode {
+
+    @RestrictTo(RestrictTo.Scope.LIBRARY_GROUP)
+    fun appendPropertiesToId(vararg properties: KProperty1<*, *>): DeliveryMode {
+        return when (this) {
+            is RedeliverOnStart -> RedeliverOnStart
+            is UniqueOnly -> UniqueOnly(subscriptionId + "_" + properties.joinToString(",") { it.name })
```
denis-bezrukov, line 16: nit: 
```suggestion
            is UniqueOnly -> UniqueOnly(properties.joinToString(",", prefix = subscriptionId + "_") { it.name })
```
will allocate slightly less memory

mvrx/src/main/kotlin/com/airbnb/mvrx/MavericksView.kt
```diff
@@ -0,0 +1,257 @@
...
+     *
+     * Use [uniqueOnly] to automatically create a [UniqueOnly] mode with a unique id for this view.
+     *
+     * Default: [RedeliverOnStart].
+     */
+    fun <S : MvRxState, A> BaseMavericksViewModel<S>.onEach(
+        prop1: KProperty1<S, A>,
+        deliveryMode: DeliveryMode = RedeliverOnStart,
+        action: suspend (A) -> Unit
+    ) = onEach1Internal(subscriptionLifecycleOwner, prop1, deliveryMode, { action(it) })
```
denis-bezrukov, line 107:
```suggestion
    ) = onEach1Internal(subscriptionLifecycleOwner, prop1, deliveryMode, action)
```

mvrx/src/main/kotlin/com/airbnb/mvrx/MavericksView.kt
```diff
@@ -0,0 +1,257 @@
...
+     *
+     * Use [uniqueOnly] to automatically create a [UniqueOnly] mode with a unique id for this view.
+     *
+     * Default: [RedeliverOnStart].
+     */
+    fun <S : MvRxState> BaseMavericksViewModel<S>.onEach(deliveryMode: DeliveryMode = RedeliverOnStart, action: suspend (S) -> Unit) =
+        onEachInternal(subscriptionLifecycleOwner, deliveryMode, action)
+
+    /**
+     * Subscribes to state changes for only a specific property and calls the subscribe with
```
denis-bezrukov, line 91:
```suggestion
     * Subscribes to state changes for only a specific property and calls the onEach with
```
?

mvrx/src/main/kotlin/com/airbnb/mvrx/MavericksView.kt
```diff
@@ -0,0 +1,257 @@
...
+     * only that single property.
```
denis-bezrukov, line 92: Could you mention, that if `action` supports cooperative cancellation, previous block will be cancelled?

mvrx/src/main/kotlin/com/airbnb/mvrx/BaseMavericksViewModel.kt
```diff
@@ -389,10 +441,9 @@ abstract class BaseMavericksViewModel<S : MvRxState>(
         } else {
             flowWhenStarted(lifecycleOwner)
         }
-        val scope = lifecycleOwner?.lifecycleScope ?: viewModelScope
-        return scope.launch {
-            flow.collectLatest { action(it) }
-        }
+        return viewModelScope.launch {
+            flow.collectLatest(action)
+        }.cancelOnClear(lifecycleOwner?.lifecycleScope)
```
denis-bezrukov, line 446: Is there a reason for this change (comparing to just launch in `lifecycleOwner?.lifecycleScope ?: viewModelScope`)? Previous version looks more idiomatic and also behavior differs

mvrx/src/main/kotlin/com/airbnb/mvrx/BaseMavericksViewModel.kt
```diff
@@ -160,55 +160,95 @@ abstract class BaseMavericksViewModel<S : MvRxState>(
...
      */
-    protected fun withState(block: (state: S) -> Unit) {
-        stateStore.get(block)
+    protected fun withState(action: suspend (state: S) -> Unit) {
+        stateStore.get { state ->
+            viewModelScope.launch {
+                action(state)
+            }
+        }
     }
```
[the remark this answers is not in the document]
denis-bezrukov:
```
withState { // W1
   setState { } // S1
}
withState { // W2
}
```
is a correct example (order W1, W2, S1 is possible)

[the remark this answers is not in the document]
gpeal: @denis-bezrukov. I'm going to revert that for this PR since I think it's out of scope. We can evaluate it subsequently.

mvrx/src/main/kotlin/com/airbnb/mvrx/BaseMavericksViewModel.kt
gpeal: @denis-bezrukov Yeah, before [this test](https://github.com/airbnb/MvRx/blob/master/mvrx/src/test/kotlin/com/airbnb/mvrx/InterViewModelSubscriberTest.kt#L251) was broken.

The tricky thing here is that we don't know which scope has a shorter lifecycle so we have to cancel on whichever is cancelled first.

denis-bezrukov: @gpeal as I see that just means we don't close CoroutineStateStore.stateChannel properly. So what do you think about just closing it when store's scope cancelled?
Then you will be able to subscribe here as previously. 

It seems that it will provide better expectations and cleaner implementation.

denis-bezrukov: we need to make flush as:
```
try {
    while (isActive) {
        flushQueuesOnce()
    }
} catch (e: Exception) {
    stateChannel.cancel(e as? CancellationException)
    throw e
}
```

mvrx/src/main/kotlin/com/airbnb/mvrx/BaseMavericksViewModel.kt
```diff
@@ -146,46 +160,335 @@ abstract class BaseMavericksViewModel<S : MvRxState>(
...
+    private operator fun CoroutineContext.plus(other: CoroutineContext?) = if (other == null) this else this + other
+
+    private fun <S : MvRxState> assertSubscribeToDifferentViewModel(viewModel: BaseMavericksViewModel<S>) {
+        require(this != viewModel) {
+            "This method is for subscribing to other view models. Please pass a different instance as the argument."
+        }
+    }
+
+    private fun Job.cancelOnClear(scope: CoroutineScope?): Job {
+        scope ?: return this
```
elihart, line 486: since you always return "this", you can remove this line and use a safe call on scope below?

mvrx/src/main/kotlin/com/airbnb/mvrx/BaseMavericksViewModel.kt
gpeal: @denis-bezrukov. Good point. Updated

PR #403, opened 2020-05-21
[2.0] Add coroutines versions of execute and setOnEach

gpeal:
This PR adds all of the coroutine equivalents for execute:
* suspend () -> T
* Deferred<T>
* Flow<T>
And a new API:
Flow<T>.setOnEach which is like onEach + setState + launchIn(viewModelScope)

cc @denis-bezrukov

### DIFF
--- a/mvrx/src/main/kotlin/com/airbnb/mvrx/BaseMavericksViewModel.kt
+++ b/mvrx/src/main/kotlin/com/airbnb/mvrx/BaseMavericksViewModel.kt
@@ -5,7 +5,10 @@ import androidx.annotation.RestrictTo
 import androidx.lifecycle.DefaultLifecycleObserver
 import androidx.lifecycle.LifecycleOwner
 import androidx.lifecycle.lifecycleScope
+import kotlinx.coroutines.CancellationException
+import kotlinx.coroutines.CoroutineDispatcher
 import kotlinx.coroutines.CoroutineScope
+import kotlinx.coroutines.Deferred
 import kotlinx.coroutines.Dispatchers
 import kotlinx.coroutines.Job
 import kotlinx.coroutines.SupervisorJob
@@ -14,10 +17,12 @@ import kotlinx.coroutines.flow.Flow
 import kotlinx.coroutines.flow.collectLatest
 import kotlinx.coroutines.flow.distinctUntilChanged
 import kotlinx.coroutines.flow.dropWhile
+import kotlinx.coroutines.flow.launchIn
 import kotlinx.coroutines.flow.map
 import kotlinx.coroutines.flow.onCompletion
 import kotlinx.coroutines.flow.onEach
 import kotlinx.coroutines.launch
+import kotlinx.coroutines.plus
 import java.util.Collections
 import java.util.concurrent.ConcurrentHashMap
 import kotlin.coroutines.CoroutineContext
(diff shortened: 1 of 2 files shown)


gpeal:
@BenSchwab I chose Main.immediate because that tends to be the default for Jetpack scopes and has become the expectation for a default dispatcher on Android. However, your point is valid. It can be aided by https://github.com/airbnb/MvRx/issues/389 though.

I'm going to merge this but opened https://github.com/airbnb/MvRx/issues/404 to continue the discussion

mvrx/src/main/kotlin/com/airbnb/mvrx/BaseMavericksViewModel.kt
```diff
@@ -164,6 +169,84 @@ abstract class BaseMavericksViewModel<S : MvRxState>(
...
+     *
+     * @param dispatcher The coroutine dispatcher that the coroutine will run on. Defaults to [Dispatchers.Main.immediate].
+     * @param retainValue A state property that, when set, will be called to retrieve an optional existing data value that will be retained across
+     *                    subsequent Loading and Fail states. This is useful if you want to display the previously succcessful data when
+     *                    refreshing.
+     * @param reducer A reducer that is applied to the current state and should return the new state. Because the state is the receiver
+     *                and it likely a data class, an implementation may look like: `{ copy(response = it) }`.
+     */
+    fun <T : Any?> Deferred<T>.execute(
+        dispatcher: CoroutineDispatcher = Dispatchers.Main.immediate,
```
elihart, line 183: @gpeal @BenSchwab what if we pull the dispatcher out of the viewmodelscope instead and use that as the default? something like
```
dispatcher: CoroutineDispatcher = viewModelScope.coroutineContext[ContinuationInterceptor]
```

The ContinuationInterceptor is used as the context key for the CoroutineDispatcher class which should let us pull out the dispatcher. This would let us default to whichever dispatcher the scope uses

mvrx/src/main/kotlin/com/airbnb/mvrx/BaseMavericksViewModel.kt
```diff
@@ -164,6 +169,84 @@ abstract class BaseMavericksViewModel<S : MvRxState>(
...
 
+    /**
+     * Run a coroutine and wrap its progression with [Async] property reduced to the global state.
+     *
+     * @param dispatcher The coroutine dispatcher that the coroutine will run on. Defaults to [Dispatchers.Main.immediate].
+     * @param retainValue A state property that, when set, will be called to retrieve an optional existing data value that will be retained across
+     *                    subsequent Loading and Fail states. This is useful if you want to display the previously succcessful data when
+     *                    refreshing.
+     * @param reducer A reducer that is applied to the current state and should return the new state. Because the state is the receiver
+     *                and it likely a data class, an implementation may look like: `{ copy(response = it) }`.
```
elihart, line 180: isn't state required to be a data class?

mvrx/src/main/kotlin/com/airbnb/mvrx/BaseMavericksViewModel.kt
```diff
@@ -164,6 +169,84 @@ abstract class BaseMavericksViewModel<S : MvRxState>(
...
+        reducer: S.(Async<T>) -> S
+    ): Job {
+        setState { reducer(Loading(value = retainValue?.get(this)?.invoke())) }
+        return viewModelScope.launch(dispatcher) {
+            try {
+                val result = invoke()
+                setState { reducer(Success(result)) }
+            } catch (e: CancellationException) {
+                @Suppress("RethrowCaughtException")
+                throw e
```
elihart, line 210: why catch this if it's just going to be rethrown.

Also, should we have a discussion about expected behavior on cancellation? with this, if the job is cancelled the state will be left Loading indefinitely, which seems misleading. we could provide a way to handle this more clearly, like a Async.Canceled or we could set the property back to Uninitialized.

or maybe it makes sense to represent it as Fail? or make a Cancel object that is a subclass of Fail?

mvrx/src/main/kotlin/com/airbnb/mvrx/BaseMavericksViewModel.kt
```diff
@@ -164,6 +169,84 @@ abstract class BaseMavericksViewModel<S : MvRxState>(
...
+
+    /**
+     * Collect a Flow and wrap its progression with [Async] property reduced to the global state.
+     *
+     * @param dispatcher The coroutine dispatcher that the coroutine will run on. Defaults to [Dispatchers.Main.immediate].
+     * @param reducer A reducer that is applied to the current state and should return the new state. Because the state is the receiver
+     *                and it likely a data class, an implementation may look like: `{ copy(response = it) }`.
+     */
+    fun <T> Flow<T>.execute(
+        dispatcher: CoroutineDispatcher = Dispatchers.Main.immediate,
```
elihart, line 225: could dispatcher be nullable here with a null default? if left null we can just use viewModelScope

mvrx/src/main/kotlin/com/airbnb/mvrx/BaseMavericksViewModel.kt
```diff
@@ -164,6 +169,84 @@ abstract class BaseMavericksViewModel<S : MvRxState>(
...
+        reducer: S.(Async<T>) -> S
+    ): Job {
+        setState { reducer(Loading<T>()) }
+        return onEach {
```
elihart, line 229: should we handle failures and/or cancellation?

mvrx/src/main/kotlin/com/airbnb/mvrx/BaseMavericksViewModel.kt
```diff
@@ -164,6 +169,84 @@ abstract class BaseMavericksViewModel<S : MvRxState>(
...
+    }
+
+    /**
+     * Collect a Flow and update state each time it emits a value. This is functionally the same as wrapping onEach with a setState call.
+     *
+     * @param dispatcher The coroutine dispatcher that the coroutine will run on. Defaults to [Dispatchers.Main.immediate].
+     * @param reducer A reducer that is applied to the current state and should return the new state. Because the state is the receiver
+     *                and it likely a data class, an implementation may look like: `{ copy(response = it) }`.
+     */
+    fun <T> Flow<T>.setOnEach(
```
elihart, line 241: we have a similar extension at airbnb named `executeWithoutAsync`. kind of an ugly name, but I like that it 1) is discoverable since people are used to the "execute" function and autocomplete helps find this. 2) helps to explain the difference with "execute"

What do you think of some other name for this and starts with execute to help get those benefits?

mvrx/src/main/kotlin/com/airbnb/mvrx/BaseMavericksViewModel.kt
gpeal: @elihart Yup, I just copied these docs from the old execute function but you're right :)

mvrx/src/main/kotlin/com/airbnb/mvrx/BaseMavericksViewModel.kt
denis-bezrukov: since coroutines 1.3.x CoroutineDispatcher is coroutine context key too, so it can be pulled directly:
```
viewModelScope.coroutineContext[CoroutineDispatcher]
```

mvrx/src/main/kotlin/com/airbnb/mvrx/BaseMavericksViewModel.kt
gpeal: @elihart I see your point but I do think that `setOnEach` is more accurate since it really has nothing to do with execute at all.

PR #411, opened 2020-05-25
Added docsify docs

gpeal:
This is the first round of updating docs for Mavericks 2.0. I used [docsify](https://docsify.js.org/) which I use for Lottie [here](http://airbnb.io/lottie). We should be able to get `airbnb.io/mvrx` to point to these.
Docsify is nice because you just write markdown and it generates the page locally so no html generation is necessary.

About half of the docs are new/rewritten and half are copied over from the wiki and lightly edited.

You can try this out locally with `npm i -g docsify` then `docsify serve ./docs`

This is just a first pass. We'll need to clean these up and add more before release. Any help making the documentation more robust would be greatly appreciated.


docs/core-concepts.md
```diff
@@ -0,0 +1,148 @@
...
+1. Easy for you and other engineer to reason through
+1. Renders the same independently of the order of events leading up to it
+1. Powerful enough to render any type of screen
+
+Mavericks will also enforce that your state class:
+1. Is a kotlin data class
+1. Uses only immutable properties
+1. Has default values for every property to ensure that your screen can be rendered immediately
+
+This concept makes reasoning about and testing a screen trivially easy because given a state class, you can have high confidence that you screen will look correct.
```
elihart, line 16:
```suggestion
This concept makes reasoning about and testing a screen trivially easy because given a state class, you can have high confidence that your screen will look correct.
```

docs/core-concepts.md
```diff
@@ -0,0 +1,148 @@
+# Core Concepts
+Mastering Mavericks only requires using three classes: MavericksState, MavericksViewModel, MavericksView.
+
+# MavericksState
```
elihart, line 4: worth mentioning the point of the `MavericksState` interface?

docs/core-concepts.md
```diff
@@ -0,0 +1,148 @@
...
+1. You can subscribe to changes to just these derived properties using `onEach`.
+1. It will _never_ be out of sync with the underlying state
+1. It is extremely easy to reason about
+1. It is extremely easy to write unit tests for
+
+# MavericksViewModel
+
+A ViewModel is responsible for:
+1. Updating state
+2. Exposing a stream of states for other to subscribe to
```
elihart, line 49:
```suggestion
2. Exposing a stream of states for other places? to subscribe to
```

docs/core-concepts.md
```diff
@@ -0,0 +1,148 @@
...
+
+#### Updating state
+From within a viewModel, you call `setState { copy(yourProp = newValue) }`. If this syntax is unfamiliar:
+1. The signature of the lambda is `S.() -> S` meaning the receiver (aka `this`) of the lambda is the current state when the lambda is invoked and it returns the new lambda
```
elihart, line 53:
```suggestion
1. The signature of the lambda is `S.() -> S` meaning the receiver (aka `this`) of the lambda is the current state when the lambda is invoked and it returns the new state
```

docs/core-concepts.md
```diff
@@ -0,0 +1,148 @@
...
+1. `copy` comes from the fact that the state class is a Kotlin [data class](https://kotlinlang.org/docs/reference/data-classes.html)
+1. The lambda is _not_ executed synchronously. It is put on a queue and run on a background thread. See [threading](threading.md) for more information
+
+#### Subscribing to state changes
+You can subscribe to state changes in your ViewModel. You may want to do this for analytics, for example. This usually done in the `init { ... }` block.
```
elihart, line 58:
```suggestion
You can subscribe to state changes in your ViewModel. You may want to do this for analytics, for example. This is usually done in the `init { ... }` block.
```

docs/core-concepts.md
```diff
@@ -0,0 +1,148 @@
...
+
+```kotlin
+onEach { state ->
```
elihart, line 61: maybe add comments
```
// Invoked with each change in state

// Invoked only when property A is changed, passing the new value of A

etc

File: docs/core-concepts.md [hunk 7]
```diff
@@ -0,0 +1,148 @@
...
+```kotlin
+onEach(YourState::propA) { a ->
+}
+onEach(YourState::propA, YourState::propB, YourState::propC) { a, b, c ->
+}
+```
+If you are calling `setState` from within an `onEach` block, consider using a derived property.
+
+#### stateFlow
+`MavericksViewModel` exposes a `stateFlow` which is a normal Kotlin Flow and can be used however you would like. Helpers such as `onEach` above are just wrappers around it with automatic lifecycle cancellation.
```
elihart, line 74: maybe worth stating the obvious that the flow exposes all future state changes?

docs/core-concepts.md
```diff
@@ -0,0 +1,148 @@
...
+#### Accessing state once
+If you just want to retrieve the value of state one time, you can use `withState { state -> ... }`.
+
+When called from within a ViewModel, this will _not_ be run synchronously. It will be placed on a background queue so that all pending `setState` reducers are called prior to your `withState` call.
+
+# MavericksView
+`MavericksView` is where you actually render your state class to the screen. Most of the time, this will be a Fragment but it doesn't have to be.
+By implementing `MavericksView`, you:
+1. You can get access to a `MavericksViewModel` via any of the view model delegates. Doing so will automatically subscribe to changes and call `invalidate()`.
+1. Override `invalidate()`. It will get called any time the state for any view model accessed by the above delegates changes.
```
elihart, line 85:
```suggestion
1. Override the `invalidate()` function. It is called any time the state for any view model accessed by the above delegates changes. `invalidate()` is used to redraw the UI on each state change
```

docs/core-concepts.md
```diff
@@ -0,0 +1,148 @@
...
+
+#### ViewModel delegates
+1. `activityViewModel()` scopes the ViewModel to the Activity. All Fragments within the Activity that request a ViewModel of this type will receive the same instance. Useful for sharing data between screens.
+1. `fragmentViewModel()` scopes the ViewModel to the Fragment. It will be accessible to children fragments but parent or sibling fragments would get a different instance.
+1. `parentFragmentViewModel()` walks up the parent fragment tree until it finds one that has already created a ViewModel of the desired type. If none is found, it will automatically create one scoped to the highest parent fragment.
+1. `existingViewModel()` Same as `activityViewModel()` except it throws if the ViewModel was not already created by somebody else.
+
+If you want multiple ViewModels of the same type, you can pass a `keyFactory` into any of the delegates.
+
+#### Subscribing to state manually
```
elihart, line 95: this seems to be covered in the `Subscribing to state changes` section above

docs/core-concepts.md
```diff
@@ -0,0 +1,148 @@
...
+Most of the time, overriding `invalidate()` and updating your views is sufficient. However, if you want to subscribe to state to do things like start animations, you may call any of the `onEach` subscriptions on your ViewModel. If your view is a Fragment, these subscriptions should be set up in `onCreate()`.
+
+#### Accessing state once
```
elihart, line 98: already covered in previous `Accessing state once` section

docs/core-concepts.md
```diff
@@ -0,0 +1,148 @@
...
+For each event it emits, it will call the `reducer` which takes the current state and returns an updated state just like `setState`
+
+Executing a network request looks like:
+```kotlin
+suspend {
+    weatherApi.fetchTemperature()
+}.execute { copy(currentTemperature = it) }
+```
+In this case:
+* `currentTemperature` is of type `Async<Int>` and set to `Uninitialized`
```
elihart, line 141:
```suggestion
* `currentTemperature` is of type `Async<Int>` and originally set to `Uninitialized`
```

docs/counter.md
```diff
@@ -0,0 +1,81 @@
...
+![Counter](/images/counter.png)
+
+When you click the counter text, the count goes up by one.
+Code for this sample can be found [here](https://github.com/airbnb/MvRx/tree/master/counter).
+
+## Create CounterState
+
+The first step in creating a Mavericks screen is to model it as a function of state. Modeling a screen as a function of state is a useful concept because it is:
+1. Easy to test
+1. Easy for you and other engineer to reason through
```
elihart, line 14:
```suggestion
1. Easy for you and other engineers to reason through
```

docs/counter.md
```diff
@@ -0,0 +1,81 @@
...
+1. Renders the same independently of the order of events leading up to it
+1. Powerful enough to render any type of screen
+
+A counter screen is simple to model.
+```kotlin
+data class CounterState(val count: Int = 0) : MavericksState
+```
+* CounterState contains a single immutable (`val`) property, count, it is an integer, and it defaults to 0.
+* CounterState is a [data class](https://kotlinlang.org/docs/reference/data-classes.html) so that Kotlin creates equals and hashCode automatically.
+* CounterState implements `MavericksState` signals the intention that this class will be used as Mavericks state.
```
elihart, line 24:
```suggestion
* CounterState implementation of `MavericksState` signals the intention that this class will be used as Mavericks state.
```

docs/counter.md
```diff
@@ -0,0 +1,81 @@
...
+Next, we need to create a ViewModel which will own CounterState and be responsible for updating it.
+```kotlin
+class CounterViewModel(initialState: CounterState) : MavericksViewModel<CounterState>(state) {
+    fun incrementCount() {
+      setState { copy(count = count + 1) }
+    }
+}
+```
+* CounterViewModel takes CounterState as a constructor parameter because Mavericks creates your initial state for you. It does this to handle [saved state](/saved-state.md) across process restoration.
+* incrementCount() calls setState which takes a reducer with the type `S.() -> S`. The receiver of the reducer is the current state when the reducer is run and it returns the updated state. `copy` is the kotlin data class copy function.
```
elihart, line 37:
```suggestion
* incrementCount() calls setState which takes a reducer with the type `CounterState.() -> CounterState `. The receiver of the reducer is the current state when the reducer is run and it returns the updated state. `copy` is the kotlin data class copy function.
```

docs/debug-checks.md
```diff
@@ -0,0 +1,9 @@
+# Debug Checks
+
+MvRx has a debug mode that enables some new checks. When you integrate MvRx into your app, you will need to create your own MvRxViewModel that extends BaseMvRxViewModel. One of the required things to override is protected abstract val debugMode: Boolean. You could set it to something like BuildConfig.DEBUG. The checks include:
```
elihart, line 3:
```suggestion
MvRx has a debug mode that enables some validation checks. When you integrate MvRx into your app, you will need to create your own MvRxViewModel that extends BaseMvRxViewModel. One of the required things to override is protected abstract val debugMode: Boolean. You could set it to something like BuildConfig.DEBUG. The checks include:
```

elihart: `that extends BaseMvRxViewModel` - not true with my PR

docs/dependency-injection.md
```diff
@@ -0,0 +1,23 @@
+# Depenency Injection
```
elihart, line 1:
```suggestion
# Dependency Injection
```

docs/debug-checks.md
```diff
@@ -0,0 +1,9 @@
...
+
+* Running all state reducers twice and ensuring that the resulting state is the same. This ensures that reducers are pure. To understand this better, read the Redux prerequisite concepts.
```
elihart, line 5: don't think `Redux` is actually mentioned anywhere else right now

docs/README.md
```diff
@@ -0,0 +1,40 @@
...
+This is what it looks like:
+```kotlin
+data class CounterState(@PersistState val count: Int = 0) : MvRxState
+
+class CounterViewModel(state: CounterState) : MavericksViewModel<CounterState>(state) {
+    fun incrementCount() = setState { copy(count = count + 1) }
+}
+
+class CounterFragment : Fragment(R.layout.fragment_counter), MavericksView {
+    private val viewModel: CounterViewModel by activityViewModel()
```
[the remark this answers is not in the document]
gpeal: It could be done with kotlin synthetics so I left it out for brevity

docs/core-concepts.md
```diff
@@ -0,0 +1,148 @@
+# Core Concepts
+Mastering Mavericks only requires using three classes: MavericksState, MavericksViewModel, MavericksView.
+
+# MavericksState
+The first step in creating a Mavericks screen is to model it as a function of state. Modeling a screen as a function of state is a useful concept because it is:
+1. Easy to test
+1. Easy for you and other engineer to reason through
+1. Renders the same independently of the order of events leading up to it
+1. Powerful enough to render any type of screen
```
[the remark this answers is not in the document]
gpeal: @elihart yes, Markdown auto-numbers so you don't have to manually renumber things if you add a new item

docs/core-concepts.md
gpeal: That is in the ViewModel section, this is the View section

docs/core-concepts.md
gpeal: Same thing, ViewModel vs View

docs/debug-checks.md
gpeal: We can update this once your PR lands. We'll want to audit the flesh out the docs prior to releasing anyway

docs/dependency-injection.md
```diff
@@ -0,0 +1,23 @@
+# Dependency Injection
+
+### Constructor Dependency Injection
+MvRx supports adding dependencies via constructor injection. To leverage this, create a `companion object` in your ViewModel and have it implement ` MvRxViewModelFactory<YourViewModel, YourState>`. That will force you to implement a `create` function. It will pass in the `ViewModelContext` that contains a reference to the activity that hosts the fragment, and a reference to the fragment if the view model was created with a fragment scope. You should use the activity/fragment, if needed, to get or create your dagger/DI component and inject what you need. **DO NOT** pass a reference of your activity to your ViewModel. It will leak your Activity. Pass the application context if you must.
```
elihart, line 4: this made me think, could we have a runtime debug check for any context in the view model constructor params? can use instance of Activity to differentiate application context or something

docs/epoxy.md
```diff
@@ -0,0 +1,91 @@
+# Epoxy
+
+[Epoxy](https://github.com/airbnb/epoxy) is an open source library by Airbnb that makes it easier to build complex  layouts. It builds on top of RecyclerView, and allows you to work with multiple view types in a declarative, reactive way.
+
+### Epoxy's Philosophy
+Epoxy uses immutable "models" to describe the interface between data and view. An `EpoxyModel` class exists for each unique view in your layout, and defines how the view is created, styled, and bound to data. Epoxy provides a powerful annotation processor to generate these models for you based on custom views or databinding layouts.
+
+To create the layout of a page, EpoxyModels are declared in the order that views should appear, with the data that should be bound to them. This is called "building models". Since the models are immutable they must be rebuilt whenever the data changes to provide a new snapshot that describes the view.
+
+This looks like this (adopted from the MvRx sample app):
```
elihart, line 10:
```suggestion
This looks like this (adopted from the Mavericks sample app):
```

spose we'll do a big find/replace before releasing?

docs/epoxy.md
```diff
@@ -0,0 +1,91 @@
...
+  }
+
+  loadingRow {
+    onBind { _, _, _ -> viewModel.fetchNextPage() }
+  }
+}
+```
+
+### Combining with MvRx
+The model building pattern of Epoxy enforces a one way data flow. When a state change occurs, models are rebuilt and Epoxy runs a diff to figure out what changed. This fits in perfectly with the reactive, state based approach of MvRx.
```
elihart, line 38:
```suggestion
The model building pattern of Epoxy enforces a one way data flow. When a state change occurs, models are rebuilt asynchronously and Epoxy runs a background diff to figure out what changed. Changes are then dispatched on the main thread to the RecyclerView. This fits in perfectly with the reactive, state based approach of MvRx.
```

docs/epoxy.md
```diff
@@ -0,0 +1,91 @@
...
+
+A pattern we use at Airbnb, and that we share in the sample app, is that each fragment contains an EpoxyController that defines how models are built. MvRx automatically manages this, and rebuilds the models whenever the state changes and the view is invalidated. This can be built into a base fragment, so that individual feature fragments have very little overhead in declaring a new page.
+
+An abridged version looks like this: (The complete code can be found in the sample app)
+```kotlin
+abstract class BaseFragment : BaseMvRxFragment() {
```
elihart, line 44:
```suggestion
abstract class BaseFragment : Fragment() {
```

docs/epoxy.md
```diff
@@ -0,0 +1,91 @@
+# Epoxy
+
+[Epoxy](https://github.com/airbnb/epoxy) is an open source library by Airbnb that makes it easier to build complex  layouts. It builds on top of RecyclerView, and allows you to work with multiple view types in a declarative, reactive way.
```
elihart, line 3: maybe up here we can stress that Epoxy can optionally be used with MvRx as the view layer

docs/fragment-arguments.md
```diff
@@ -0,0 +1,31 @@
...
+Although Mavericks, doesn't have to be used with Fragments, it has some helpers in case it is.
+
+### Setting Mavericks Fragment arguments
+When creating a Fragment:
+1. Create a `Parcelable` or `Serializable` object
+2. Set the Fragment arguments to `yourArgs.asMavericksArgs()`
+
+```kotlin
+val fragment = YourFragment()
+fragment.arguments = yourArgs.asMavericksArgs()
```
elihart, line 12: I would if we could also/alternatively have an extension `fragment.setMvRxArgs(Parcelable)`

docs/fragment-arguments.md
```diff
@@ -0,0 +1,31 @@
...
+
+### Retrieving your arguments in your Fragment
+Create a property in your fragment:
+```kotlin
+private val args: YourArgsType by args()
+```
+That's it!
+
+### Using Fragment args in the initial value for `MavericksState`
+Create a secondary constructor that has 1 parameter of the same type as your Fragment arguments. Mavericks will automatically call this instead.
```
elihart, line 23:
```suggestion
Create a secondary constructor that has 1 parameter of the same type as your Fragment arguments. Mavericks will automatically get the arguments off your Fragment and call this constructor with them when creating your State.
```

docs/index.html
```diff
@@ -0,0 +1,46 @@
+<!DOCTYPE html>
+<html lang="en">
+<head>
+  <meta charset="UTF-8">
+  <title>Lottie Docs</title>
```
elihart, line 5:
```suggestion
  <title>Mavericks Docs</title>

```

docs/persist-state.md
```diff
@@ -0,0 +1,18 @@
+# Persistence
+In Android, there are two major cases in which persistence must be considered:
+
+#### Configuration Changes / Back Stack
+Traditionally, you would use savedInstanceState to save important information. However, with Mavericks, you will get the exact same instance of your ViewModel back so you have everything you need to re-render the view. Generally, screens won't need to implement savedInstanceState with Mavericks.
```
elihart, line 5: Maybe worth mentioning this is standard Jetpack ViewModel behavior, and link to jetpack docs if there are related ones

docs/proguard.md
```diff
@@ -0,0 +1,60 @@
+# Proguard/Dexguard/R8
+
+The MvRx lib itself has a consumer Proguard file that configures Proguard. As the lib uses Kotlin, Kotlin reflection and RxJava you might have to add rules to handle those to the Proguard rules of your app itself. (If you did not add them for other purposes already). These are not part of the consumer Proguard file, as they are not part of the lib specific Proguard configuration. This is what that part of the configuration looks like for our sample app:
```
elihart, line 3:
```suggestion
The Mavericks lib itself has a consumer Proguard file that configures Proguard. As the main lib uses Kotlin, and optional add-on modules use Kotlin reflection and RxJava, you might have to add rules to handle those to the Proguard rules of your app itself. (If you did not add them for other purposes already). These are not part of the consumer Proguard file, as they are not part of the lib specific Proguard configuration. This is what that part of the configuration looks like for our sample app:
```

docs/threading.md
```diff
@@ -0,0 +1,32 @@
+# Threading
+
+One challenging aspect of Android is that everything happens on the main thread by default. Interacting with views must be there but many other things such as business logic are only there because it is too much work to do so otherwise. One of the core tenants of Mavericks is that it is thread-safe. Everything non-view related in Mavericks can and does run on background threads. Mavericks abstracts away most of the challenges of multi-threading. However, it is important to be aware of this when using Mavericks.
```
elihart, line 3:
```suggestion
One challenging aspect of Android is that everything happens on the main thread by default. Interacting with views must be done there but many other things such as business logic are only there because it is too much work to do otherwise. One of the core tenants of Mavericks is that it is thread-safe. Everything non-view related in Mavericks can and does run on background threads. Mavericks abstracts away most of the challenges of multi-threading. However, it is important to be aware of this when using Mavericks.
```

docs/dependency-injection.md
gpeal: @elihart Yeah, good idea or`context == context.applicationContext`

PR #421, opened 2020-06-21
Fix failing data class check

elihart:
I noticed that the immutability check for data classes was incorrectly failing in our project. After investigating I realized that internal properties have a componentN function name with a `$module_name` suffix. I fixed the check for data classes, and also had to fix the copy function for restoring state.

I've made a 1.5.1 release for this already


gpeal:
@elihart Can you put a PR up against 2.0?

elihart:
I already cherry picked it onto the branch

PR #422, opened 2020-06-28
[2.0] Add BaseMvRxViewModel subscribe for backwards compatiblity.

gpeal:
While testing the 2.0 upgrade at Tonal, it looks like we're using this restricted API so to improve backwards compatiblity, let's re-add it.

Oddly, there is no IDE warning even though the inspection is set to error for us.

@elihart With this change, I think we can start testing this at Tonal in production.

### DIFF
--- a/mvrx-rxjava2/src/main/kotlin/com/airbnb/mvrx/BaseMvRxViewModel.kt
+++ b/mvrx-rxjava2/src/main/kotlin/com/airbnb/mvrx/BaseMvRxViewModel.kt
@@ -1,6 +1,7 @@
 package com.airbnb.mvrx
 
 import android.util.Log
+import androidx.annotation.RestrictTo
 import androidx.lifecycle.Lifecycle
 import androidx.lifecycle.LifecycleOwner
 import androidx.lifecycle.LifecycleRegistry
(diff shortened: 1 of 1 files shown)


mvrx-rxjava2/src/main/kotlin/com/airbnb/mvrx/BaseMvRxViewModel.kt
```diff
@@ -146,6 +147,18 @@ abstract class BaseMvRxViewModel<S : MvRxState>(
      */
     protected fun subscribe(subscriber: (S) -> Unit): Disposable = onEachInternal(null, action = { subscriber(it) }).toDisposable()
 
+    /**
+     * Subscribe to state when this LifecycleOwner is started.
+     */
+    @RestrictTo(RestrictTo.Scope.LIBRARY)
```
elihart, line 153: i'm a bit confused, did you mean to restrict this to the library? I thought you wanted to expose it for you?

gpeal: @elihart I left it with the same visibility as it is in 1.0. As mentioned, for some reason, Android Studio never showed us the warning for this.

denis-bezrukov: @gpeal if I remember correctly, lint started showing this warning after upgrading to agp 4.0

PR #427, opened 2020-07-27
Force dispatch onEach/selectSubscribe coroutine

denis-bezrukov:
It fixes behavior change introduced in 2.0.
Since 2.0 selectSubscribe/onEach/invalidate is invoked (first time) with onStart in the stacktrace because we use Dispatchers.Main.immediate in VM scope. It's ok in most of the cases, but in tricky cases like navigation in may lead to issues. 

The fix helps to dispatch coroutine even if  isDispatchNeeded = true in scope's dispatcher.

### DIFF
--- a/mvrx/src/main/kotlin/com/airbnb/mvrx/MavericksViewModel.kt
+++ b/mvrx/src/main/kotlin/com/airbnb/mvrx/MavericksViewModel.kt
@@ -533,6 +533,9 @@ abstract class MavericksViewModel<S : MvRxState>(
         }
         val scope = lifecycleOwner?.lifecycleScope ?: viewModelScope
         return scope.launch(start = CoroutineStart.UNDISPATCHED) {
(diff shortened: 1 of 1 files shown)


mvrx/src/main/kotlin/com/airbnb/mvrx/MavericksViewModel.kt
```diff
@@ -529,7 +532,8 @@ abstract class MavericksViewModel<S : MvRxState>(
             flowWhenStarted(lifecycleOwner)
         }
         val scope = lifecycleOwner?.lifecycleScope ?: viewModelScope
-        return scope.launch {
+        return scope.launch(start = CoroutineStart.UNDISPATCHED) {
```
gpeal, line 535: @denis-bezrukov Can you add a comment here explaining it like in your pr description?

PR #430, opened 2020-07-30
Fix crash when calling setState/withState after view model is cleared

denis-bezrukov:


mvrx/src/test/kotlin/com/airbnb/mvrx/SetStateWithStateAfterScopeCancellation.kt
```diff
@@ -0,0 +1,28 @@
...
+
+class SetStateWithStateAfterScopeCancellation : BaseTest() {
+    data class State(val foo: Int) : MvRxState
+
+    @Test
+    fun setStateAfterScopeCancellation() {
+        val scope = CoroutineScope(Dispatchers.Default + Job())
+        scope.cancel()
+        val store = CoroutinesStateStore(State(foo = 0), scope)
+        store.set { copy(foo = foo + 1) }
```
gpeal, line 17: Can you assert that the stat is not set to 1?

Issue #339, opened 2020-01-10
Koin viewmodels' injection


Jorikua:
I want to use Koin's extension functions for viewmodels' injections. Is there a way not to write `MvRxViewModelFactory` and use all the advantages of Koin? It's not very convenient to manually inject all of the dependencies into viewmodel through Factory.

dmytroreutov:
Facing same issue. Should be upvoted

gpeal:
@Jorikua @dmytroreutov At Airbnb and Tonal, we use dagger so that is where we have put the most effort. There is a dagger sample module now. Feel free to take a look and see if you can adapt it for koin. Fair warning though, we will probably not have the time to investigate koin integration ourselves though.

However, if changes are necessary to make it work well, we might consider it depending on what is necessary.

PR #432, opened 2020-08-02
Add a sample app to demonstrate integration of Koin with MvRx

friendoye:
Inspired by #339.

This PR adds a new sample app, which showcases usage of Koin with MvRx. In this app custom `KoinMvRxViewModelFactory` was introduced, which delegates creation of specific ViewModel directly to Koin, rather than fetching each constructor parameter manually. Also, this sample, in many ways, is very similar to Dagger sample app.

Finally, this PR addresses such issue as usage of custom Koin scopes during creation of ViewModel in `MvRxViewModelFactory`. Because there might be a situation in Koin, when ViewModel's factory is tied to specific scope, there should be a way for `MvRxViewModelFactory` to know, which scope should be used during creation. That's why `customData` property was introduced in `ViewModelContext`.

Curious to hear WDYT about this PR!


itsandreramon:
How about using KoinComponent?

PR #435, opened 2020-08-13
Add test cases that demonstrates issue with priority between stateFlow and lifecycle flow.

denis-bezrukov:
Here are two tests, I will add commit that fixes them to this PR soon

### DIFF
--- a/mvrx/src/main/kotlin/com/airbnb/mvrx/MvRxLifecycleAwareFlow.kt
+++ b/mvrx/src/main/kotlin/com/airbnb/mvrx/MvRxLifecycleAwareFlow.kt
@@ -17,15 +17,13 @@ import kotlinx.coroutines.flow.onCompletion
  */
 fun <T : Any> Flow<T>.flowWhenStarted(owner: LifecycleOwner): Flow<T> {
     val startedChannel = startedChannel(owner.lifecycle)
-    return onCompletion {
-        startedChannel.cancel()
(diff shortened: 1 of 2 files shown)


mvrx/src/main/kotlin/com/airbnb/mvrx/MvRxLifecycleAwareFlow.kt
```diff
@@ -17,15 +17,13 @@ import kotlinx.coroutines.flow.onCompletion
...
-        startedChannel.cancel()
-    }.combineTransform<T, Boolean, T>(startedChannel.consumeAsFlow()) { value, started ->
+    return startedChannel.consumeAsFlow().combineTransform(this.onCompletion { startedChannel.cancel() }) { started, value ->
         if (started) emit(value)
     }
 }
 
 private fun startedChannel(owner: Lifecycle): Channel<Boolean> {
-    val channel = Channel<Boolean>(Channel.BUFFERED)
+    val channel = Channel<Boolean>(Channel.CONFLATED)
```
denis-bezrukov, line 26: It doesn't affect the fix, but @gpeal and me decided to change it to conflated. Conflated better here for performance reason - we don't need to keep 64 item buffer per collector. @gpeal I can put it to separate PR though

gpeal: You can leave it in this PR :)

PR #439, opened 2020-08-30
[2.0] Renamed more classes from MvRx to Mavericks

gpeal:
This renames most of the remaining MvRx references to Mavericks. There is still more work to do though but I wanted to limit the size of this one PR.

I also did a few cleanups in docs and samples


hellodagger/src/main/java/com/airbnb/mvrx/hellodagger/di/AssistedViewModelFactory.kt
```diff
@@ -19,7 +19,7 @@ import com.airbnb.mvrx.hellodagger.base.BaseViewModel
  * First we define our ViewModel with an @AssistedInject annotated constructor, and a Factory interface
  * implementing AssistedViewModelFactory.
  *
- * class MyViewModel @AssistedInject constructor(...): BaseMvRxViewModel<MyState>(...) {
+ * class MyViewModel @AssistedInject constructor(...): MavericksViewModelViewModel<MyState>(...) {
```
elihart, line 22:
```suggestion
 * class MyViewModel @AssistedInject constructor(...): MavericksViewModel<MyState>(...) {
```

mvrx-rxjava2/src/main/kotlin/com/airbnb/mvrx/MvRx.kt
```diff
@@ -0,0 +1,13 @@
+package com.airbnb.mvrx
+
+/**
+ * Exists for backwards compatibility.
```
elihart, line 4: 👍

gpeal:
@elihart There's MavericksViewModelFactory. and maybe a few there small ones. The package name is a bit more confusing. At first, @BenSchwab and I were hesitant because it would make the upgrade hard but if the classes are all mirrored in rxjava2 such that it wouldn't be a problem then maybe we can. We can try it and see if it's possible to change the core artifact package while retaining backwards compatibility

PR #443, opened 2020-09-06
Add Jetpack navigation module v2

marukami:
Same as #351 except updated for the 2.0 release


mvrx-navigation/src/main/kotlin/com/airbnb/mvrx/navigation/MvRxNavDirections.kt
```diff
@@ -0,0 +1,47 @@
...
+ *       <argument
+ *          android:name="mvrx:arg:title"
+ *          app:argType="string" />
+ *
+ *   </fragment>
+ * ```
+ */
+class MvRxNavDirections(
+    @IdRes private val actionId: Int,
+    private val data: Parcelable,
```
gpeal, line 32: I think you want a version of this for Serializable too

mvrx-navigation/src/main/kotlin/com/airbnb/mvrx/navigation/navigationLifecycleAwareLazy.kt
```diff
@@ -0,0 +1,82 @@
...
+    private var _value: Any? = UninitializedValue
+
+    // final field is required to enable safe publication of constructed instance
+    private val lock = this
+
+    @VisibleForTesting
+    internal val lifecycleObserver: DefaultLifecycleObserver =
+        object : DefaultLifecycleObserver {
+            override fun onStart(owner: LifecycleOwner) {
+                super.onStart(owner)
```
gpeal, line 35: You don't need super here

mvrx-navigation/src/main/kotlin/com/airbnb/mvrx/navigation/navigationLifecycleAwareLazy.kt
```diff
@@ -0,0 +1,82 @@
...
+                try {
+                    if (!isInitialized()) value
+                    owner.lifecycle.removeObserver(this)
+                } catch (cause: IllegalStateException) {
+                    throw IllegalStateException(
+                        """
+                            During device re-configuration or launch after process death the navController is not accessible until onViewCreated.
+                             Try moving any ViewModel access to onViewCreated or later and use subscriptionLifecycleOwner for any subscriptions.
```
gpeal, line 43: Hmm, this is an interesting case. `viewModel.onEach` calls should be done from `onCreate`.
Is that not possible for nav graph scoped view models?

Also, from a semantics standpoint, `subscriptionLifecycleOwner` is not something most people who are using Mavericks will understand. I like that you provided resolution steps but it led to my concern above.

gpeal (commented):
This looks quite good. I think we can land this if we can figure out what to do about the onCreate vs onStart issue.

Could you retarget this on master now that it has merged? I'll try to review this in a timely manner now.

mvrx-navigation/src/main/kotlin/com/airbnb/mvrx/navigation/navigationLifecycleAwareLazy.kt
marukami: > Hmm, this is an interesting case. viewModel.onEach calls should be done from onCreate.
Is that not possible for nav graph scoped view models?

The NavController is not available until the activity has re-attached the NavHost fragment after the Activity.onCreate has setContent and the FragmentManager has re-attached the fragments. So, while this will work when first creating the fragment, the NavController will not be accessible until the Fragments `onCreateView`.

> Also, from a semantics standpoint, subscriptionLifecycleOwner is not something most people who are using Mavericks will understand. I like that you provided resolution steps but it led to my concern above.

I don't think you can fix this case. Nav Graph ViewModels are limited by the accessibility of the BackStack entry and they are only guaranteed to be accessible via NavController in or after `onCreateView/onViewCreared`. We could just omit the last part and suggest moving ViewModel access to that lifecycle point.

```suggestion
                             Try moving any ViewModel access to `onViewCreated`.
```

mvrx-navigation/src/main/kotlin/com/airbnb/mvrx/navigation/MvRxNavDirections.kt
marukami: Ah, yes, I forgot Nav args can be Serializable.

marukami: Updated using a create to pick between `Parcelable` or `Serializable` data types. Let me know what you think.

mvrx-navigation/src/main/kotlin/com/airbnb/mvrx/navigation/MvRxNavDirections.kt
```diff
@@ -21,24 +25,47 @@ import com.airbnb.mvrx.Mavericks
...
     @IdRes private val actionId: Int,
-    private val data: Parcelable,
+    private val parcelable: Parcelable? = null,
+    private val serializable: Serializable? = null,
     private val title: String? = null
 ) : NavDirections {
 
     companion object {
         const val KEY_ARG_TITLE = "mvrx:arg:title"
+        fun create(@IdRes actionId: Int, title: String?, data: Parcelable): MvRxNavDirections =
```
gpeal, line 47: You want one of these with just 2 args right and title can default to null for both create functions.

gpeal (commented):
Just one comment!

This will also need to be re-targeted to master.

Before we land this, can you commit to maintaining this module if there are subsequent bugs, feature requests or underlying navigation library updates?

mvrx-navigation/src/main/kotlin/com/airbnb/mvrx/navigation/MvRxNavDirections.kt
marukami: Ah, yep.

marukami:
> This will also need to be re-targeted to master.

Done

> Before we land this, can you commit to maintaining this module if there are subsequent bugs, feature requests or underlying navigation library updates?

Yes. I can do that.

mvrx-navigation/build.gradle
```diff
@@ -0,0 +1,56 @@
...
+
+    implementation project(":mvrx")
+    implementation project(":testing")
+    implementation Libraries.lifecycleCommon
+
+    testImplementation Libraries.rxAndroid
+    testImplementation TestLibraries.junit
+    testImplementation TestLibraries.mockito
+    testImplementation TestLibraries.roboeletric
+    testAnnotationProcessor "com.google.auto.service:auto-service:1.0-rc4"
```
gpeal, line 38: Could you update these and the other dependencies to use the Libraries and TestLibraries objects?

mvrx-navigation/src/main/kotlin/com/airbnb/mvrx/navigation/navigationLifecycleAwareLazy.kt
```diff
@@ -0,0 +1,94 @@
...
+
+    @Volatile
+    @SuppressWarnings("Detekt.VariableNaming")
+    private var _value: Any? = UninitializedValue
+
+    // final field is required to enable safe publication of constructed instance
+    private val lock = this
+
+    @VisibleForTesting
+    internal val lifecycleObserver: DefaultLifecycleObserver =
```
gpeal, line 32: You changed this from the original impl and marked it as `@VisibleForTesting` but I don't see any tests for it. If you aren't adding new tests, can you keep the impl as close as possible so it can be more easily diffed?

mvrx-navigation/src/test/kotlin/com/airbnb/mvrx/navigation/HostFragment.kt
```diff
@@ -0,0 +1,24 @@
...
+
+class HostFragment : Fragment(R.layout.mvrx_fragment_host) {
+
+    companion object {
+        var accessViewModelInOnCreate = false
+    }
+
+    var viewModel: NavigationViewModel? = null
+    override fun onViewCreated(view: View, savedInstanceState: Bundle?) {
+        super.onViewCreated(view, savedInstanceState)
```
gpeal, line 18: Don't need super here

gpeal (commented):
This is looking great. Final round of comments I think!

mvrx-navigation/src/main/kotlin/com/airbnb/mvrx/navigation/navigationLifecycleAwareLazy.kt
marukami: Ah, yer. If there were any tests I've since removed them. We can match the original impl more here.

mvrx-navigation/build.gradle
```diff
@@ -0,0 +1,65 @@
...
+    implementation Libraries.lifecycleCommon
+
+    testImplementation Libraries.rxAndroid
+    testImplementation TestLibraries.junit
+    testImplementation TestLibraries.mockito
+    testImplementation TestLibraries.roboeletric
+
+    // module specific dependencies that don't need to be shared via a common dependencies file.
+    def nav_version = "2.3.0"
+    api "androidx.navigation:navigation-fragment-ktx:$nav_version"
```
gpeal, line 51: Great! Just update these and the PR is good to go

gpeal (approved):

gpeal:
@marukami Looks great! If you fix the merge conflict, I'll land this. Thanks for all of the hard work you've put into this.

marukami:
@gpeal I fixed the conflict and updated the changelog.

gpeal:
@marukami Thanks again for your work here. In the future, I recommend merging master and adding commits. If you rebase or force push, GitHub replays the entire commit history again and if you look at the PR history, it's impossible to see the subset of changes you made since the last push and now the entire PR has to be reviewed again.

PR #444, opened 2020-09-09
Improve performance of lifecycle aware flow

denis-bezrukov:
we noticed a performance regression when main thread is busy
after reviewing `combineTransform` implementation I found that it converts flows to _fair channels_. Fair channel uses `yield` each time in fast-path offer, that means it dispatches coroutine per each element. 

In this PR I rewrote `flowWhenStarted` by forking combineTransform implementation and adopting it for us. It includes two changes: 
1. use non-fair channels
2. end flow if startedFlow/vm.stateflow is ended (combineTransform waited for completion of both flows, so we used `this.onCompletion { startedChannel.cancel() }`)

I also want to cover this with more tests for
a) prevent this regression in future
b) ensure nothing was broken due change to non-fair channel
But operator is ready for review.

### DIFF
--- a/mvrx/src/main/kotlin/com/airbnb/mvrx/MavericksLifecycleAwareFlow.kt
+++ b/mvrx/src/main/kotlin/com/airbnb/mvrx/MavericksLifecycleAwareFlow.kt
@@ -6,19 +6,51 @@ import androidx.lifecycle.DefaultLifecycleObserver
 import androidx.lifecycle.Lifecycle
 import androidx.lifecycle.LifecycleOwner
 import kotlinx.coroutines.channels.Channel
+import kotlinx.coroutines.channels.ReceiveChannel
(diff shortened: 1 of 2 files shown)


gpeal (commented):
Operator LGTM! Let's see if we can add those tests! Thanks @denis-bezrukov!

elihart (approved):
Thanks for improving this. It seems good, but I think it would be helpful to put your pr summary details into this code so that we have a record of why it is customized like this

gpeal (approved):

elihart:
@denis-bezrukov @gpeal I think this change broke two tests (`ViewModelSubscriberTest$testCancelledIfViewModelCleared`). I see this failing in my PR https://github.com/airbnb/MvRx/pull/447 and tried this commit and the one before it and it seems to have started with this commit. Also checks don't seem to have run on this PR

PR #449, opened 2020-09-16
Add on demand execute function on IO thread

rizmaulana:
It will reduce boilerplate to add manually execute on IO thread when doing heavy task like network call

From :
```
service.getUser()
      .subscribeOn(Schedulers.io())
      .execute{
          copy(user = it)
       }
```

To : 
```
service.getUser()
      .executeOnIo{
          copy(user = it)
       }
```


gpeal:
@rizmaulana This is an easy enough extension to add to your own app. I would rather keep the API surface area of MvRx as small as possible.

Minirogue:
What about an optional scheduler argument for the existing `execute` functions? It wouldn't increase the number of functions in the api. The same could be useful for coroutine dispatchers (although, I guess coroutine contexts would be the correct kind of object) in 2.0.

gpeal:
@Minirogue The coroutines extensions do have an optional Dispatchers param actually! However, it's more important for coroutines because subscribeOn isn't an operator that you can put inline the way it is for Rx.

rizmaulana:
Got it @gpeal 

But what about add `Schedulers.io()` as default scheduler? I think most of observable job is like doing http connection or writing to database, then developers don't need add `.subscribeOn(Schedulers.io())` manually. When developers want to force observable on mainThread, they can add `.subscribeOn(AndroidSchedulers.mainThread())` manually

gpeal:
@rizmaulana 
@elihart  We talked about this a bit for 2.0. Some considerations:
1) We don't want to change the default for existing users. That will have unintended consequences.
2) IO is _probably_ is what we want most of the time but it's also presumptuous
3) It might be odd to have the rxjava interface not specify a scheduler but have the coroutines one default to IO

rizmaulana:
@gpeal

1. Got it
2. Why?
3. It's also strange behaviour at the first time RxJava without `subscribe`, but now I love `execute` very much.

gpeal:
@rizmaulana Because IO has a specific intention whereas execute is generic and can be used for anything so it defaults to not explicitly changing the dispatcher.

rizmaulana:
@gpeal Ok noted, I got your point.
Thank you for the discussion, I'll close this PR

PR #453, opened 2020-09-20
[2.0] Cleaned up and obfuscated some @RestrictTo APIs

gpeal:
One quality of life regression from 1.0 is that intellisense was recommending lots of `onAsyncInternal` functions all the time. This partly because `@RestrictTo` simply [doesn't do anything](https://issuetracker.google.com/issues/168357308). To improve this, I moved them to extension functions and obfuscated their names. I was also able to reduce the visibility of a few other `@RestrictTo` APIs.

I'm open to other ideas as long as we can prevent onAsyncInternal from showing up just like any other function.

@denis-bezrukov


denis-bezrukov:
could it be `internal` ? or is it used it in other artifacts like mvrx-rxjava ? can we use public api from rxjava artifact? (like use onEach for selectSubscribe instead of onEachInternal)

mvrx/src/main/kotlin/com/airbnb/mvrx/MavericksFactory.kt
```diff
@@ -4,8 +4,7 @@ import androidx.annotation.RestrictTo
 import androidx.lifecycle.ViewModel
 import androidx.lifecycle.ViewModelProvider
 
-@RestrictTo(RestrictTo.Scope.LIBRARY)
-class MavericksFactory<VM : MavericksViewModel<S>, S : MavericksState>(
+internal class MavericksFactory<VM : MavericksViewModel<S>, S : MavericksState>(
```
elihart, line 7: we had been using this at airbnb. we may not need to any longer with the upstreamed mocking, but i'd rather leave this public for now just in case we need it. this doesn't seem like one that will accidentally be used

elihart (approved):
Thanks for looking into this, it seems fine to me. not sure we have another choice.

I think `RestrictTo` only causes lint failures

gpeal:
@denis-bezrukov We don't expose the LifecycleOwner version of onEach publicly and I think it's not necessary to because of stateFlow.

@elihart The lint rule also seems to not actually work (see the issue I linked)...

mvrx/src/main/kotlin/com/airbnb/mvrx/MavericksFactory.kt
gpeal: @elihart It's less about using it and more about it showing up every tine you type Mavericks in intellisense for things like your ViewModel/State/etc.

gpeal: For that reason, I want to keep the `Mavericks*` public class namespace as small as possible.

Issue #389, opened 2020-05-12
[2.0] Investigate setting a StrictMode thread policy to MvRx thread


gpeal:
Related to https://github.com/airbnb/MvRx/pull/386 and could let us crash in debug when people do things like network and disk access from a MvRx thread. Good idea @denis-bezrukov

cc @elihart @BenSchwab

PR #454, opened 2020-09-20
StateStore StrictMode prototype

denis-bezrukov:
this is an implementation of #389 

Added an extension that converts `ThreadPolicy` to a `CoroutineContext.Element` so then it can be simply injected to any coroutine scope/context. Implementation is similar to `ThreadLocalElement` from coroutines lib

Also added storeContextOverride to ConfigFactory so ThreadPolicy can be injected via `Maverics.initialize`:
```
val policy = StrictMode.ThreadPolicy.Builder()
    .detectNetwork()
    .penaltyDialog()
    .build()
Mavericks.initialize(this, viewModelConfigFactory = MavericksViewModelConfigFactory(
    context = this,
    storeContextOverride = if (isDebuggable()) policy.asContextElement() else EmptyCoroutineContext
))
```

it also means users will be able to provide their own dispatcher for state store, like Dispatchers.Default (we considered using it by default instead of cached executor).

ThreadPolicy has a lot of options, and now users can decide what to detect (network, disk reads, disk writes, slow operations) and how to react on violation (death, log, dialog, screen flash etc). I don't think we need to enforce certain policy. Example of detectNetwork + penaltyDialog: 
image

### DIFF
--- a/docs/debug-checks.md
+++ b/docs/debug-checks.md
@@ -7,3 +7,17 @@ MvRx has a debug mode that enables some validation checks. When you integrate Mv
 * State types are not one of: ArrayList, SparseArray, LongSparseArray, SparseArrayCompat, ArrayMap, and HashMap.
 * State class visibility is public so it can be created and restored
 * Each instance of your State is not mutated once it is set on the viewmodel
+
+### Avoid slow operations in `withState`/`setState`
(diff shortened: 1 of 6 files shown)


gpeal (commented):
I really like this idea and implementation.
@elihart @BenSchwab What do you think?

gpeal:
@denis-bezrukov If we move forward with this, we'll want to add docs in the docs/ folder as well.

elihart (approved):
looks great! thanks for this, good idea. I agree with @gpeal that just some documentation before merging would be nice. Would likely also be good to update the wiki docs in this same PR as well

gpeal (approved):

PR #461, opened 2020-10-09
Tweaks and additions to the mocking framework

elihart:
I've completed Airbnb's migration to mavericks 2.0, which included moving us to the now open source mocking framework. Along the way I encountered a few things that needed to be changed.

- parentFragmentViewModel needed to be able to correct be mocked out when a parent fragment isn't present, which was as simple as throwing the right exception that the mocking framework already looks for
- Added support for registering onExecute listeners for viewmodels
- fixed issues with not being able to provide a custom empty mocks object
- fixed some generics
- updated the mockable view model config factory to support the new coroutine context overrides


mvrx/src/main/kotlin/com/airbnb/mvrx/MavericksStateFactory.kt
```diff
@@ -77,21 +78,29 @@ internal fun <VM : MavericksViewModel<S>, S : MavericksState> createStateFromCon
...
+                    stateClass.constructors.lastOrNull()
+                }?.let { c ->
                     c.isAccessible = true
                     c.newInstance() as? S
                 }
             }
         } catch (@Suppress("TooGenericExceptionCaught") e: Throwable) {
             // Throw this exception if something we explicitly tried failed.
-            throw java.lang.IllegalStateException("Failed to create initial state!", e)
+            throw IllegalStateException("Failed to create initial state!", e)
```
gpeal, line 94: `error(...)`?

gpeal (approved):
Thanks for doing this! I'm very excited about how this is coming together!

mvrx/src/main/kotlin/com/airbnb/mvrx/MavericksStateFactory.kt
elihart: ah, can't use that with a nested exception actually

PR #464, opened 2020-10-14
Coroutine api tweaks

elihart:
Makes a few changes to how coroutines are executed in view models, as well as the public API for controlling coroutines:

- MockableMavericks initialization now accepts the context override parameters for easy configuration. We may want to add this to `Mavericks.initialize` as well
- coroutine `execute` functions in MavericksViewModel now have a null dispatcher default instead of Dispatchers.main.immediate. When null, the plain view model scope is used
- for viewmodel subscription functions, when the subscriber's lifecycle is passed and the lifecycle coroutine scope is used to launch the coroutine, the global coroutine context override is first added to that scope
- Flow.execute had a `catch` statement added, which catches exceptions and reduces them with `Fail`, to make it consistent with other execute functions

Overall, this makes it so that the coroutine context override applied in Mavericks initialization is used in all viewmodel coroutine uses so we can have a consistent, flexible global default. Individual executions can still specify a custom dispatchers.

### DIFF
--- a/gradle/wrapper/gradle-wrapper.properties
+++ b/gradle/wrapper/gradle-wrapper.properties
@@ -1,5 +1,5 @@
 distributionBase=GRADLE_USER_HOME
 distributionPath=wrapper/dists
-distributionUrl=https\://services.gradle.org/distributions/gradle-6.7-bin.zip
+distributionUrl=https\://services.gradle.org/distributions/gradle-6.7-all.zip
 zipStoreBase=GRADLE_USER_HOME
 zipStorePath=wrapper/dists
--- a/gradle/wrapper/gradle-wrapper.properties
+++ b/gradle/wrapper/gradle-wrapper.properties
@@ -1,6 +1,5 @@
-#Sun Oct 11 14:29:09 CDT 2020
 distributionBase=GRADLE_USER_HOME
 distributionPath=wrapper/dists
+distributionUrl=https\://services.gradle.org/distributions/gradle-6.7-bin.zip
 zipStoreBase=GRADLE_USER_HOME
 zipStorePath=wrapper/dists
-distributionUrl=https\://services.gradle.org/distributions/gradle-6.7-rc-5-all.zip
--- a/gradle/wrapper/gradle-wrapper.properties
+++ b/gradle/wrapper/gradle-wrapper.properties
@@ -1,5 +1,6 @@
+#Sun Oct 11 14:29:09 CDT 2020
 distributionBase=GRADLE_USER_HOME
 distributionPath=wrapper/dists
-distributionUrl=https\://services.gradle.org/distributions/gradle-6.7-rc-5-all.zip
 zipStoreBase=GRADLE_USER_HOME
 zipStorePath=wrapper/dists
+distributionUrl=https\://services.gradle.org/distributions/gradle-6.7-rc-5-all.zip
--- a/mvrx-mocking/src/main/kotlin/com/airbnb/mvrx/mocking/MockableMavericks.kt
+++ b/mvrx-mocking/src/main/kotlin/com/airbnb/mvrx/mocking/MockableMavericks.kt
@@ -2,12 +2,14 @@ package com.airbnb.mvrx.mocking
 
 import android.content.Context
 import android.content.pm.ApplicationInfo
+import androidx.lifecycle.LifecycleOwner
 import com.airbnb.mvrx.CoroutinesStateStore
 import com.airbnb.mvrx.DefaultViewModelDelegateFactory
 import com.airbnb.mvrx.Mavericks
 import com.airbnb.mvrx.MavericksViewModel
 import com.airbnb.mvrx.MavericksState
 import com.airbnb.mvrx.MavericksStateStore
+import com.airbnb.mvrx.MavericksView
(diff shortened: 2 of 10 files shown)


gradle.properties
```diff
@@ -1,4 +1,4 @@
-VERSION_NAME=2.0.0-beta2
+VERSION_NAME=2.0.0-airbnb14
```
gpeal, line 1: You're going to want to revert this before landing :)

gpeal (commented):
@elihart Is this just to set the default for execute as IO? If so, lets revisit the discussion to make it the default. I think there is a compelling case for it to be default.

mvrx/src/main/kotlin/com/airbnb/mvrx/MavericksViewModel.kt
```diff
@@ -399,7 +416,8 @@ abstract class MavericksViewModel<S : MavericksState>(
         } else {
             flowWhenStarted(lifecycleOwner)
         }
-        val scope = lifecycleOwner?.lifecycleScope ?: viewModelScope
+
+        val scope = lifecycleOwner?.lifecycleScope?.let { it + configFactory.subscriptionCoroutineContextOverride } ?: viewModelScope
```
denis-bezrukov, line 420: nit, probably it looks slightly better 
```suggestion
        val scope = lifecycleOwner?.lifecycleScope?.plus(configFactory.subscriptionCoroutineContextOverride) ?: viewModelScope
```

btw don't you need to add it to case when viewModelScope is used? (MavericksView is only an example in the subscriptionCoroutineContextOverride doc: 
>Provide a context that will be added to the coroutine scope when a subscription is registered

)

mvrx/src/main/kotlin/com/airbnb/mvrx/MavericksViewModel.kt
```diff
@@ -236,25 +244,34 @@ abstract class MavericksViewModel<S : MavericksState>(
...
 
         setState { reducer(Loading()) }
-        return onEach {
-            setState { reducer(Success(it)) }
-        }.launchIn(viewModelScope + dispatcher)
+
+        return catch { error -> setState { reducer(Fail(error)) } }
+            .onEach {
+                setState { reducer(Success(it)) }
+            }.launchIn(viewModelScope + (dispatcher ?: EmptyCoroutineContext))
```
denis-bezrukov, line 251: can these two lines be consistent:
```suggestion
        return catch { setState { reducer(Fail(it)) } }
            .onEach { setState { reducer(Success(it)) } }
            .launchIn(viewModelScope + (dispatcher ?: EmptyCoroutineContext))
```

or 
```suggestion
        return catch { error -> setState { reducer(Fail(error)) } }
            .onEach { value -> setState { reducer(Success(value)) } }
            .launchIn(viewModelScope + (dispatcher ?: EmptyCoroutineContext))
```

mvrx/src/main/kotlin/com/airbnb/mvrx/MavericksViewModel.kt
elihart: nice syntax idea. 

good catch, I suppose we should add it in both cases. Previously it didn't matter since it was the same context, but that changed once I added a dedicated `subscriptionCoroutineContextOverride`

mvrx-mocking/src/main/kotlin/com/airbnb/mvrx/mocking/MockableMavericks.kt
```diff
@@ -109,22 +117,64 @@ object MockableMavericks {
...
+         * Application context. If provided this will be used to register a [ViewModelStatePrinter] for each ViewModel to support mock state printing.
+         */
+        applicationContext: Context?,
+        /**
+         * Provide a default coroutine context for viewModelScope in all view models. It will be added after [SupervisorJob]
+         * and [Dispatchers.Main.immediate].
+         */
+        viewModelCoroutineContext: CoroutineContext = EmptyCoroutineContext,
+        /**
+         * Provide a coroutine context that will be used in the [CoroutinesStateStore]. All withState/setState calls will be executed in this context.
```
gpeal, line 140: Can you indicate what the defaults are here? It looks like it defaults to EmptyCoroutineContext but in reality, it has a sane default but this just lets you override it.

elihart: sure

gpeal (approved):

gradle/wrapper/gradle-wrapper.properties
```diff
@@ -1,5 +1,5 @@
 distributionBase=GRADLE_USER_HOME
 distributionPath=wrapper/dists
-distributionUrl=https\://services.gradle.org/distributions/gradle-6.7-rc-5-all.zip
+distributionUrl=https\://services.gradle.org/distributions/gradle-6.7-bin.zip
```
denis-bezrukov, line 3: nit, you probably need to use -all:
```suggestion
distributionUrl=https\://services.gradle.org/distributions/gradle-6.7-all.zip
```
I don't know why cmd updater switches to -bin for 6.7..

elihart: thanks!

PR #468, opened 2020-10-26
Launcher and mock speed optimizations

elihart:
When the launcher includes many different mavericks views and the mocks for the views are complex it can take a long time to load them all. instead, we really just need to initially get the list of view classes and can defer loading the mocks until a view is selected.

I have also updated the class loading to use a flow for incremental loading and slightly optimized the reflection. It will also now include views where mocks have not been added in the list, and when clicked will show a message saying they have not added mocks yet.

I have also bumped coroutines to latest 1.4.0 where StateFlow is stable

lastly, i've updated the docs for a dedicated page about the launcher, which includes details about how to open it from command line


docs/launcher.md
```diff
@@ -0,0 +1,107 @@
...
+
+## Caveats
+Since each mocked fragment is launched in isolation it must be able to function independently, without assuming any other fragments or activities are present. For example, it must not make assumptions about what its parent fragment or parent activity is.
+
+This means that if you need inter-fragment communication you cannot assume another fragment exists, but instead must use decoupled approaches, such as a Dagger injected class.
+
+Note that the same restriction applies for fragments being run in automatic test suites, so all Mavericks views should follow this.
+
+## How It Works
+Mavericks inspects the app Dex files to find all mockable Mavericks views and their declared mock states and arguments. This process can take some time, so Fragment and mock information is saved in shared preferences for faster loading on subsequent app starts.
```
BenSchwab, line 101: Is this worth an investigation to datastore down the road? If so maybe remove the specific technology we use the store the data

BenSchwab (approved):

gpeal (approved):

PR #469, opened 2020-10-27
Migrated CoroutinesStateStore to SharedFlow

gpeal:
SharedFlow is great and is perfect for the use case we need to achieve in the coroutines store.

@denis-bezrukov


mvrx/src/test/kotlin/com/airbnb/mvrx/CoroutineStateStoreReplayTest.kt
```diff
@@ -23,18 +23,14 @@ class CoroutineStateStoreReplayTest {
 
     @Test
     fun replayTest() = runBlocking {
-        repeat(100) {
-            singleReplayTestIteration(N = 5000, subscribers = 10)
+        repeat(1000) {
+            // The store's backing SharedFlow has a buffer of 64 elements per subscriber. After that, it will drop
+            // the oldest messages.
+            singleReplayTestIteration(N = 64, subscribers = 10)
```
elihart, line 29: potentially put 64 into an internal const to reuse the value with the state store implementation

elihart (approved):
Very cool, love the improved simplicity. thanks for jumping on this!

mvrx/src/main/kotlin/com/airbnb/mvrx/CoroutinesStateStore.kt
```diff
@@ -28,39 +24,24 @@ class CoroutinesStateStore<S : MavericksState>(
...
-        get() = flow {
-            val (initialState, subscription) = updateMutex.withLock {
-                state to stateChannel.openSubscription()
-            }
-            subscription.consume {
-                emit(initialState)
-                emitAll(this)
-            }
-        }
+    override val flow: Flow<S> = stateSharedFlow.asSharedFlow()
```
denis-bezrukov, line 44: What do you think about exposing SharedFlow instead of Flow?

denis-bezrukov (commented):
are you planing to update `ScriptableMavericksStateStore` and `SynchronousMavericksStateStore` stores as well?

**Behavior change:**
I have a concerns about intoducing this behavior change, because it may affect not only fast and slow collectors, but also collectors launched in the single-threaded dispatcher (which is specifically true for Android with Main thread dispatcher). In single-threaded dispatcher collectors may be fast, but values will still drop because single collector may drain the buffer in a single call frame. It will be hard to debug and reproduce such issues.

And the second reason why I advocate for `BufferOverflow.SUSPEND` is that slow-consumers can easily turn on this flow into drop-oldest strategy ( using `.buffer(onBufferOverflow = BufferOverflow.DROP_OLDEST)` ). But when you have a store with a flow that uses DROP_OLDEST strategy, you can't switch to SUSPEND strategy as a collector.

mvrx/src/main/kotlin/com/airbnb/mvrx/CoroutinesStateStore.kt
```diff
@@ -135,5 +113,11 @@ class CoroutinesStateStore<S : MavericksState>(
 
     companion object {
         private val flushDispatcher = Executors.newCachedThreadPool().asCoroutineDispatcher()
+
+        /**
+         * The buffer size that will be allocated by [MutableSharedFlow] for each subscriber.
```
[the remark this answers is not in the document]
gpeal: @denis-bezrukov Reading the code, you're right. I misunderstood the way it worked. Thanks!

mvrx/src/main/kotlin/com/airbnb/mvrx/CoroutinesStateStore.kt
gpeal: @denis-bezrukov What purpose would it be used for?

mvrx/src/main/kotlin/com/airbnb/mvrx/CoroutinesStateStore.kt
```diff
@@ -28,39 +24,24 @@ class CoroutinesStateStore<S : MavericksState>(
...
-    private val stateChannel = BroadcastChannel<S>(capacity = Channel.BUFFERED)
-    private val updateMutex = Mutex()
-    override var state = initialState
+    private val stateSharedFlow = MutableSharedFlow<S>(
+        replay = 1,
+        extraBufferCapacity = SubscriberBufferSize,
+        onBufferOverflow = BufferOverflow.DROP_OLDEST,
+    ).apply { tryEmit(initialState) }
+
+    override val state: S get() = stateSharedFlow.replayCache.first()
```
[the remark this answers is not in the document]
gpeal: @denis-bezrukov Good points! I don't think we need the lock anymore though right? It was just used to synchronize the our previous SharedFlow impl.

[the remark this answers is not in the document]
denis-bezrukov: yes, we don't need it. 
Probably it makes sense to mark state `@Transient` though

mvrx/src/main/kotlin/com/airbnb/mvrx/CoroutinesStateStore.kt
denis-bezrukov: idk honestly. Initially I though about compose usecase - to synchronously retrieve the current state and then subscribe to updates. But it looks it's pretty similar to withState(vm) { it } + flow

mvrx-mocking/src/main/kotlin/com/airbnb/mvrx/mocking/SynchronousMavericksStateStore.kt
```diff
@@ -2,33 +2,26 @@ package com.airbnb.mvrx.mocking
...
+        extraBufferCapacity = 64,
+        onBufferOverflow = BufferOverflow.SUSPEND,
+    ).apply { tryEmit(initialState) }
 
-    init {
-        coroutineScope.coroutineContext[Job]!!.invokeOnCompletion {
-            stateChannel.cancel()
-        }
-    }
+    override val flow: Flow<S> = stateSharedFlow.asSharedFlow()
```
denis-bezrukov, line 24: in this store it makes sense to add `.distinctUntilChanged()` because same values aren't dropped on reducer level. The same for Scriptable store

mvrx/src/main/kotlin/com/airbnb/mvrx/ScriptableMavericksStateStore.kt
```diff
@@ -17,14 +15,16 @@ import kotlinx.coroutines.flow.flow
 @Suppress("EXPERIMENTAL_API_USAGE")
```
denis-bezrukov, line 15: nit: shouldn't be necessary anymore

denis-bezrukov (approved):
LGTM 🚀

mvrx-rxjava2/src/test/kotlin/com/airbnb/mvrx/ViewModelSubscriberTest.kt
```diff
@@ -781,13 +781,6 @@ class ViewModelSubscriberTest : BaseTest() {
         assertTrue(job.isCompleted)
     }
 
-    @Test
-    fun testCancelledIfViewModelCleared() = runBlocking {
```
gpeal, line 785: @elihart @denis-bezrukov I had to remove this test. I'm not sure why it still passed after the change to not cancel the store's state flow but it seems like this isn't the expected behavior anymore right?

denis-bezrukov: @gpeal  It passed because this test under the hood used another store  (sync state store via mockable state store. And sync store used to close the channel in invokeOnCancel).

Yep, it is not expected anymore

PR #478, opened 2020-11-15
Mark all @RestrictTo APIs with @InternalMavericksApi

### LINKED ISSUE #474: Use @OptIn for internal Mavericks functions
This will help us avoid some of the ugly naming we use for the MvRx module to communicate with the RxJava module.
https://zsmb.co/mastering-api-visibility-in-kotlin/

gpeal:
`@RestrictTo` doesn't really work so this approach will be much more aggressive about failing builds where internal APIs are used.
Fixes #474 

@denis-bezrukov

### DIFF
--- a/launcher/build.gradle
+++ b/launcher/build.gradle
@@ -1,10 +1,19 @@
 import com.airbnb.mvrx.JacocoReportTask
+import org.jetbrains.kotlin.gradle.tasks.KotlinCompile
 
 apply plugin: 'com.android.library'
 apply plugin: 'kotlin-android'
 apply plugin: 'kotlin-kapt'
 apply plugin: "jacoco"
 apply plugin: "com.vanniktech.maven.publish"
+
+tasks.withType(KotlinCompile).all {
+    kotlinOptions {
+        freeCompilerArgs += [
+                '-Xopt-in=com.airbnb.mvrx.InternalMavericksApi',
+        ]
+    }
+}
 
 android {
     resourcePrefix 'mavericks_'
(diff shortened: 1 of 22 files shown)


mvrx-mocking/build.gradle
```diff
@@ -7,6 +8,14 @@ apply plugin: "jacoco"
 apply plugin: "com.vanniktech.maven.publish"
 apply plugin: 'kotlin-android-extensions'
 
+tasks.withType(KotlinCompile).all {
```
denis-bezrukov, line 11: nit: for configuration avoidance:
```suggestion
tasks.withType(KotlinCompile).configureEach {
```
same for other `.all` usages

denis-bezrukov (approved):

launcher/build.gradle
```diff
@@ -1,11 +1,20 @@
 import com.airbnb.mvrx.JacocoReportTask
+import org.jetbrains.kotlin.gradle.tasks.KotlinCompile
 
 apply plugin: 'com.android.library'
 apply plugin: 'kotlin-android'
 apply plugin: 'kotlin-kapt'
 apply plugin: "jacoco"
 apply plugin: "com.vanniktech.maven.publish"
 
+tasks.withType(KotlinCompile).all {
```
elihart, line 10: I think this can be done in the `AndroidMetaModulePlugin` (which is applied to all projects already) to avoid duplication

elihart (approved):
@gpeal great idea, just had one suggestion to remove duplication

launcher/build.gradle
gpeal: @elihart We don't want this applied to any of the other sample modules.

PR #487, opened 2020-11-26
Add MavericksViewModel.awaitState

gpeal:
Fixes https://github.com/airbnb/MvRx/issues/484

One thing I noticed is that execute and setOnEach is a public API and I think it should be protected. I can change the visibility in a subsequent PR unless you can think of a valid use case.

@denis-bezrukov


mvrx/src/main/kotlin/com/airbnb/mvrx/MavericksViewModel.kt
```diff
@@ -158,6 +159,16 @@ abstract class MavericksViewModel<S : MavericksState>(
         }
     }
 
+    /**
+     * Calling this function suspends until all pending setState reducers are run and then returns the latest state.
+     * As a result, it is safe to call setState { } and assume that the result from a subsequent awaitState() call will have that state.
+     */
+    suspend fun awaitState(): S {
```
elihart, line 166: nice implementation 👍

elihart (approved):
Thanks, this is great!

agreed it looks like execute and setOnEach can be protected

PR #420, opened 2020-06-18
Draft - Adds Hilt example.

carlosmuvi:
Creates a `HelloHilt` app module replicating the existing `HelloDagger` example.

## Differences vs Vanilla Dagger sample

- Application now includes `@HiltAndroidApp`.
- AppComponent has been removed, and existing modules are now `@Install`ed in Hilt's ApplicationComponent.
- An `@EntryPoint` replaces `AppComponent#viewModelFactories`.

## Limitations

Hilt doesn't support @AssistedInject yet. Hilt checks that included modules also have the `@InstallIn` annotation. Temporarily we'd have to disable that check.

(see https://dagger.dev/hilt/compiler-options for more detail).

PR #495, opened 2021-01-07
Adds Hilt Example (Dagger @AssistedInject edition)

marukami:
Dagger Hilt snapshot release added `@AssistedInject`. So, I rebased the branch from @carlosmuvi #420 onto master and updated dagger to use the snapshot `@AssistedInject`. 

As per #420 PR the new module adds an Hilt app module.

# Differences vs Vanilla Dagger sample

The application now includes @HiltAndroidApp.
AppComponent has been removed, and existing modules are now installed in Hilt's `SingletonComponent`.
An `@EntryPoint` replaces AppComponent#viewModelFactories.


hellohilt/README.md
```diff
@@ -0,0 +1,91 @@
...
+    }
+    dependencies {
+        classpath "com.google.dagger:hilt-android-gradle-plugin:HEAD-SNAPSHOT"
+        // rest of classpath(s)…
+    }
+}
+
+// module/build.gradle
+```groovy
+repositories {
```
gpeal, line 23: I believe putting this in the top level build.gradle file like:
```
allprojects {
    repositories {
        ...
    }
}
```
would be more standard

hellohilt/README.md
```diff
@@ -0,0 +1,91 @@
...
+    kapt "com.google.dagger:hilt-android-compiler:${hiltVersion}"
+    implementation "com.google.dagger:hilt-android:${hiltVersion}"
+}
+```
+
+## Key Features
+
+* **Injecting state into ViewModels with AssistedInject**
+  
+  Since the `initialState` parameter is only available at runtime, Dagger can not provide this dependency for us. We need the [AssistedInject](https://github.com/square/AssistedInject) library for this purpose.
```
gpeal, line 38: You can update this doc :)

hellohilt/README.md
```diff
@@ -0,0 +1,91 @@
...
+  companion object: HiltMavericksViewModelFactory<MyViewModel, MyState>(MyViewModel::class.java)
+}
+```
+
+* Tell Dagger to include your ViewModel's AssistedInject Factory in a Multibinding map.
+
+```kotlin
+interface AppModule {
+
+    @[Binds IntoMap ViewModelKey(MyViewModel::class)]
```
gpeal, line 68: Interesting syntax. I've never seen the annotation array like this before! Good to know it's possible but I think it'll be easier for people to understand if you split it into 3 lines because that syntax is more common.

hellohilt/README.md
```diff
@@ -0,0 +1,91 @@
...
+
+```kotlin
+class MyFragment : BaseMvRxFragment() {
+  val viewModel: MyViewModel by fragmentViewModel()
+}
+```
+
+## How it works
+
+`HiltMavericksViewModelFactory` will try loading the custom entry point from the `SingletonComponent`
```
gpeal, line 90: What if you wanted this to be scoped to a different component? This should theoretically be able to be scoped by Fragment, Activity, ActivityRetained, or Singleton right?

hellohilt/build.gradle
```diff
@@ -0,0 +1,51 @@
+apply plugin: "com.android.application"
+apply plugin: "kotlin-android"
+apply plugin: "kotlin-android-extensions"
```
gpeal, line 3: kotlin-android-extensions is deprecated now. Let's move to the parcelize plugin for all new things and we can remove it from the rest soon.

hellohilt/build.gradle
```diff
@@ -0,0 +1,51 @@
...
+apply plugin: 'dagger.hilt.android.plugin'
+apply plugin: "kotlin-kapt"
```
gpeal, line 5: nit: can you put the kotlin plugins together?

hellohilt/build.gradle
```diff
@@ -0,0 +1,51 @@
...
+
+repositories {
```
gpeal, line 7: Per my comment above, let's move this to the top level build.gradle

hellohilt/build.gradle
```diff
@@ -0,0 +1,51 @@
...
+    implementation Libraries.hilt
+    implementation Libraries.rxJava
+    implementation Libraries.viewModelKtx
+    implementation Libraries.multidex
+    implementation project(":mvrx-rxjava2")
+    implementation project(":mvrx-mocking")
+
+    debugImplementation Libraries.fragmentTesting
+
+    androidTestImplementation InstrumentedTestLibraries.core
```
gpeal, line 44: I don't think you use these

hellohilt/src/main/java/com/airbnb/mvrx/hellohilt/HelloFragment.kt
```diff
@@ -0,0 +1,34 @@
...
+import kotlinx.android.synthetic.main.fragment_hello.helloButton
+import kotlinx.android.synthetic.main.fragment_hello.messageTextView
+
+@AndroidEntryPoint
+class HelloFragment : Fragment(R.layout.fragment_hello), MvRxView {
+
+    val viewModel: HelloViewModel by fragmentViewModel()
+
+    override fun onViewCreated(view: View, savedInstanceState: Bundle?) {
+        helloButton.setOnClickListener { viewModel.sayHello() }
```
gpeal, line 23: Can you migrate this to view binding like the sample app? You can copy in [FragmentViewBindingDelegate](https://github.com/airbnb/MvRx/blob/master/sample/src/main/java/com/airbnb/mvrx/sample/utils/FragmentViewBindingDelegate.kt)

hellohilt/src/main/java/com/airbnb/mvrx/hellohilt/HelloRepository.kt
```diff
@@ -0,0 +1,14 @@
...
+import java.util.concurrent.TimeUnit
+import javax.inject.Inject
+
+class HelloRepository @Inject constructor() {
+
+    fun sayHello(): Observable<String> {
+        return Observable
+            .just("Hello, world!")
+            .delay(2, TimeUnit.SECONDS)
+    }
```
gpeal, line 13: Can you make this coroutines since most new things will be using that? It can be:
```suggestion
    suspend fun sayHello()String {
        delay(2_000)
        return "Hello World"
    }
```

hellohilt/src/main/java/com/airbnb/mvrx/hellohilt/HelloViewModel.kt
```diff
@@ -0,0 +1,34 @@
...
+    @Assisted state: HelloState,
+    private val repo: HelloRepository
+) : BaseMvRxViewModel<HelloState>(state) {
+
+    init {
+        sayHello()
+    }
+
+    fun sayHello() {
+        repo.sayHello().execute { copy(message = it) }
```
gpeal, line 25:
```suggestion
        suspend { repo.sayHello() }.execute { copy(message = it) }
```

hellohilt/src/main/java/com/airbnb/mvrx/hellohilt/di/AppModule.kt
```diff
@@ -0,0 +1,21 @@
...
+/**
+ * The InstallIn SingletonComponent scope must match the context scope used in [HiltMavericksViewModelFactory]
+ *
+ * If you want an Activity or Fragment scope
+ */
+@Module
+@InstallIn(SingletonComponent::class)
+interface AppModule {
+
+    @[Binds IntoMap ViewModelKey(HelloViewModel::class)]
```
gpeal, line 19: Can you split this into 3 lines (see comment in README)

gpeal (commented):
Thanks for doing this! Excited to see AssistedInject in Dagger!

hellohilt/build.gradle
marukami: Ah, yep. Did not see that.

hellohilt/README.md
marukami: Oh dam, totally missed this. Yep

hellohilt/README.md
marukami: I was thinking about that. It should be as simple as creating a custom `@EntryPoint` for each scope and adding `@Qualifier` to the ViewModel's to split them into each scope… I think

marukami:
Closing in favour of #503

PR #500, opened 2021-01-17
Rename all gradle properties to Mavericks

gpeal:


mvrx-rxjava2/gradle.properties
```diff
@@ -1,4 +1,4 @@
-POM_NAME=MvRx
-POM_ARTIFACT_ID=mvrx-rxjava2
+POM_NAME=Mavericks
+POM_ARTIFACT_ID=mavericks-rxjava2
```
elihart, line 2: should this pr include updates to the docs that reference these artifacts as well?

elihart (approved):

gpeal:
@elihart Good call! Updated the docs. It's probably worth doing _one_ more to update the module names.

gpeal:
@elihart Can you take another look at this?

elihart (approved):
@gpeal looks great, thanks! Still planning to time merging this with when we do the repo rename?

gpeal:
@elihart I don't think this needs to wait for the repo rename. Do you?

elihart:
The readme details and docs links won't be right until we do, and there may be some confusion in people on pre-2.0 versions that look at it

PR #508, opened 2021-02-02
Updated Dagger sample and docs to use Dagger assisted injeciton

gpeal:
As of 2.31.0, we can migrate from square assisted inject to dagger's internal version.


elihart (approved):
looks great!

PR #510, opened 2021-02-10
Fix typo in async documentation.

robertoestivill:
Documentation specifies the `execute( )` parameter to be named `retainValue` while the example was using `retailValue`.


gpeal (approved):
Thanks!

PR #514, opened 2021-02-28
[Compose] Create mavericks-compose

### LINKED ISSUE #437: (Better) Support for Jetpack Compose
I've been using MvRx (stable release) with Jetpack Compose in a recent project, and although it is possible to simply call `setContent()` every time `invalidate()` gets called, that's not really the "Compose way" of doing things. Compose already has handy shortcuts for accessing state from `LiveData` and `RxJava2`, and it would be nice if there was a MvRx equivalent, like `viewModel.subscribeAsState()` that could completely avoid `invalidate`. 

The implementation of this for RxJava2 can be seen [here](https://cs.android.com/androidx/platform/frameworks/support/+/androidx-master-dev:compose/runtime/runtime-rxjava2/src/main/java/androidx/compose/runtime/rxjava2/RxJava2Adapter.kt;l=44) and it seems to me like it would be pretty easy to adapt that for MvRx.

gpeal:
This PR introduces official Mavericks Compose APIs.
Usage is as follows:
```kotlin
@Composable
fun MyComponent() {
    val viewModel: CounterViewModel = mavericksViewModel()
    val count by viewModel.collectState(CounterState::count)
    Button(onClick = activityScopedViewModel::incrementCount) {
        Text("Count: $count")
    }
}
```

Things seem to be working well and you can scope the view model to arbitrary places by overriding the lifecycle owner (and optionally the ViewModelStoreOwner and SavedStateRegistryOwner).

Ideas and other thoughts welcome but this seems to be a nice parallel to the Jetpack ViewModel [interop](https://developer.android.com/jetpack/compose/interop#viewmodel).

There were a few small changes needed to base Mavericks so I'll put those up in separate individual PRs as well.

For this PR, we could:
1) Land this on master. This would mean that master is using AGP 7 alphas which should be fine but isn't ideal.
2) Land this on a long running feature branch `jetpack-compose` and merge future PRs and changes on master into that branch.

I think just going straight for option 1 is fine but I'm open to a discussion.

Fixes #437


docs/jetpack-compose.md
```diff
@@ -1,7 +1,37 @@
...
 
-When Compose and the associated tooling (Android Studio, Android Gradle Plugin, etc) are closer to being finalized, we will release an official Compose artifact.
+To get started, create a MavericksViewModel just like you always have. Now, with the `mavericks-compose` artifact, you can get or create a ViewModel in a composable like this:
+
+```kotlin
+@Composable
+fun MyComponent() {
+    val viewModel: CounterViewModel = mavericksViewModel()
+    val count by viewModel.collectState(CounterState::count)
+    Button(onClick = activityScopedViewModel::incrementCount) {
```
elihart, line 14: what is `activityScopedViewModel` referencing here.? should it be `viewModel`?

docs/jetpack-compose.md
```diff
@@ -1,7 +1,37 @@
...
+        Text("Count: $count")
+    }
+}
+```
+
+If you call `collectAsState()` with no parameters, it will return a Compose state property with the whole ViewModel state class and will update any time the state changes.
+
+The recommended approach is to subscribe individually to a small set of State properties and to break down large Composables into smaller ones. This allows Compose to more efficiently recompose your UI when state changes.
```
elihart, line 22: Do we provide utils to easily subscribe to nested properties for efficiency? For example `viewModel.collectState(CounterState::listing).map { it.host }`

docs/jetpack-compose.md
```diff
@@ -1,7 +1,37 @@
...
+When considering what type of code to put in Compose and what to put in Mavericks, consider:
+
+1. Compose renders your screen as a function of State.
+2. Mavericks creates and updates the state.
+
+To use a concrete example, if you are fetching the weather, you would use Mavericks to fetch the weather for a specific zip code from an API using `execute`.
+
+Your screen would then subscribe to that state and emit the appropriate UI using composables.
+
+If the user wanted to look up the weather for another location, they would use the UI build with compose to type in a new location, and hit submit. That composable would then call a function on the MavericksViewModel to update the location which would then trigger a new weather API `execute` call.
```
elihart, line 35:
```suggestion
If the user wanted to look up the weather for another location, they would use the UI built with compose to type in a new location, and hit submit. That composable would then call a function on the MavericksViewModel to update the location which would then trigger a new weather API `execute` call.
```

docs/jetpack-compose.md
```diff
@@ -1,7 +1,37 @@
...
-We haven't settled on a final API yet but have already begin to explore what one might look like. [Here](https://gist.github.com/gpeal/5bed618a843d4d71bb1518fa8317c5df) are some snippets of working code that enables you to use Mavericks with Compose today. It is likely that the final API will look extremely similar to this.
+To use Mavericks with Jetpack Compose, add the `com.airbnb.android:mavericks-compose` artifact (TBD on versioning).
 
-When Compose and the associated tooling (Android Studio, Android Gradle Plugin, etc) are closer to being finalized, we will release an official Compose artifact.
+To get started, create a MavericksViewModel just like you always have. Now, with the `mavericks-compose` artifact, you can get or create a ViewModel in a composable like this:
+
+```kotlin
+@Composable
+fun MyComponent() {
+    val viewModel: CounterViewModel = mavericksViewModel()
```
elihart, line 12: it might be helpful to mention how mavericks knows which scope to subscribe to (activity, fragment, or something else). Can we manually specify scope to differentiate?

docs/jetpack-compose.md
elihart: looks like this uses a compose State, so I assume that has mapping. maybe worth calling out as an example

mvrx-compose/src/main/kotlin/com/airbnb/mvrx/compose/MavericksComposeExtensions.kt
```diff
@@ -0,0 +1,113 @@
...
+    return stateFlow.collectAsState(initial = withState(this) { it })
+}
+
+/**
+ * Creates a Compose State variable that will only update when the value of this property changes.
+ * Prefer this to subscribing to entire state classes which will trigger a recomposition whenever any state variable changes.
+ * If you find yourself subscribing to many state properties in a single composable, consider breaking it up into smaller ones.
+ */
+@Composable
+fun <VM : MavericksViewModel<S>, S : MavericksState, A> VM.collectState(prop1: KProperty1<S, A>): State<A> {
```
elihart, line 43: have you considered a syntax with a lambda with state as receiver so we could do `vm.collectState { myProp }`

I think this would have the benefits of:
- shorter from not invoking the state class name
- maybe faster, with no reflection
- easily allow watching nested properties `vm.collectState { myProp.nestedProp }`

mvrx-compose/src/main/kotlin/com/airbnb/mvrx/compose/MavericksComposeExtensions.kt
```diff
@@ -0,0 +1,113 @@
...
+}
+
+/**
+ * Get or create a [MavericksViewModel] scoped to the closest [LocalLifecycleOwner]. This should be sufficient for most cases. By default, it
+ * will be the host Activity or Nav Graph (if used). However, you can provide a custom scope by overriding the lifecycleOwner parameter.
+ *
+ * The ViewModelStoreOwner will default to the provided LifecycleOwner if it also implements ViewModelStoreOwner. Many LifecycleOwners such as
+ * Fragment, Activity, and NavBackStackEntry do. The same applies to the SavedStateRegistry.
+ *
+ * If the provided LifecycleOwner doesn't implement ViewModelStoreOwner or SavedStateRegistryOwner and you haven't specified one, it will default
```
elihart, line 54: do we think there are good use cases where LifecycleOwner doesn't implement ViewModelStoreOwner , or is that more likely to be an accident? Just wondering if we should throw instead to avoid people making mistakes, and force them to be explicit

mvrx-compose/src/main/kotlin/com/airbnb/mvrx/compose/MavericksComposeExtensions.kt
```diff
@@ -0,0 +1,113 @@
...
+ * to LocalViewModelStoreOwner and LocalSavedStateRegistryOwner.
+ *
+ * You can call functions on this ViewModel to update state. To subscribe to its state, call collectAsState(YourState::yourProp) or collectAsState().
+ */
+@Composable
+inline fun <reified VM : MavericksViewModel<S>, reified S : MavericksState> mavericksViewModel(
```
elihart, line 60: do you think it would be helpful to provide helpers like `fragmentViewModel` and `activityViewModel` that people are already familiar with, so they can more easily specify without needing to pass arguments?

docs/jetpack-compose.md
elihart: I see the code comments explain this well. might be nice to have a brief summary or mention in these docs though

sample-compose/src/main/java/com/airbnb/mvrx/compose/sample/ComposeSampleActivity.kt
```diff
@@ -0,0 +1,99 @@
...
+            }
+        }
+    }
+
+    @Composable
+    fun CounterScreen(title: String) {
+        // This will get or create a ViewModel scoped to the closest LocalLifecycleOwner which is the NavHost.
+        val navScopedViewModel: CounterViewModel = mavericksViewModel()
+        // This will get or create a ViewModel scoped to the Activity.
+        val activityScopedViewModel: CounterViewModel = mavericksViewModel(LocalContext.current as LifecycleOwner)
```
elihart, line 68: the syntax `mavericksViewModel(LocalContext.current as LifecycleOwner)` for getting an activity scoped view model doesn't seem too intuitive to me

elihart (approved):
@gpeal looks great! thanks for doing all this work, it's really exciting that mavericks can support Compose out of the box now and integrate so cleanly 🎉

sample-compose/src/main/java/com/airbnb/mvrx/compose/sample/ComposeSampleActivity.kt
gpeal: Agreed. Do you have other ideas?

docs/jetpack-compose.md
gpeal: Copy and paste :)

docs/jetpack-compose.md
gpeal: I just added an overload to allow you to do `collectState { it.listing.host }` which would allow for this. You could also make host a derived prop.

mvrx-compose/src/main/kotlin/com/airbnb/mvrx/compose/MavericksComposeExtensions.kt
gpeal: Added!

mvrx-compose/src/main/kotlin/com/airbnb/mvrx/compose/MavericksComposeExtensions.kt
gpeal: @elihart I have no idea to be honest. What do you think? It could be error prone to provide incorrect values with different lifecycles here so we could start with it restricted but could open it up only if we know of a valid use case. What do you think?

sample-compose/src/main/java/com/airbnb/mvrx/compose/sample/ComposeSampleActivity.kt
gpeal: I added `mavericksActivityViewModel()`

mvrx-compose/src/main/kotlin/com/airbnb/mvrx/compose/MavericksComposeExtensions.kt
elihart: I agree that until we have a better handle on the use cases it would be safer to start more restrictive and loosen it up later if needed

sample-compose/src/main/java/com/airbnb/mvrx/compose/sample/ComposeSampleActivity.kt
elihart: I like `mavericksActivityViewModel`. Maybe also `mavericksLocalViewModel`? ( suppose that would replace fragment scoping).

i think this is something we can leave out for now though until we get a better handle on use cases

elihart (approved):
@gpeal looks great, thanks for the clean ups.

PR #515, opened 2021-02-28
Upgrade Kotlin and Fragment

gpeal:
This is in preparation for the Mavericks PR.


buildSrc/src/main/java/dependencies.kt
```diff
@@ -3,7 +3,7 @@ object Versions {
     const val buildTools = "29.0.3"
     const val compileSdk = 29
     const val gradlePlugin = "4.1.1"
-    const val kotlin = "1.4.10"
+    const val kotlin = "1.4.30"
```
elihart, line 6:
```suggestion
    const val kotlin = "1.4.31"
```

elihart (approved):
Great clean up, thanks!

buildSrc/src/main/java/dependencies.kt
gpeal: @elihart I actually didn't use 1.4.31 because the Compose Compiler seems like it's hard coded with supported versions and every build was throwing warnings when I used 1.4.31

PR #516, opened 2021-02-28
Replace FragmentActivity with ComponentActivity

gpeal:
This will help with the migration to Compose because it will only need androidx.activity not androidx.appcompat.


elihart (approved):
👍

PR #518, opened 2021-03-01
Remove @RestrictTo in favor of just @InternalMavericksApi

gpeal:
The Kotlin opt-in annotations work more reliably than the Android lint rules and there is no need for both.

This is necessary because when I upgraded to AGP 7 for Compose, it broke a bunch of `@RestrictTo` calls across Mavericks libraries that were marked as LIBRARY_GROUP and are in the same library group but fail lint (but don't show up as errors in Android Studio). The entire RestrictTo feature has been continually broken in some way while `@InternalMavericksApi` has been working reliably so let's just stick to it.

I left one `@RestrictTo` in `BaseMvRxViewModel` for legacy reasons. It has some valid use cases for external users so we never added `@InternalMavericksApi` and I'm leaving it untouched.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,8 @@
 # Change Log
 
+## Version 2.1.0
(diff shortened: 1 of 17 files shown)


elihart (approved):
makes sense to me 👍

PR #524, opened 2021-03-09
enable hilt sample

diousk:
hilt sample is not import in settings.gragle therefore it can not be seen in android studio's project list


gpeal (approved):
Thanks! Not sure how this happened :)

PR #525, opened 2021-03-09
Remove createUnsafe and don't auto-subscribe on background threads

marukami:
Turns out `createUnsafe` does not work… at least, it breaks more than it fixes 😅 I double-checked all the test and ran a few demo apps to check.

I think a safer fallback is if the ViewModel is been created off the main thread then you must manually trigger the initialize by reading the ViewModel. The Mockables seemed fine in the limited test. helloDagger and the sample projects all seem to work fine with this method.

Sorry about this @gpeal I should have been more thorough in my testing, I got too focused on just the mockable not crashing.

### DIFF
--- a/mvrx/src/main/kotlin/com/airbnb/mvrx/lifecycleAwareLazy.kt
+++ b/mvrx/src/main/kotlin/com/airbnb/mvrx/lifecycleAwareLazy.kt
@@ -3,6 +3,8 @@
 package com.airbnb.mvrx
 
 import android.annotation.SuppressLint
+import android.os.Handler
+import android.os.Looper
 import androidx.lifecycle.DefaultLifecycleObserver
 import androidx.lifecycle.LifecycleOwner
 import androidx.lifecycle.LifecycleRegistry
@@ -30,13 +32,14 @@ class lifecycleAwareLazy<out T>(private val owner: LifecycleOwner, initializer: 
 
     init {
         // owner.lifecycle.addObserver must be invoked on main thread otherwise addObserver will throw an IllegalStateException.
-        // createUnsafe disables the main thread check.
-        LifecycleRegistry.createUnsafe(owner).addObserver(object : DefaultLifecycleObserver {
-            override fun onCreate(owner: LifecycleOwner) {
-                if (!isInitialized()) value
-                owner.lifecycle.removeObserver(this)
-            }
-        })
+        if (Looper.myLooper() == Looper.getMainLooper()) {
+            owner.lifecycle.addObserver(object : DefaultLifecycleObserver {
+                override fun onCreate(owner: LifecycleOwner) {
+                    if (!isInitialized()) value
+                    owner.lifecycle.removeObserver(this)
+                }
+            })
+        }
     }
 
     @Suppress("LocalVariableName")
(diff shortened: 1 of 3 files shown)


gpeal:
We _could_ post a runnable to invoke this on the next frame, then check to make sure it's not already destroyed to initialize it.

Does this solution work for robolectric tests? Doesn't it essentially re-add the main thread check that broke robolectric in the first place?

gpeal:
@elihart

marukami:
> Does this solution work for robolectric tests? 

Yes, they all seem to pass, the last solution fails the tests. 

> Doesn't it essentially re-add the main thread check that broke robolectric in the first place?

robolectric works fine and does not throw as far as I can tell.

> We could post a runnable to invoke this on the next frame, then check to make sure it's not already destroyed to initialize it.

Sure, let me update it this morning (my time)

mvrx/src/main/kotlin/com/airbnb/mvrx/lifecycleAwareLazy.kt
```diff
@@ -29,9 +34,18 @@ class lifecycleAwareLazy<out T>(private val owner: LifecycleOwner, initializer:
...
 
     init {
-        // owner.lifecycle.addObserver must be invoked on main thread otherwise addObserver will throw an IllegalStateException.
-        // createUnsafe disables the main thread check.
-        LifecycleRegistry.createUnsafe(owner).addObserver(object : DefaultLifecycleObserver {
+        if (isMainThread()) {
+            initializeOnCreatedOrGreater(owner)
+        } else {
+            Handler(Looper.getMainLooper()).post {
+                if (owner.lifecycle.currentState < Lifecycle.State.CREATED || isInitialized()) return@post
```
elihart, line 41: @marukami this currently returns and does not initialize if the lifecycle is at the `INITIALIZED` state, which seems like it could be a problem. afaict we would only want to cancel if the lifecycle is destroyed

```suggestion
                if (owner.lifecycle.currentState == DESTROYED || isInitialized()) return@post
```

mvrx/src/main/kotlin/com/airbnb/mvrx/lifecycleAwareLazy.kt
```diff
@@ -29,9 +34,18 @@ class lifecycleAwareLazy<out T>(private val owner: LifecycleOwner, initializer:
...
+                initializeOnCreatedOrGreater(owner)
+            }
+        }
+    }
+
+    private fun initializeOnCreatedOrGreater(owner: LifecycleOwner) {
+        owner.lifecycle.addObserver(object : DefaultLifecycleObserver {
```
elihart, line 48: as an optimization this could check the state and if it is created or greater then do the initialization immediately, instead of going through the work to create, add, and then remove an observer

elihart (commented):
Generally looks great, just have some edge cases thoughts

mvrx/src/main/kotlin/com/airbnb/mvrx/lifecycleAwareLazy.kt
marukami: Oh yer. Sure, that sounds like a plan. Updated PR

mvrx/src/main/kotlin/com/airbnb/mvrx/lifecycleAwareLazy.kt
marukami: Ah, yep. I follow what you mean here.

elihart (approved):
Looks great to me, thanks!

PR #539, opened 2021-04-21
Fix typos in kdocs in MavericksViewModel

pmecho:
Looks like there are a few typos in the `reducer` param documentation on some functions so fixing those.


elihart (approved):
Thanks!

Issue #535, opened 2021-04-15
[Proposal] Global error handler for execute functions


jakoss:
I have the requirement to trace all errors in app, even if those are properly handled by the code and UI. So i need to create some global error handler for all `execute` function, that will be called on `Fail` reducer set. My proposal:

Add new property to `MavericksViewModelConfig` with signature something like:
```
val executeErrorHandler: ((Exception) -> Unit)? = null
```
I'm not sure that should be called handler, since it wouldn't have any impact on error handling, it's just for logging/tracing purposes. Maybe name `executeErrorLogger` would be more appropriate?

Next, i'd add the same field to `MavericksViewModelConfigFactory` that would be passed down to configs.

In `MavericksViewModel` i'd add call `config.executeErrorHandler?.invoke(e)` after `setState { reducer(Fail(e ...` for all `execute` fucntions.

What do you think about that? If it's fine i'll create a PR with that.

Also, i have a minor question about error handling - is there a particular reason to catch `Exception` instead of `Throwable`?

gpeal:
@jakoss How would you propose failures would be detected. Reducers are opaque to Mavericks. We don't know what each one does and wouldn't be able to tell you where the error originated from either.

re: Throwable. We could probably switch to that if it's causing issues. It just hasn't caused any issues (that we've heard of) yet.

jakoss:
I wouldn't detect failure on reducer, just after it in execute functions:
here: https://github.com/airbnb/mavericks/blob/f4b749240322463dbc07deb91490f912f6e6f029/mvrx/src/main/kotlin/com/airbnb/mvrx/MavericksViewModel.kt#L230
and here: https://github.com/airbnb/mavericks/blob/f4b749240322463dbc07deb91490f912f6e6f029/mvrx/src/main/kotlin/com/airbnb/mvrx/MavericksViewModel.kt#L262

We have Exception there.

In my experience with android i had 2 situations with code that crashed on production (native library integration issue) just because i was catching Exception instead of Throwable. Since Throwable handles basically all - it might be safer, especially if user assumes that execute is totally safe in that regard

gpeal:
@elihart What do you think of adding a config for this?

@jakoss I'l be open to a PR to change Exception to Throwable

elihart:
This all sounds great to me, thanks for laying out a clear plan. Agree with `Throwable`.

[...]

PR #540, opened 2021-04-23
Error handling enhancements

jakoss:
Minor enhancements, that allow for better error handling in `execute` functions:

- Catch `Throwable` instead of `Exception` - to handle wider range of errors
- Mark `execute` and `setOnEach` functions as `open`, so user can override them to enhance `Async` handling (for example by adding logging)

More context at #535


elihart (approved):
Thanks, lgtm

PR #553, opened 2021-06-10
Bumps AGP and Hilt versions to unblock development

allenchen1154:
When using Android Studio Arctic Fox, the current AGP version errors during sync with 
```
Gradle sync failed: Invalid injected android support version '202.7660.26.42.7322048', expected to be of the form 'w.x.y.z'
```

See https://issuetracker.google.com/issues/187470273


elihart (approved):
Thanks!

Issue #551, opened 2021-06-04
Allow retrieval of ViewModels without using declaration site types


allenchen1154:
In certain situations, library or shared code may wish to retrieve a Mavericks ViewModel while only being able to use a parent type of the runtime one at the declaration site. For example:
```kotlin
// Parent ViewModel and State, located in a library module
abstract class ParentViewModel<S : ParentState>(initialState: S) : BaseMvRxViewModel<S>(initialState) { ... }

abstract class ParentState : MvRxState

// Child ViewModel and State, located in a module that depends on the library module.
class ChildViewModel(initialState: ChildState) : ParentViewModel<ChildState>(initialState) { ... }

class ChildState : ParentState { ... }

class FragmentA {
    val childViewModel: ChildViewModel by activityViewModel()

    // FragmentA can launch FragmentB
}

class FragmentB {
    val parentViewModel: ParentViewModel<ParentState> by activityViewModel()
}

```
Currently, the retrieval in `FragmentB` would fail unless we provide a custom key factory shared with `FragmentA`. Also, if the `Activity` hosting the `Fragment`s is recreated with `FragmentB` on top of the stack, the delegate will fail to recreate the ViewModel, since it doesn't know the runtime types are `ChildViewModel`/`ChildState`.

In order to accomplish this, I believe we would need the ability to pass `KClass` arguments representing both the ViewModel and its generic state type. Currently, the delegates in `MavericksExtensions.kt` only allow the former to be passed:

```kotlin
class FragmentBArgs(
    val viewModelClass: KClass<out BaseMvRxViewModel<*>>,
    val stateClass: KClass<out MvRxState>,
)

class FragmentB {
    val args: FragmentBArgs by args()

    val parentViewModel: ParentViewModel<ParentState> by activityViewModel(viewModelClass = args.viewModelClass, stateclass = args.stateclass)
}
```

gpeal:
@allenchen1154 All of the view model delegates except `existingViewModel()` have the ability to get _or create_ a new view model so we would need the final concrete type in order to create it.

`existingViewModel()` might be able to do this. It will crash if the ViewModel doesn't already exist. It currently behaves like `activityViewModel()` so if you are okay with that as a scope, maybe something could be done.

BenSchwab:
@gpeal @allenchen1154 It does seem like using `existingViewModel` is the right direction. But in the recreate case we would need to know the type of the state class to be properly recreated.

[...]

PR #556, opened 2021-06-17
Add support for retrieving ViewModels with covariant state types without using declaration site types

allenchen1154:
In certain situations, library or shared code may wish to retrieve a Mavericks ViewModel while only having access to a parent type of the runtime one at the declaration site.

These changes allow callers of the delegates in MavericksExtensions.kt to provide a `KClass` parameter specifying the actual runtime class of the desired `MavericksState`, rather than always using the type of the declared property.

Because such `MavericksViewModel` parent types are often generics with a bound on the `MavericksState` class, a common pattern is to declare a covariant type for the state, such as `ParentViewModel<out ParentState>`. In order to support this without intermediate variable declarations or type casts, we also need to add support for covariant state types to several Mavericks features, such as `withState()` calls, the `mockSingleViewModel()` family of calls, and the `selectSubscribe()` calls.

Also attempts to add unit tests covering these covariant use cases. 

See https://github.com/airbnb/mavericks/issues/551 for more details.


mvrx/src/test/kotlin/com/airbnb/mvrx/AbstractViewModelDeclarationTest.kt
```diff
@@ -0,0 +1,31 @@
...
+
+class AbstractViewModelDeclarationTest : BaseTest() {
+    /**
+     * Tests that [FragmentWithAbstractViewModelDeclaration] can create an instance of [ChildViewModel]
+     * using [KClass] arguments to the delegates in MavericksExtensions.kt, while declaring it as
+     * an abstract parent class.
+     */
+    @Test
+    fun testFragmentCreationDoesNotError() {
+        launchFragment<FragmentWithAbstractViewModelDeclaration>()
```
BenSchwab, line 29: I think we need to interact with the view model in some way, as the fragmentViewModel delegate is lazy. Do we need this test at all? I think it's behavior is covered in the next file `CovariantStateTest`

BenSchwab (approved):
This looks great to me - I think this covers most functions but I noticed you didn't add out to `subscribe` in MvRxViewExtensions - was there a reason? Ideally we support that one too.

BenSchwab:
Closing in favor of https://github.com/airbnb/mavericks/pull/565

PR #560, opened 2021-06-30
Adding UniqueUpdatesOnly delivery mode for subscriptions.

podarsmarty:
Adding a new delivery mode so that people can subscribe to only changes and exclude the initial state from being emitted.


gpeal:
@podarsmarty Can you do this manually like:
```
var isFirst = true
viewMode.onEach(MyState::myProp) { myprop ->
    if (isFirst) {
        isFirst = false
        return@onEach
    }
}
```

elihart:
Yes, that should work @gpeal but we are seeing this need in multiple places and it would be very nice to have a built in way to accomplish it. Unfortunately it is not easy to provide our own internal extensions for this behavior due to how many `forEach` overloads there are.

I think an alternative approach would be to leverage the state flow and have an extension function we apply there, but the "unique only" logic is internal to Mavericks and available to apply to state flow operators.

gpeal:
@elihart out of curiosity, what is the use case?

podarsmarty:
@gpeal the use case I was running into is say the state stores a map. the key/values are what you want to use to populate your query params to the server. If any of them change you want to do a refresh of the page

mvrx/src/main/kotlin/com/airbnb/mvrx/MavericksViewModel.kt
```diff
@@ -440,33 +434,42 @@ abstract class MavericksViewModel<S : MavericksState>(
         }
     }
 
+    private fun <T : Any> Flow<T>.configureAndFilter(lifecycleOwner: LifecycleOwner, subscriptionId: String, skipCount: Int = 0): Flow<T> {
```
elihart, line 437: I think it would be nice to expose a Flow extension function with similar behavior so users can have easy delivery mode configuration when working with the `stateFlow` directly. We'd probably want to think a bit more carefully about what that Api looks like though

mvrx/src/main/kotlin/com/airbnb/mvrx/DeliveryMode.kt
```diff
@@ -11,6 +11,7 @@ sealed class DeliveryMode {
         return when (this) {
             is RedeliverOnStart -> RedeliverOnStart
             is UniqueOnly -> UniqueOnly(properties.joinToString(",", prefix = subscriptionId + "_") { it.name })
+            is UniqueUpdatesOnly -> UniqueOnly(properties.joinToString(",", prefix = subscriptionId + "_") { it.name })
```
elihart, line 14: @podarsmarty was this a copy paste mistake?
```suggestion
            is UniqueUpdatesOnly -> UniqueUpdatesOnly(properties.joinToString(",", prefix = subscriptionId + "_") { it.name })
```

elihart (commented):
I think this implementation is reasonable and good functionality to provide, but if you can suggest an alternate way to accomplish it through the library @gpeal I'd be open to that too

mvrx/src/main/kotlin/com/airbnb/mvrx/DeliveryMode.kt
podarsmarty: oops yea, my b

gpeal:
@elihart @podarsmarty What about something like:
```
onViewCreated() {
    viewLifecycleOwner.lifecycleScope.launchRepeatingJob(ON_START) {
        viewModel.stateFlow
            .map { it.foo }
            .distinctUntilChanged()
            .drop(1)
            .collect(::something)
        }
    }
}
```
Would that work? You could make a helper if your base fragment or something that did that too

elihart:
I don't think that works because it does not prevent the redelivery of the last value on view recreation, in order to do that the logic in `resolveSubscription` needs to be exposed, which is what I was referencing above https://github.com/airbnb/mavericks/pull/560#discussion_r662636356

However, @podarsmarty if your use case is to make a network request with query params change shouldn't the subscription be in the ViewModel, not the Fragment? In that case there is no resubscription needed.

podarsmarty:
@elihart hmm yea that should work if we were only doing the refresh. we are doing a couple things to update things like the toolbar. it is still possible without this but it becomes pretty messy since we will need to subscribe to the same thing.

I think it would also break down if the fragment had 2 view models so that it could share some stuff with a diff screen. ie - viewmodel a has some property that would trigger a refresh in viewmodel b.

@gpeal is the concern b/c we dont want to have a growing list of delivery modes?

gpeal:
For separate subscriptions for data refreshing in the ViewModel and updating the toolbar in the Fragment, that seems perfectly reasonable to me and I agree with Eli's sentiment about the data refreshing belonging int he ViewModel.

For 2 ViewModels in one fragment, at Tonal, we used to do this and encountered exactly that issue. Now, we utilize Dagger/DI for any shared data so that all data wiring happens in your object graph or view models. Coupling data wiring in Fragments makes things harder to tests and couples functionality to UI more tightly than you usually want.

Yeah, I think one of the biggest benefits of Mavericks is that it has a small API surface area so we need to be really careful about adding anything that isn't necessary or significantly increases the usability of the library.

podarsmarty:
@gpeal hmm sounds like we prob want to figure out the API @elihart mentioned now rather then later then?

A "Custom" delivery mode that takes a lambda that acts on a flow would work. We would just have to expose lastDeliveredValue though

podarsmarty:
@gpeal @elihart, aight i went through and updated this so that it is just a singular custom delivery mode. This way people can get the behavior of unique updates only or customize it to suit their needs without adding another delivery mode

mvrx/src/main/kotlin/com/airbnb/mvrx/DeliveryMode.kt
```diff
@@ -34,3 +40,15 @@ object RedeliverOnStart : DeliveryMode()
...
+ * The subscription will receive the most recent state update when transitioning from locked to unlocked states (stopped -> started) after the
+ * caller's provided custom behavior via [flowConfiguration]. If [distinctOnly] is true, this will behave like []UniqueOnly].
+ *
+ * Likewise, when a [MavericksView] resubscribes after a configuration change the most recent update will be emitting after processing the desired
+ * custom behavior.
+ *
+ * @param subscriptionId A uniqueIdentifier for this subscription. It is an error for two unique only subscriptions to
+ * have the same id.
+ */
+class Custom(val subscriptionId: String, val distinctOnly: Boolean, val flowConfiguration: (Flow<*>, DeliveryMode) -> Flow<*>) : DeliveryMode()
```
elihart, line 54: important to add kdoc for each new parameter here. same for `fun custom(` below

elihart, line 54: I think it's a little weird to expose both `flowConfiguration` and `distinctOnly`, since if we are providing custom flow configuration then they can apply `distinctUntilChanged` easily enough on their own, and our goal is not to parameterize each possible flow operator

mvrx/src/main/kotlin/com/airbnb/mvrx/MavericksViewModel.kt
```diff
@@ -416,18 +417,22 @@ abstract class MavericksViewModel<S : MavericksState>(
...
-                .dropWhile { it == lastDeliveredValue }
-                .flowWhenStarted(lifecycleOwner)
-                .distinctUntilChanged()
-                .onEach { lastDeliveredStates[deliveryMode.subscriptionId] = it }
-        } else {
-            flowWhenStarted(lifecycleOwner)
+        val flow = when {
+            lifecycleOwner == null || MavericksTestOverrides.FORCE_DISABLE_LIFECYCLE_AWARE_OBSERVER -> this
+            deliveryMode is UniqueOnly -> configureAndFilter(lifecycleOwner, deliveryMode.subscriptionId, true, deliveryMode) { flow, _ ->
+                flow.dropWhile { it == lastDeliveredValue(deliveryMode.subscriptionId) }
```
elihart, line 423: is there a preexisting bug where an null value will not get delivered, even if no value was previously delivered, because `null` also represents not present in the map?

elihart (commented):
@podarsmarty I think this is reasonable, but also a bit of a weird middle ground that combines our custom subscription function with functional flows. Instead of trying to wed the two, maybe we should focus instead on what this would look like completely functionally with flows.

I believe usage of the current implementation would like like this
```
viewModel.onEach(State::prop, deliveryMode = Custom(customId, true) { flow, deliveryMode ->
    flow.drop(1)
        .dropWhile { it == viewModel.lastDeliveredValue((deliveryMode as? com.airbnb.mvrx.Custom)?.subscriptionId) }
}) {
  // use value
}
```

With pure flows and extension functions it could instead look like
```
viewModel.stateFlow
    .map { it.prop }
    .drop(1)
    .uniqueOnly(id = "prop subscription") // extension function to provide unique subscription across recreation
    .flowWhenStarted(this) // extension to start flow per lifecycle event
    .collectLatest(viewModel) { // extension to start collection using viewmodel scope and dispatcher logic

    }
    ```
    
And if we cleaned up the extension functions to combine them in a reasonable way we may be able to make it even simpler
```
viewModel.stateFlow
    .map { it.prop }
    .drop(1)
    // extension can combine functionality of unique only, flowWhenStarted, and collect latest.
    // If defined within `MavericksView` it can have access to subscriptionLifecycleOwner.
    .collectLatest(viewModel, deliveryMode = UniqueOnly("unique id")) {

    }
    ```
    
What do you think?

mvrx/src/test/kotlin/com/airbnb/mvrx/FragmentSubscriberTest.kt
```diff
@@ -58,13 +71,19 @@ open class ViewSubscriberFragment : Fragment(), MavericksView {
...
 
     fun setFoo(foo: Int) = viewModel.setFoo(foo)
 
     override fun invalidate() {
         invalidateCallCount++
     }
+
+    private fun custom(customId: String) = custom(customId, true) { flow, deliveryMode ->
+        flow.drop(1)
+            .dropWhile { it == viewModel.lastDeliveredValue((deliveryMode as? Custom)?.subscriptionId) }
```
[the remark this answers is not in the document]
podarsmarty: hmm i like that, updating

mvrx/src/main/kotlin/com/airbnb/mvrx/MavericksViewModel.kt
podarsmarty: yea, i think that would be the case. If initial value is null and it is never changed, uniqueOnly would not deliver that initial value

mvrx/src/main/kotlin/com/airbnb/mvrx/DeliveryMode.kt
podarsmarty: hmm i think if we do that we would need flowConfiguration to also handle `flowWhenStarted`. do we actually want to do that?

podarsmarty:
@elihart hmm i think that would work. in that case would we even want to keep Delivery? Cause then we'd have Redeliver & Unique via delivery mode and anything custom via a stateflow directly. im guessing wed just want to remove delivery mode in that case and folks would use the uniqueOnly extension on the flow?

mvrx/src/main/kotlin/com/airbnb/mvrx/MavericksViewModel.kt
```diff
@@ -415,68 +418,72 @@ abstract class MavericksViewModel<S : MavericksState>(
...
-                activeSubscriptions += deliveryMode.subscriptionId
-            }
+    override fun toString(): String = "${this::class.java.simpleName} $state"
+}
 
-            override fun onDestroy(owner: LifecycleOwner) {
-                activeSubscriptions.remove(deliveryMode.subscriptionId)
-            }
+fun <T : Any, S: MavericksState> Flow<T>.collectLatest(
+    viewModel: MavericksViewModel<S>,
```
elihart, line 438: I  think we can actually remove the viewmodel parameter from this once the function is moved to the MavericksView. The big benefit to this is it removes the weirdness around the expectation that the viewmodel you use is the same one that the state flow comes from, and creates a nicer API.

The viewmodel right now is used to store the active subscriptions and last value - that can instead live in the mavericks view via the same pattern we use for the mvrx view id
```
val mvrxViewId: String
        get() = when (this) {
            is ViewModelStoreOwner -> ViewModelProvider(this).get(MavericksViewIdViewModel::class.java).mavericksViewId
```

The viewmodel is also used for the config factory for the coroutine context override, but that can be accessed statically via `Mavericks.viewModelConfigFactory`?

We would then also have to refactor this function to be more generic to work for both the mavericksview case and the old subscription case inside the viewmodel, but we can help do that by creating a new class to hold the activeSubscriptions and lastDeliveredStates values and manage them.

What do you think? I know this is even more work, but should make the API quite a bit nicer and would be a hard thing to change later

mvrx/src/test/kotlin/com/airbnb/mvrx/FragmentSubscriberTest.kt
```diff
@@ -22,22 +25,30 @@ open class ViewSubscriberFragment : Fragment(), MavericksView {
...
     var viewCreatedSubscribeCallCount = 0
     var viewCreatedUniqueOnlyCallCount = 0
+    var viewCreatedCustomCallCount = 0
 
     override fun onCreate(savedInstanceState: Bundle?) {
         super.onCreate(savedInstanceState)
         viewModel.onEach { subscribeCallCount++ }
         viewModel.onEach(uniqueOnly("onCreate")) { subscribeUniqueOnlyCallCount++ }
+        viewModel.stateFlow.drop(1).collectLatest(viewModel, subscriptionLifecycleOwner, UniqueOnly("onCreate_Custom")) {
+            subscribeCustomCallCount++
```
elihart, line 50: Thanks for all the tests, I like how easy to use this API is shaping up to be

mvrx/src/main/kotlin/com/airbnb/mvrx/MavericksViewModel.kt
```diff
@@ -415,68 +418,72 @@ abstract class MavericksViewModel<S : MavericksState>(
...
+    lifecycleOwner: LifecycleOwner? = null,
+    deliveryMode: DeliveryMode,
```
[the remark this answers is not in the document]
podarsmarty: since we expect users to be using this to customize behavior, i was thinking we force them to supply the delivery mode they desire

mvrx/src/main/kotlin/com/airbnb/mvrx/DeliveryMode.kt
```diff
@@ -21,16 +21,16 @@ sealed class DeliveryMode {
...
+ * only if the state has changed while locked. This will include the initial state as a state update.
  *
  * Likewise, when a [MavericksView] resubscribes after a configuration change the most recent update will only be emitted
  * if the state has changed while locked.
  *
  * @param subscriptionId A uniqueIdentifier for this subscription. It is an error for two unique only subscriptions to
  * have the same id.
  */
-class UniqueOnly(val subscriptionId: String) : DeliveryMode()
+class UniqueOnly(override val subscriptionId: String) : DeliveryMode(subscriptionId)
```
elihart, line 36: it's redundant to pass the `subscriptionId` to the super class as a parameter, as well as override it here. Either don't make it open and pass it in the constructor, or take it out of the super constructor and put it in the body as an open property and override it here

mvrx/src/main/kotlin/com/airbnb/mvrx/MavericksView.kt
```diff
@@ -50,6 +52,15 @@ interface MavericksView : LifecycleOwner {
             )
         }
 
+    val mavericksViewSubscriptionViewModel: MavericksViewSubscriptionViewModel
+        get() = when (this) {
+            is ViewModelStoreOwner -> ViewModelProvider(this).get(MavericksViewSubscriptionViewModel::class.java)
```
elihart, line 57: might as well combine the MavericksViewIdViewModel and this one, into something like `MavericksViewInternalViewModel` and use it for all internal data we want to manage

mvrx/src/main/kotlin/com/airbnb/mvrx/MavericksView.kt
```diff
@@ -295,4 +306,16 @@ interface MavericksView : LifecycleOwner {
         onFail: (suspend (Throwable) -> Unit)? = null,
         onSuccess: (suspend (T) -> Unit)? = null
     ) = _internalSF(subscriptionLifecycleOwner, asyncProp, deliveryMode, onFail, onSuccess)
+
+    fun <T : Any> collectLatest(
```
elihart, line 310: this will need kdoc

elihart: can you change the Flow to be the function receiver?

mvrx/src/main/kotlin/com/airbnb/mvrx/MavericksView.kt
```diff
@@ -295,4 +306,16 @@ interface MavericksView : LifecycleOwner {
...
+        flow: Flow<T>,
+        deliveryMode: DeliveryMode,
+        action: suspend (T) -> Unit
+    ): Job = flow.collectLatest(
+        subscriptionLifecycleOwner,
+        mavericksViewSubscriptionViewModel.lastDeliveredStates,
```
elihart, line 316: this does `ViewModelProvider(this).get(MavericksViewSubscriptionViewModel::class.java)` twice - better to store in a local property to optimize.

mvrx/src/main/kotlin/com/airbnb/mvrx/MavericksViewModel.kt
```diff
@@ -56,7 +51,7 @@ abstract class MavericksViewModel<S : MavericksState>(
 
     private val stateStore = config.stateStore
     private val lastDeliveredStates = ConcurrentHashMap<String, Any>()
-    private val activeSubscriptions = Collections.newSetFromMap(ConcurrentHashMap<String, Boolean>())
+    private val activeSubscriptions: MutableSet<String> = mutableSetOf()
```
elihart, line 54: why remove the concurrency?

mvrx/src/main/kotlin/com/airbnb/mvrx/MavericksViewModel.kt
```diff
@@ -415,68 +410,29 @@ abstract class MavericksViewModel<S : MavericksState>(
...
-                .flowWhenStarted(lifecycleOwner)
-                .distinctUntilChanged()
-                .onEach { lastDeliveredStates[deliveryMode.subscriptionId] = it }
-        } else {
-            flowWhenStarted(lifecycleOwner)
-        }
-
-        val scope = (lifecycleOwner?.lifecycleScope ?: viewModelScope) + configFactory.subscriptionCoroutineContextOverride
-        return scope.launch(start = CoroutineStart.UNDISPATCHED) {
+    ): Job = if (lifecycleOwner != null) {
```
elihart, line 413: use block bodies for multi line functions

mvrx/src/main/kotlin/com/airbnb/mvrx/MavericksViewModel.kt
```diff
@@ -415,68 +410,29 @@ abstract class MavericksViewModel<S : MavericksState>(
...
             // Use yield to ensure flow collect coroutine is dispatched rather than invoked immediately.
             // This is necessary when Dispatchers.Main.immediate is used in scope.
             // Coroutine is launched with start = CoroutineStart.UNDISPATCHED to perform dispatch only once.
             yield()
-            flow.collectLatest(action)
-        }
-    }
-
-    @Suppress("EXPERIMENTAL_API_USAGE")
-    private fun <T> Flow<T>.assertOneActiveSubscription(owner: LifecycleOwner, deliveryMode: UniqueOnly): Flow<T> {
```
elihart, line 444: I think I'm missing something, where did this code get moved to?

mvrx/src/main/kotlin/com/airbnb/mvrx/MavericksViewModel.kt
podarsmarty: oh oops, didnt realize newSetFromMap kept it backed with the map you gave it. reverted

mvrx/src/main/kotlin/com/airbnb/mvrx/MavericksViewModel.kt
podarsmarty: omg, sorry. i had moved it to a new file and it wasnt tracked/committed

mvrx/src/main/kotlin/com/airbnb/mvrx/MavericksView.kt
podarsmarty: yup, done

mvrx/src/main/kotlin/com/airbnb/mvrx/MavericksView.kt
```diff
@@ -295,4 +300,31 @@ interface MavericksView : LifecycleOwner {
         onFail: (suspend (Throwable) -> Unit)? = null,
         onSuccess: (suspend (T) -> Unit)? = null
     ) = _internalSF(subscriptionLifecycleOwner, asyncProp, deliveryMode, onFail, onSuccess)
+
+    /**
+     * Subscribes to on the given Flow. This can be utilized to create customized subscriptions that for example drop the first
```
elihart, line 305: I think we need a bit more context here about how this works, and how its different from normal flow subscriptions.

```suggestion
     * Subscribes to the given Flow within the coroutine scope of the `subscriptionLifecycleOwner`, starting the flow only when the lifecycle is started, and executing with the coroutine context of Mavericks' `subscriptionCoroutineContextOverride`. This can be utilized to create customized subscriptions, for example, to drop the first
```

mvrx/src/main/kotlin/com/airbnb/mvrx/MavericksView.kt
```diff
@@ -295,4 +300,31 @@ interface MavericksView : LifecycleOwner {
...
+     * element in the flow before continuing.
+     *
+     * @param deliveryMode If [UniqueOnly], when this [MavericksView] goes from a stopped to start lifecycle a value
```
elihart, line 308:
```suggestion
     * @param deliveryMode If [UniqueOnly], when this [MavericksView] goes from a stopped to started lifecycle a value
```

mvrx/src/main/kotlin/com/airbnb/mvrx/MavericksView.kt
```diff
@@ -295,4 +300,31 @@ interface MavericksView : LifecycleOwner {
...
+     * will only be emitted if the value has changed. This is useful for transient views that should only
+     * be shown once (toasts, poptarts), or logging. Most other views should use false, as when a view is destroyed
```
elihart, line 310: false is not a valid option, do you mean this?
```suggestion
     * be shown once (toasts, poptarts), or logging. Most other views should use RedeliverOnStart, as when a view is destroyed
```

mvrx/src/main/kotlin/com/airbnb/mvrx/MavericksView.kt
```diff
@@ -295,4 +300,31 @@ interface MavericksView : LifecycleOwner {
...
+     * and recreated the previous state is necessary to recreate the view.
+     *
+     * Use [uniqueOnly] to automatically create a [UniqueOnly] mode with a unique id for this view.
+     *
+     * @param action supports cooperative cancellation. The previous action will be cancelled if it as not completed before
```
elihart, line 315:
```suggestion
     * @param action supports cooperative cancellation. The previous action will be cancelled if it is not completed before
```

mvrx/src/main/kotlin/com/airbnb/mvrx/MavericksView.kt
```diff
@@ -295,4 +300,31 @@ interface MavericksView : LifecycleOwner {
...
+     * the next one is emitted.
+     */
+    fun <T : Any> Flow<T>.collectLatest(
+        deliveryMode: DeliveryMode,
```
elihart, line 319: use a default value of RedeliverOnStart here, since that is what most people need?

mvrx/src/main/kotlin/com/airbnb/mvrx/MavericksViewInternalViewModel.kt
```diff
@@ -0,0 +1,27 @@
+package com.airbnb.mvrx
+
+import androidx.lifecycle.SavedStateHandle
+import androidx.lifecycle.ViewModel
+import java.util.UUID
+import java.util.concurrent.ConcurrentHashMap
+
+class MavericksViewInternalViewModel(state: SavedStateHandle): ViewModel() {
+    val lastDeliveredStates: ConcurrentHashMap<String, Any> = state[PERSISTED_LAST_DELIVERED_STATES_KEY] ?: ConcurrentHashMap<String, Any>().also {
```
elihart, line 9: mark these properties all as internal? It's a bit unfortunate that the class is public, but I see why you had to do that

elihart: is this a new feature you are adding to save the states and subscriptions across process death?

it seems like a good thing to have, but also seems like something we need to be careful about to get right. Are you sure that the ConcurrentHashMap can be saved and retrieved out of the saved state correctly? it should be serializable, but a quick look at the SavedStateHandle source seems like it needs to be a Parcelable

mvrx/src/main/kotlin/com/airbnb/mvrx/MavericksView.kt
elihart: oh, maybe requiring that is needed to disambiguate it from the normal collectLatest

mvrx/src/main/kotlin/com/airbnb/mvrx/MavericksView.kt
elihart: I think we should also add that this is intended to be used with viewmodel.stateflow - are there any issues that arise if this is used with a non mvrx state flow?

mvrx/src/main/kotlin/com/airbnb/mvrx/MavericksView.kt
podarsmarty: hmm it should be fine, the regular one doesnt have defaults.

mvrx/src/main/kotlin/com/airbnb/mvrx/MavericksViewInternalViewModel.kt
podarsmarty: hmm the docs https://developer.android.com/topic/libraries/architecture/viewmodel-savedstate#non-parcelable seem to imply both would work. the perusing the code, validate does check for acceptable type of which serializable is one

mvrx/src/main/kotlin/com/airbnb/mvrx/MavericksView.kt
podarsmarty: hmm i think it would still continue to function unless multiple subscriptions end up clashing. added a comment nonetheless explaining intended use

mvrx/src/main/kotlin/com/airbnb/mvrx/MavericksViewModel.kt
```diff
@@ -416,62 +412,19 @@ abstract class MavericksViewModel<S : MavericksState>(
...
-            this
-        } else if (deliveryMode is UniqueOnly) {
-            val lastDeliveredValue: T? = lastDeliveredValue(deliveryMode)
-            this
-                .assertOneActiveSubscription(lifecycleOwner, deliveryMode)
-                .dropWhile { it == lastDeliveredValue }
-                .flowWhenStarted(lifecycleOwner)
-                .distinctUntilChanged()
-                .onEach { lastDeliveredStates[deliveryMode.subscriptionId] = it }
+        return if (lifecycleOwner != null) {
```
elihart, line 415: would it be worth making `collectLatest(lifecycleOwner,` nullable so we can pass null here and not duplicate this launch logic?

view-binding-utils/src/main/java/com/airbnb/mvrx/viewbinding/FragmentViewBindingDelegate.kt
```diff
@@ -21,16 +24,18 @@ class FragmentViewBindingDelegate<T : ViewBinding>(
...
-            viewLifecycleOwner.lifecycle.addObserver(object : LifecycleObserver {
-                @OnLifecycleEvent(Lifecycle.Event.ON_DESTROY)
-                fun onDestroy() {
-                    // Lifecycle listeners are called before onDestroyView in a Fragment.
-                    // However, we want views to be able to use bindings in onDestroyView
-                    // to do cleanup so we clear the reference one frame later.
-                    clearBindingHandler.post { binding = null }
-                }
-            })
+        GlobalScope.launch(Dispatchers.Main) {
```
elihart, line 27: can you share a github comment here about what this is fixing?

Also, global scope looks suspicious here - can you use the lifecycle of the fragment instead? `fragment.lifecycleScope.launch`

elihart (approved):
@podarsmarty thanks, this looks great! It seems just about ready to merge to me, just one minor questions about consolidating some duplicated code.

Can you update the PR title and summary to reflect what this change has become? Putting an example in the pr summary would be great too.

@gpeal do you have any concerns with this change?

mvrx/src/main/kotlin/com/airbnb/mvrx/MavericksViewModel.kt
podarsmarty: hmm im thinking with collectLatest we would want it the owner be explicitly sent in since it just an extension/not tied to something like a VM which could have fallback logic

view-binding-utils/src/main/java/com/airbnb/mvrx/viewbinding/FragmentViewBindingDelegate.kt
podarsmarty: yea tied to the lifecycle should work. lemme update

podarsmarty: added too description as well

mvrx/src/main/kotlin/com/airbnb/mvrx/MavericksView.kt
```diff
@@ -295,4 +300,33 @@ interface MavericksView : LifecycleOwner {
...
+     * will only be emitted if the value has changed. This is useful for transient views that should only
+     * be shown once (toasts, poptarts), or logging. Most other views should use [RedeliverOnStart], as when a view is destroyed
+     * and recreated the previous state is necessary to recreate the view.
+     *
+     * Use [uniqueOnly] to automatically create a [UniqueOnly] mode with a unique id for this view.
+     *
+     * @param action supports cooperative cancellation. The previous action will be cancelled if it is not completed before
+     * the next one is emitted.
+     */
+    fun <T : Any> Flow<T>.collectLatest(
```
denis-bezrukov, line 320: `<T : Any>` means you can't use it on flows with nullable values. 
so it's impossible to use with the following scenario: 
```
stateFlow.map { it.someNullableProperty }.collectLatest { 
    ....
}

```

Is it intentional? 

previously it was ok because each value (including nullable) was wrapped into `MavericksTuple1`

mvrx/src/main/kotlin/com/airbnb/mvrx/DeliveryMode.kt
```diff
@@ -7,6 +7,8 @@ import kotlin.reflect.KProperty1
  * See: [RedeliverOnStart], [UniqueOnly].
  */
 sealed class DeliveryMode {
+    abstract val subscriptionId: String
```
denis-bezrukov, line 10: as I see you still use `subscriptionId` only for UniqueOnly mode, do you need this change?

view-binding-utils/src/main/java/com/airbnb/mvrx/viewbinding/FragmentViewBindingDelegate.kt
```diff
@@ -21,16 +23,18 @@ class FragmentViewBindingDelegate<T : ViewBinding>(
...
-            viewLifecycleOwner.lifecycle.addObserver(object : LifecycleObserver {
-                @OnLifecycleEvent(Lifecycle.Event.ON_DESTROY)
-                fun onDestroy() {
-                    // Lifecycle listeners are called before onDestroyView in a Fragment.
-                    // However, we want views to be able to use bindings in onDestroyView
-                    // to do cleanup so we clear the reference one frame later.
-                    clearBindingHandler.post { binding = null }
-                }
-            })
+        fragment.lifecycleScope.launch {
```
denis-bezrukov, line 26: why did you need this change?

denis-bezrukov (commented):
`collectLatest` conflicts with an extension from coroutines stdlib. 
Both extensions can be called as `flow.collectLatest { }`, so if you're calling collectLatest from suspending context, it's unclear which one is called.

mvrx/src/main/kotlin/com/airbnb/mvrx/DeliveryMode.kt
podarsmarty: technically not with this current implementation. but i do feel it makes it does make senses for the different delivery modes to provide one.

view-binding-utils/src/main/java/com/airbnb/mvrx/viewbinding/FragmentViewBindingDelegate.kt
podarsmarty: this is the fix mentioned in the description "Also includes a fix for FragmentViewBindingDelegate which can try to observe on a background thread causing a crash. (reproed in sample app launcher)"

mvrx/src/main/kotlin/com/airbnb/mvrx/MavericksView.kt
podarsmarty: yea you wouldn't be able to use it on nullables. everything in the codepath has this assumption. one could still wrap the nullable property into something like MavericksTuple to achieve what is done for you with selectSubscribes right? or is there a usecase you are thinking of that would need this extension to explicitly handle nullable values?

podarsmarty:
> `collectLatest` conflicts with an extension from coroutines stdlib.
> Both extensions can be called as `flow.collectLatest { }`, so if you're calling collectLatest from suspending context, it's unclear which one is called.

open to suggestions for the fun name. had a hard time coming up with one

mvrx/src/main/kotlin/com/airbnb/mvrx/MavericksView.kt
denis-bezrukov: >one could still wrap the nullable property into something like MavericksTuple to achieve what is done for you with selectSubscribes right?

MavericksTuple is internal, so it can't be used, but it can be replaced with Optional/listOf() etc. 
Problem here is that usually it complicates the code too much (adding boxing/unboxing it manually, plus missing this extenssion from auto-complete), we had a lot of similar code with RxJava which doesn't support nullable values, so if there is a way to allow nullable values, that would be better. 

>or is there a usecase you are thinking of that would need this extension to explicitly handle nullable values?

I feel it's confusing to allow
```
viewModel.onEach(State::nullableProp) { prop ->
}
```
but do not allow
```
viewModel.stateFlow
   .map { it.nullableProp }
   .drop(1)
   .collectLatest { prop ->
   }
```

and this one is really look weird: 
```
viewModel.stateFlow
   .map { it.nullableProp }
   .drop(1)
   .map { listOf(it) } 
   .collectLatest { propList ->
       val prop = propList.firstOrNull()
       ...
   }
```

denis-bezrukov:
>open to suggestions for the fun name. had a hard time coming up with one

yep, the best I can think now is `consumeLatest { }`

elihart:
I think the name should be left as `collectLatest` but remove the default value for `deliveryMode` so that there is no conflict with function resolution. Using the same name will be easier for understanding behavior and discoverability.

mvrx/src/main/kotlin/com/airbnb/mvrx/FlowExtensions.kt
```diff
@@ -0,0 +1,73 @@
...
+    lastDeliveredStates: ConcurrentHashMap<String, Any?>,
+    activeSubscriptions: MutableSet<String>,
+    deliveryMode: DeliveryMode,
+    action: suspend (T) -> Unit
+): Job {
+    val flow = when {
+        MavericksTestOverrides.FORCE_DISABLE_LIFECYCLE_AWARE_OBSERVER -> this
+        deliveryMode is UniqueOnly -> {
+            this.assertOneActiveSubscription(lifecycleOwner, activeSubscriptions, deliveryMode.subscriptionId)
+                .dropWhile { it == lastDeliveredStates[deliveryMode.subscriptionId] }
```
denis-bezrukov, line 30: if `lastDeliveredStates[deliveryMode.subscriptionId]  == null` you can't distinguish if it's missing or it's null, would be better to do
```suggestion
                .dropWhile { lastDeliveredStates.contains(deliveryMode.subscriptionId) && it == lastDeliveredStates[deliveryMode.subscriptionId] }
```

elihart: yeah, this was an existing bug, but would be good to fix

denis-bezrukov: previously it did work because `lastDeliveredStates` contained only non-null objects (and all nullable fields were wrapped into MavericksTuple1), so null was the clear indicator the key is missing in the map

podarsmarty: hmm yea you are right. previously every that was wrapped in a mavericks tuple would still work but if not it could behave weird. the suggested change would also catch this as well. updating

PR #562, opened 2021-07-16
Fix typo

erikhuizinga:
The sentence now uses similar language as on line 93:
https://github.com/airbnb/mavericks/blob/2.3.0/docs/core-concepts.md?plain=1#L93

> [...] this will _not_ be run synchronously. [...]


elihart (approved):
Thanks!

PR #565, opened 2021-07-30
Add covariant recreation support

allenchen1154:
Adds support for recreating MavericksViewModels and MavericksState from a persisted Bundle when the declaration site uses a covariant type. This is accomplished by serializing the Java Class of both the ViewModel and State types, and storing them in the StateRestorer class. These values are then used in MavericksFactory and MavericksStateFactory if present, overriding the declaration site types.

Also adds a test in StateRestorationTest.kt to cover this newly supported behavior. Without these changes, the test fails with an IllegalStateException when trying to instantiate the abstract State class from the declaration site.

### DIFF
--- a/mvrx/src/main/kotlin/com/airbnb/mvrx/MavericksViewModelProvider.kt
+++ b/mvrx/src/main/kotlin/com/airbnb/mvrx/MavericksViewModelProvider.kt
@@ -104,7 +104,7 @@ object MavericksViewModelProvider {
         val restoredStateClass = getSerializable(KEY_MVRX_SAVED_STATE_CLASS) as? Class<out S>
 
         requireNotNull(restoredState) { "State was not saved prior to restoring!" }
-        requireNotNull(restoredViewModelClass) { "Viewmodel class was not properly saved prior to restoring!" }
+        requireNotNull(restoredViewModelClass) { "ViewModel class was not properly saved prior to restoring!" }
         requireNotNull(restoredStateClass) { "State class was not properly saved prior to restoring!" }
 
         val restoredContext = when (viewModelContext) {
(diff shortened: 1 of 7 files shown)


allenchen1154:
@BenSchwab @elihart

mvrx/src/main/kotlin/com/airbnb/mvrx/MavericksViewModelProvider.kt
```diff
@@ -91,21 +96,28 @@ object MavericksViewModelProvider {
...
-    private fun <S : MavericksState> Bundle.toStateRestorer(viewModelContext: ViewModelContext): StateRestorer<S> {
+    @Suppress("UNCHECKED_CAST")
+    private fun <VM : MavericksViewModel<S>, S : MavericksState> Bundle.toStateRestorer(viewModelContext: ViewModelContext): StateRestorer<VM, S> {
         val restoredArgs = get(KEY_MVRX_SAVED_ARGS)
         val restoredState = getBundle(KEY_MVRX_SAVED_INSTANCE_STATE)
+        val restoredViewModelClass = getSerializable(KEY_MVRX_SAVED_VM_CLASS) as? Class<out VM>
+        val restoredStateClass = getSerializable(KEY_MVRX_SAVED_STATE_CLASS) as? Class<out S>
 
         requireNotNull(restoredState) { "State was not saved prior to restoring!" }
+        requireNotNull(restoredViewModelClass) { "Viewmodel class was not properly saved prior to restoring!" }
```
BenSchwab, line 107: nit
```suggestion
        requireNotNull(restoredViewModelClass) { "ViewModel class was not properly saved prior to restoring!" }
```

BenSchwab (approved):
LGTM!

For more context, we have a use case where we retrieve a covariant existing view model. MvRx technically doesn't support this safely (need to mark state with `out` variance in many places), but it does work via an unsafe cast, minus that fact in process restoration, if an existing view model that leverage the covariant type is create first, MvRx will fail to recreate the view model, as the VM/S class might be super types of the actual VM/S class.

allenchen1154:
This also fixes the same issue as https://github.com/airbnb/mavericks/pull/556, but without changing the user facing APIs.

PR #566, opened 2021-08-05
Add support for restoring ViewModels that were initially created with a companion factory in a superclass

allenchen1154:
Companion factories may be declared in any superclass of a given Mavericks ViewModel. Therefore, when recreating an instance of a ViewModel, we need to store the ViewModel class from the declaration site, instead of the resolved class. In cases where a companion factory was used, we can call the same one that was used to initially create the ViewModel. If a companion factory was not used, we can safely call the constructor of the original declaration site class as before.

In situations where the State class does not have a default constructor and must be provided via a companion factory, this behavior was leading to a crash. This adds a unit test to `StateRestorationTest` that covers one such scenario.

@BenSchwab

### DIFF
--- a/mvrx/src/main/kotlin/com/airbnb/mvrx/MavericksViewModelProvider.kt
+++ b/mvrx/src/main/kotlin/com/airbnb/mvrx/MavericksViewModelProvider.kt
@@ -66,7 +66,7 @@ object MavericksViewModelProvider {
             // Save the view model's state to the bundle so that it can be used to recreate
             // state across system initiated process death.
             viewModelContext.savedStateRegistry.registerSavedStateProvider(key) {
-                viewModel.viewModel.getSavedStateBundle(restoredContext.args, viewModelClass)
+                viewModel.viewModel.getSavedStateBundle(restoredContext.args, viewModelClass, stateClass)
             }
         } catch (e: IllegalArgumentException) {
             // The view model was already registered with the context. We only want the initial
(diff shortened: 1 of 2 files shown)


mvrx/src/main/kotlin/com/airbnb/mvrx/MavericksFactory.kt
```diff
@@ -43,16 +43,19 @@ private fun <VM : MavericksViewModel<S>, S : MavericksState> createViewModel(
     val viewModelClass = stateRestorer?.viewModelClass ?: declaredViewModelClass
     val stateClass = stateRestorer?.stateClass ?: declaredStateClass
 
-    val factoryViewModel = viewModelClass.factoryCompanion()?.let { factoryClass ->
```
BenSchwab, line 46: Could we first try the declared viewModelClass / StateClass for a no-args constructor?

We then only use the persisted state class in the event the declared class fails. I think it achieves the same result as here, but seems like a bit safe change, as it only changes a case that would have failed before.

BenSchwab: For example, I think this this approach prioritizes the most concrete factory. So if we had BaseVM.Factory and ChildVM.Factory - and BaseVMFactory produces a ChildVM, then in the restore case ChildVM would be called.

Alternatively, what we need to record in the saved state is the VM class that we _actually_ invoked to create the state. 

Stepping back -- the real goal is to record whatever VM/S class was used to create the view model so we can do the same thing in recreate.

BenSchwab (approved):

BenSchwab (approved):

Issue #569, opened 2021-08-27
Koltin 1.5.21 (for Compose 1.0.1)


inktomi:
I've love to help with getting the project onto Compose 1.0.1 rather than on the alpha releases. This would entail also updating to Kotlin 1.5.21 (.30 isn't supported by Compose 1.0.1..).

Would this be something that you'd be open to some help on? I think it'd be a matter of updating the dependency and then fixing any detekt issues that come up, correct? 

I've attempted this locally, and at least running `:mvrx:detekt` didn't report any errors, so if that's the correct command it seems like a straightforward upgrade.

elihart:
sure, help updating dependencies is always appreciated. You can run detekt for the whole project with `./gradlew detekt`

inktomi:
Thank you, I'll try to get that up Monday or early next week. Thank you!

Matt

kaushalyap:
@elihart Is Compose support still experimental?

elihart:
yes, in that it is being used but the api's may change

PR #570, opened 2021-08-30
#569: Updated Compose to latest, and Kotlin to latest compatible version

inktomi:
According to the Mavericks team `./gradlew detekt` should run all the checks. It reports no issues with this change, so I think it's ready to submit upstream.

### DIFF
--- a/buildSrc/src/main/java/dependencies.kt
+++ b/buildSrc/src/main/java/dependencies.kt
@@ -3,19 +3,19 @@ object Versions {
     const val buildTools = "30.0.3"
     const val compileSdk = 30
     const val gradlePlugin = "7.0.0"
-    const val kotlin = "1.4.32"
+    const val kotlin = "1.5.21"
     const val minSdk = 16
     const val targetSdk = 29
 
     // Android libraries
     const val activity = "1.2.3"
-    const val activityCompose = "1.3.0-alpha08"
+    const val activityCompose = "1.3.1"
     const val appcompat = "1.3.0"
     const val arch = "2.1.0"
     const val cardview = "1.0.0"
     const val constraintlayout = "2.0.0"
     const val coordinatorLayout = "1.1.0"
-    const val compose = "1.0.0-beta07"
+    const val compose = "1.0.1"
     const val core = "1.5.0"
     const val fragment = "1.3.4"
     const val lifecycle = "2.3.1"


elihart (approved):
Thanks, looks good. I'll wait  to merge for the kotlin warning fixes you mentioned though

inktomi:
I appreciate it, thank you! I'll push one fix, just for some variable shadowing. I also forgot to update the Kotlin gradle plugin, and the Coroutines lib, so I'll fix those items and anything else that's uncovered. There's some items in `MAvericksLifecycleAwareFlow` for instance. 

The others might be better on a different PR as they're larger refactors or otherwise more odd issues which are preexisting in the code. 

- `MavericksMockPrinter` is using the `AsyncTask.THREAD_POOL_EXECUTOR`, which is deprecated. Perhaps this would be better as a coroutine, or, if not, maybe we could create our own thread pool to avoid the deprecation warning. 
- `BaseMvRxViewModel` has a nullability issue around line 132, where we use a nullable type where it expects a non-null. This one is sort of a head scratcher though, I need to do some more digging to fully understand it.  The full error is "Type mismatch: value of a nullable type (T) -> Async<V>! is used where non-nullable type is expected. This warning will become an error soon. See https://youtrack.jetbrains.com/issue/KT-36770 for details"

elihart:
Travis Ci is currently not running checks on PR for some reason. i need to look into that, and then we can get tests running to check the change, but in general it looks good. thanks for the fixes!

elihart:
@inktomi can you push a new commit? I think I've fixed the travis integration and want to see if tests will run now

inktomi:
I'll look into the build issues, seems like some unit tests and lint perhaps.

elihart:
Thanks. I'll merge once the checks pass

buildSrc/src/main/java/dependencies.kt
```diff
@@ -3,19 +3,19 @@ object Versions {
     const val buildTools = "30.0.3"
     const val compileSdk = 30
     const val gradlePlugin = "7.0.0"
-    const val kotlin = "1.4.32"
+    const val kotlin = "1.5.21"
```
inktomi, line 6: Compose in the released versions don't support Kotlin 1.5.3x yet.

inktomi:
Looks like all we needed was a compatible version of Dagger @elihart. 

There are a bunch of Kotlin warnings that it'd be nice to fix, but I am happy to work on those on another PR (or a few PRs) to make it easier to review since by themselves they're somewhat unrelated to just updating the versions.

elihart (approved):
Great, thanks!

Issue #580, opened 2021-10-05
Crash on Hilts @AndroidEntryPoint Fragment with ComposeView


jakoss:
Maverics `mavericksViewModel()` has a wrong assertion that `LocalContext.current` can be cast to `ComponentActivity`. If we add `@AndroidEntryPoint` to fragment `LocalContext.current` will point to `ContextWrapper` instead, causing crash with message `Composable is not hosted in a ComponentActivity!`

Reproduceable code
```kotlin
@AndroidEntryPoint
class TestFragment : Fragment() {
    override fun onCreateView(
        inflater: LayoutInflater,
        container: ViewGroup?,
        savedInstanceState: Bundle?
    ) = ComposeView(requireContext()).apply {
        setContent {
            // this is null when i add @AndroidEntryPoint to this fragment
            val activity = LocalContext.current as? ComponentActivity
        }
    }
}

```

I think maverics should handle other possible values of `LocalContext` as shown in this SO thread: https://stackoverflow.com/questions/64675386/how-to-get-activity-in-compose

Some more information about the source of the issue: https://stackoverflow.com/questions/64341415/dagger-hilt-android-internal-managers-viewcomponentmanagerfragmentcontextwrappe/64342511

Another non-related question - what's the blocker for mavericks-compose stable release? I'm playing with it for the last 3 months and it works great so far

PR #582, opened 2021-10-07
Fix issue when the LocalContext is not directly an Activity

jakoss:
Fixes: #580

`LocalContext.current` can be a `ContextWrapper` that's not directly an Activity. We have to iterate through `baseContext` to reach proper one.

Also i updated compose and kotlin versions.


elihart (approved):
thanks for the addition 👍

PR #584, opened 2021-10-11
Fix comment, was a copy paste from the method above.

bmarty:


mvrx/src/main/kotlin/com/airbnb/mvrx/MavericksExtensions.kt
```diff
@@ -204,7 +204,7 @@ inline fun <T, reified VM : MavericksViewModel<S>, reified S : MavericksState> T
     }
 
 /**
- * [fragmentViewModel] except scoped to the current Activity. Use this to share state between different Fragments.
+ * ViewModel scoped to the Activity.
```
gpeal, line 207: This isn't much clearer. This sounds like it would be describing `activityViewModel` when this is for view models created inside of the Activity class itself.

bmarty: What do you propose?

mvrx/src/main/kotlin/com/airbnb/mvrx/MavericksExtensions.kt
```diff
@@ -204,7 +204,11 @@ inline fun <T, reified VM : MavericksViewModel<S>, reified S : MavericksState> T
     }
 
 /**
- * ViewModel scoped to the Activity.
+ * /**
```
gpeal, line 207: This isn't formatted correctly

bmarty: updated

PR #586, opened 2021-10-29
update to Compose 1.0.4, Kotlin 1.5.31, Koin 3.1.3

kenyee:
Also gets rid of deprecated jcenter() repo dependency


elihart (approved):
Great, thank you!

PR #589, opened 2021-11-16
Rename master for inclusiveness 

### LINKED ISSUE #587: Move from `master` to `main` or similar for the default branch
The default for new projects on GitHub is a `main` branch, rather than a `master` branch. Though this seems like a simple enough difference, for some `master` can be a racially loaded term. In keeping with [AirBnB's pledge to support inclusiveness](https://news.airbnb.com/an-update-on-our-efforts-to-support-racial-justice/) would you be willing to migrate Mavericks from `master` to something else for the default branch?

inktomi:
This fixes #587 

I searched the repo for `master` and `blacklist`/`whitelist` and adjusted the code as needed. I don't think there is much to do with Travis, but if we can define a branch to pull, it might need to be switched from `master` to `main` once GitHub is updated with the new default. 

My proposal for the next steps would be:

- Merge in this PR to `master`
- Create a new branch called `main` from the latest `master` and push it up
- Check Travis and other third party tools to ensure they're working against `main`
- Delete `master`


docs/core-concepts.md
```diff
@@ -2,7 +2,7 @@
 Mastering Mavericks only requires using three classes: `MavericksState`, `MavericksViewModel`, and `MavericksView`.
 
 ## MavericksState
-The first step in creating a Mavericks screen is to model it as a function of state. The MavericksState interface [doesn't do anything](https://github.com/airbnb/MvRx/blob/master/mavericks/src/main/kotlin/com/airbnb/mvrx/MavericksState.kt) itself but signals the intention of your class to be used as state.
+The first step in creating a Mavericks screen is to model it as a function of state. The MavericksState interface [doesn't do anything](https://github.com/airbnb/MvRx/blob/main/mavericks/src/main/kotlin/com/airbnb/mvrx/MavericksState.kt) itself but signals the intention of your class to be used as state.
```
inktomi, line 5: This URL doesn't actually point to anything. I wasn't sure what to use instead, but I wanted to call that out.

elihart:
Thanks for this, I'll try to follow up with the next steps soon

inktomi:
Sure, happy to work on whatever I can help with. Thank you!

elihart:
also just dropping this here for later https://github.com/github/renaming

github makes it pretty easy to rename

docs/core-concepts.md
nashcft: Current correct URL to `MavericksState` is https://github.com/airbnb/mavericks/blob/master/mvrx/src/main/kotlin/com/airbnb/mvrx/MavericksState.kt, so `https://github.com/airbnb/mavericks/blob/main/mvrx/src/main/kotlin/com/airbnb/mvrx/MavericksState.kt`?

elihart:
rename is complete :)

Issue #573, opened 2021-09-02
Passing an argument in state for compose.


DikarevPavel:
Hi all. At the moment, when navigating from FirstScreen to SecondScreen, 
there is no way to directly pass arguments to mavericksViewModel() or retrieve an argument from the navigation. 
And in order for the viewModel to receive arguments, you need to pass them through "intent.putExtra (args)", 
this is if the navigation is in the MainActivity, and if the navigation is outside the MainActivity, 
then you will have to pass the activity to all navigation in order to use the "intent. putExtra (args) ".
For example:

```
data class TestArg(
    val count: Int = 0,
) : Serializable

data class ExampleState(
    val count: Int = 0,
) : MavericksState {
    constructor(args: TestArg): this(count = args.count)
}

class ExampleViewModel(initialState: ExampleState) : MavericksViewModel<ExampleState>(initialState) {

}

class MainActivity : AppCompatActivity() {

    override fun onCreate(savedInstanceState: Bundle?) {
        super.onCreate(savedInstanceState)
        setContent {
            CounterScreenNavHost(intent)
        }
    }
}


@Composable
fun CounterScreenNavHost(intent: Intent) {
    val navController = rememberNavController()
    NavHost(navController, startDestination = "firstScreen") {
        composable("firstScreen") {
            FirstScreen {
                navController.navigate("secondScreen")
            }
        }
        composable("secondScreen") { stackEntry ->
            val testArg = TestArg(2)
            intent.putExtra(KEY_ARG, testArg) // problem when we have many screens with other args, because its one activity.
            SecondScreen {
                navController.popBackStack()
            }
        }
    }
}

@Composable
fun FirstScreen(navTo: () -> Unit) {
    HcmScaffold(
        content = {
            Box(
                contentAlignment = Alignment.Center,
                modifier = Modifier.padding(it).clickable {
                    navTo()
                }
            ) {
                Text("Example ")
            }
        }
    )
}

@Composable
fun SecondScreen(navBack: () -> Unit) {
    val viewModel = mavericksViewModel<ExampleViewModel, ExampleState>()
    val state = viewModel.collectAsState().value
    HcmScaffold(
        content = {
            Box(
                contentAlignment = Alignment.Center,
                modifier = Modifier.padding(it).clickable {
                    navBack()
                }
            ) {
                Text("Example ${state.count}")
            }
        }
    )
}
```
I believe that in order to solve this problem, we need the ability to pass an optional parameter 
[...]

PR #595, opened 2021-12-15
Add ability to manually pass argument to composable viewModel factories

jakoss:
Added a `argsFactory` parameter to `mavericksViewModel` so user can manually pass argument to view model.

Fixes: https://github.com/airbnb/mavericks/issues/573


mvrx-compose/src/main/kotlin/com/airbnb/mvrx/compose/MavericksComposeExtensions.kt
```diff
@@ -61,16 +64,27 @@ inline fun <reified VM : MavericksViewModel<S>, reified S : MavericksState> mave
...
-        when (scope) {
-            is Fragment -> {
-                val args = scope.arguments?.get(Mavericks.KEY_ARG)
-                FragmentViewModelContext(activity, args, scope)
-            }
-            else -> {
-                val args = activity.intent.extras?.get(Mavericks.KEY_ARG)
-                ActivityViewModelContext(activity, args, viewModelStoreOwner, savedStateRegistry)
+        var view: View? = localView
+        var fragment: Fragment? = null
```
elihart, line 71: I think the try/catch approach is fine, and is maybe preferable to not rely on the implementation detail of the `fragment_container_view_tag` tag. Main thing I would say though is to extract this to a dedicated function and add some documentation and where this pattern was taken from (eg point to the androidx code)

sample-compose/src/main/java/com/airbnb/mvrx/compose/sample/ComposeSampleActivity.kt
```diff
@@ -25,7 +27,32 @@ import com.airbnb.mvrx.compose.mavericksViewModel
 
 data class CounterState(
     val count: Int = 0,
-) : MavericksState
+) : MavericksState {
+    constructor(arguments: ArgumentsTest): this(count = arguments.count)
+}
+
+data class ArgumentsTest(val count: Int) : Parcelable {
```
elihart, line 34: if you're able to add any actual unit tests that would be great, so we can catch regressions automatically in the future

elihart (commented):
Thanks @jakoss, the general approach looks good, just a few requests for cleaning it up

mvrx-compose/src/main/kotlin/com/airbnb/mvrx/compose/MavericksComposeExtensions.kt
jakoss: The issue here is that this whole factory is inline. So any function we want to extract have to be public in na namespace of mavericks

jakoss: I solved that by using `@InternalMavericksApi`, I hope that's enough

sample-compose/src/main/java/com/airbnb/mvrx/compose/sample/ComposeSampleActivity.kt
jakoss: I can add unit tests, but we need whole setup for compose tests and that seems like a work for separate PR. Do you agree with that?

jakoss: I updated the test in sample-compose in this PR. I will work on more generic tests for mvrx-compose

jakoss:
@elihart I've added a test setup for `mvrx-compose` with 2 initial tests - passing arguments via `argsFactory` and passing arguments from the `Fragment.argument`. This should make it for regression tests

mvrx-compose/src/main/kotlin/com/airbnb/mvrx/compose/MavericksComposeExtensions.kt
```diff
@@ -34,25 +39,16 @@ import kotlin.reflect.KProperty1
...
-    } else {
-        while (currentContext is ContextWrapper) {
-            if (currentContext is ComponentActivity) {
-                activity = currentContext
-                break
-            }
-            currentContext = currentContext.baseContext
-        }
-    }
+    val activity = extractActivityFromContext(LocalContext.current)
```
elihart, line 51: thanks

elihart (approved):
looks great, appreciate the contribution. I'll push a release now

mvrx-compose/src/main/kotlin/com/airbnb/mvrx/compose/MavericksComposeExtensions.kt
elihart: @jakoss that's fine. in the future for inline functions you can also make them `internal` and annotate with `@PublishedApi`

PR #600, opened 2022-01-03
Add support to use Mavericks with JUnit 5

itsandreramon:
This PR adds support to use Mavericks in JUnit 5 Unit Tests by adding a MvRxExtension that is compatible with the JUnit 5 [extension system](https://junit.org/junit5/docs/current/user-guide/#extensions), similar to JUnit 4 rules. To avoid code duplication, a new MvRxTestLifecycleCallbacks class is added, which wraps common before & after test code that is shared among the JUnit 5 Extension & JUnit 4 Rule, but can be removed if wanted. If any naming conventions are violated, please let me know.

**Why?**
I want to run my plain Android Unit Tests with JUnit 5, since JUnit 4 is mostly only needed if running integration tests. Since you cannot use the existing [MvRxTestRule](https://github.com/airbnb/mavericks/blob/main/testing/src/main/kotlin/com/airbnb/mvrx/test/MvRxTestRule.kt) in JUnit 5, I tried to copy the source in my own app and migrate it myself. I got it working pretty fast and the migration is straight forward.

**Usage**
```kotlin
import com.airbnb.mvrx.test.MvRxTestExtension
import com.airbnb.mvrx.withState
import org.junit.Assert.assertEquals
import org.junit.jupiter.api.Test
import org.junit.jupiter.api.extension.RegisterExtension

class CounterViewModelTest {

    @Test
    fun testIncrementCount() {
        val viewModel = CounterViewModel(CounterState())
        viewModel.incrementCount()
        withState(viewModel) { state ->
            assertEquals(1, state.count)
        }
    }

    companion object {
        @JvmField
        @RegisterExtension
        val mvrxTestExtension = MvRxTestExtension()
    }
}
```

**Additional information**
To run the Jupiter test one has to add support for it inside the build.gradle of the module. Since the other tests run the tests using JUnit 4, the Counter sample is modified to use the Test Extension. This way, both versions are tested.


itsandreramon:
cc @gpeal @elihart

itsandreramon:
any reason for ignoring this? its kinda disrespectful if someone puts in hours of their free time trying to contribute and you don't even get feedback. @gpeal @elihart

elihart:
sorry for the delay. this will be a great thing to have, we're just busy and will get to it when we can, hopefully before long.

elihart (approved):
@itsandreramon this looks great, thanks for the clean approach with wrapping MvRxTestLifecycleCallbacksImpl. 

Don't have any changes to request, but can you rebase and push again? Since you opened this we switched to github actions for running checks, so it doens't look like they have run on this PR.

I also realized we don't have any info in the docs about the testing rules (https://github.com/airbnb/mavericks/blob/main/docs/_sidebar.md), if you have any interest in adding something there about this it would be great too, but no pressure.

Thanks again, appreciate the contribution 🙏

elihart:
Thanks, Detekt is failing but I'll just merge this and fix the issues myself in a followup for simplicity.

PR #618, opened 2022-02-28
Use passed scope as fragment if possible

jakoss:
In `mavericksViewModel` we are trying to find a parent from a `View` and that works fine for basic use case. But the problem pops up when user wants to pass the fragment directly and it's not the view fragment (for example user is passing parent fragment to reuse view model across the tabs).

This PR fixes that. If passed `scope` is Fragment - let's just use that instead of searching for it in view tree


jakoss:
@elihart This is related to https://github.com/airbnb/mavericks/pull/595, somehow i missed that case

elihart (approved):
thanks!

PR #633, opened 2022-06-08
[Mavericks] Add functions to enable number mock view models to ten

ruishi01:
Unblock integration test for fragment with more than seven view models.


elihart (approved):
thanks, lgtm

Issue #632, opened 2022-06-08
Mavericks as generic purpose state management


sav007:
I know this feature request might be a bit of stretch but want to start a conversation.

Mavericks showed a great result as state management in Android VM world but looks like its purpose can be extended beyond it. What if Mavericks becomes a generic purpose state management where it can be used in pure Kotlin world? Like extract the engine that has 0 dependencies to Android artifacts and provide extension to connect it with Android VM?

There is a potential that Mavericks can be used anywhere where stateful component is needed (like repository that requires to have a state). Also having separate pure Kotlin engine will make it possible to use it outside of VM like it was mentioned in this [issue](https://github.com/airbnb/mavericks/issues/605) and make it possible to bring it to iOS by making it [KMP](https://github.com/airbnb/mavericks/issues/558)

I know these questions have been already asked before but has anyone done any assessment on the effort required for this kind of work? Are you open for such transformation? Do you think it makes sense and the time is right?

elihart:
I agree with the sentiment and would love to see a general purpose multiplatform solution, but have not looked into feasibility or effort required. It's also unlikely I would be able to work on this, so it would need to come from external contributors. I would be open reviewing changes though, but it would need to be well planned out.

As far as whether now is the right time, I think it depends on the state of KMM tooling. I haven't used it myself so it is hard for me to say. I want to avoid making the repository hard to maintain due to KMM APIs changing/breaking overtime and configuration nightmares.

sav007:
Just to clarify that this feature request is not primarily about KMP, but rather extract the core of Mavericks ( state management) as pure Kotlin module and make VM integration as an extension.

KMP is just one of side effects of such transformation from other applications, where state management is required.

I think this process can be broken into 2 phases: first will be extracting core part and make it 0 dependent on Android platform, build extension for VM integration. And only after this second phase is making core as KMP module.

Even implementing only first phase I think provide more flexibility of using Mavericks outside of Android VM.

elihart:
Gotcha, I see that a main use case is to use mavericks in a repository pattern outside of a UI scope.

Fwiw, we already use our viewmodels as Singletons using this pattern
```
/**
 * Singletons in the app that are not tied to an Activity/Fragment lifecycle should implement this.
 *
 * This allows our singletons to manage their State using the same tools that our UI ViewModels do.
 *
 * For further documentation, see https://airbnb.quip.com/uZtMAaM5dpcR/Android-patterns-for-singletons-and-data-repositories.
 */
[...]

PR #635, opened 2022-06-13
Introduce Mavericks core module

sav007:
Move everything that could be Android platform agnostic into the core module.

Part of https://github.com/airbnb/mavericks/issues/632


mvrx-core/src/main/java/com/airbnb/mvrx/MavericksStateModel.kt
```diff
@@ -0,0 +1,407 @@
...
+            // Use yield to ensure flow collect coroutine is dispatched rather than invoked immediately.
+            // This is necessary when Dispatchers.Main.immediate is used in scope.
+            // Coroutine is launched with start = CoroutineStart.UNDISPATCHED to perform dispatch only once.
+            yield()
+            collectLatest(action)
+        }
+    }
+
+    private fun <S : MavericksState> assertSubscribeToDifferentViewModel(viewModel: MavericksStateModel<S>) {
+        require(this != viewModel) {
```
elihart, line 401:
```suggestion
        require(this !== viewModel) {
```

mvrx-core/src/main/java/com/airbnb/mvrx/MavericksMutabilityHelper.kt
```diff
@@ -30,6 +25,16 @@ fun assertMavericksDataClassImmutability(
 ) {
     require(kClass.java.isData) { "Mavericks state must be a data class! - ${kClass.simpleName}" }
 
+    val disallowedFieldCollectionTypes = listOfNotNull(
```
elihart, line 28: seems like a good solution 👍

mvrx-core/src/main/java/com/airbnb/mvrx/MavericksRepository.kt
```diff
@@ -0,0 +1,419 @@
...
+     *
+     * Mocking out the state store cannot accomplish this on its own, because in some cases we may
+     * want the state store to initially be mocked, with state changes blocked, but later on we may
+     * want it to allow state changes.
+     *
+     * This prevents the case of an executed async call from modifying state once the state stored
+     * is "enabled", even if the execute was performed when the state store was "disabled" and we
+     * didn't intend to allow operations to change the state.
+     */
+    protected open fun <S : MavericksState> onExecute(
```
elihart, line 277: I don't follow this change. To maintain previous behavior, doesn't this need to call `config.onExecute`?

why open this up for override?

elihart (commented):
Looking really good! mainly just have one question

mvrx-core/src/main/java/com/airbnb/mvrx/MavericksRepository.kt
sav007: My bad, left from couple iterations of refactoring. Addressed in https://github.com/airbnb/mavericks/pull/635/commits/3f768692c00a06d0030eece840aea79d0f5eeddb

mvrx/src/main/kotlin/com/airbnb/mvrx/MavericksViewModel.kt
```diff
@@ -34,402 +18,46 @@ import kotlin.reflect.KProperty1
...
-    @PublishedApi
-    internal val configFactory = Mavericks.viewModelConfigFactory
-
-    @Suppress("LeakingThis")
-    @InternalMavericksApi
-    val config: MavericksViewModelConfig<S> = configFactory.provideConfig(
-        this,
-        initialState
-    )
+    val viewModelScope = coroutineScope
```
gpeal, line 32: `coroutineScope` also gets inherited by children now, too, right? If so, let's see if we can figure out how to hide that. One idea that would take more code but would work is to have this wrap and delegate to the repo rather than extend it.

sav007: I thought about using delegate vs inheritance over here, it's for sure will introduce some API duplication but let me check how bad is that

sav007: Actually I think this is very good idea, as it feels more right that VM has reference to repository vs extends it. Even if it requires some boilerplate glue code this looks more correct.

sav007: Addressed in https://github.com/airbnb/mavericks/pull/635/commits/7566758ec578ed15bb8ac7534c37f58ef674eff0

sav007:
@elihart fixed lint issue

PR #639, opened 2022-07-07
Mavericks Extensions argsOrNull

sanogueralorenzo:
Adding Mavericks Extensions argsOrNull which can be useful for those cases where you have an optional argument that might be sent or not.

As an example, imagine a shared screen / flow from different entry points. While this can be fixed in multiple ways in the client's code having an explicit enum that represents null / empty / none it ends up being more code which ultimately represents the same thing (although this introduces the option of nullable arguments which can also be seen as a downside)

Thoughts?

### DIFF
--- a/mvrx/src/main/kotlin/com/airbnb/mvrx/MavericksExtensions.kt
+++ b/mvrx/src/main/kotlin/com/airbnb/mvrx/MavericksExtensions.kt
@@ -247,6 +247,31 @@ fun <V : Any> args() = object : ReadOnlyProperty<Fragment, V> {
 }
 
 /**
+ * Fragment argument delegate that makes it possible to set fragment args without
+ * creating a key for each one.
+ *
+ * To create nullable arguments, define a property in your fragment like:
+ *     `private val listingId by argsOrNull<MyArgs?>()`
+ *
+ * Each fragment can only have a single argument with the key [Mavericks.KEY_ARG]
+ */
+fun <V : Any> argsOrNull() = object : ReadOnlyProperty<Fragment, V?> {
+    var value: V? = null
+    var read: Boolean = false
+
+    override fun getValue(thisRef: Fragment, property: KProperty<*>): V? {
+        if (!read) {
+            val args = thisRef.arguments
+            val argUntyped = args?.get(Mavericks.KEY_ARG)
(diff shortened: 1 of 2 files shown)


mvrx/src/main/kotlin/com/airbnb/mvrx/MavericksExtensions.kt
```diff
@@ -246,6 +246,31 @@ fun <V : Any> args() = object : ReadOnlyProperty<Fragment, V> {
     }
 }
 
+/**
+ * Fragment argument delegate that makes it possible to set fragment args without
+ * creating a key for each one.
+ *
+ * To create nullable arguments, define a property in your fragment like:
+ *     `private val listingId by argsOrNull<MyArgs?>()`
```
gpeal, line 254:
```suggestion
 *     `private val listingId: MyArgs? by argsOrNull()`
```
Can you update the doc above, too?

mvrx/src/main/kotlin/com/airbnb/mvrx/MavericksExtensions.kt
```diff
@@ -246,6 +246,31 @@ fun <V : Any> args() = object : ReadOnlyProperty<Fragment, V> {
...
+ *
+ * Each fragment can only have a single argument with the key [Mavericks.KEY_ARG]
+ */
+fun <V : Any> argsOrNull() = object : ReadOnlyProperty<Fragment, V?> {
```
gpeal, line 258:
```suggestion
fun <V> argsOrNull() = object : ReadOnlyProperty<Fragment, V?> {
```
I _think_ you can safely emit the type narrowing since the return type is nullable.

gpeal (commented):
@elihart This seems like a reasonable addition to me. Wdyt?

sanogueralorenzo:
@elihart what do you think of `by argsOrNull`?

elihart:
sure, seems helpful. only comment is that it would be nice to have a test

sanogueralorenzo:
Sounds good @elihart added tests https://github.com/airbnb/mavericks/pull/639/commits/cde19593968347d92d7a5fd31d13cd103951ee4d

elihart (approved):
Thanks!

PR #647, opened 2022-08-11
[Compose] Don't use Flow operators in a Composable

gpeal:
This fixes a new lint error. The issue is outlined in [this blog post](https://jetc.dev/slack/2022-01-15-where-flow-operators-go.html).


mvrx-compose/src/main/kotlin/com/airbnb/mvrx/compose/MavericksComposeExtensions.kt
```diff
@@ -143,10 +144,10 @@ fun <VM : MavericksViewModel<S>, S : MavericksState> VM.collectAsState(): State<
  * Prefer the overload with a state property reference to ensure that your composable only recomposes when the properties it uses changes.
  */
 @Composable
-@SuppressLint("FlowOperatorInvokedInComposition")
 fun <VM : MavericksViewModel<S>, S : MavericksState, O> VM.collectAsState(mapper: (S) -> O): State<O> {
-    // TODO(gpeal) fix this lint error.
-    return stateFlow.map { mapper(it) }.distinctUntilChanged().collectAsState(initial = withState(this) { mapper(it) })
+    val updatedMapper by rememberUpdatedState(mapper)
+    val mappedFlow = remember { stateFlow.map { updatedMapper(it) }.distinctUntilChanged() }
```
elihart, line 149: this behavior is a little interesting i think. afaict if the mapper changes it won't immediately produce a new value in the flow corresponding to the new mapping, it will only affect the next emitted value.

that's potentially confusing to end users, might make sense to have it re-emit the last value with the new mapping

elihart (approved):
thanks for fixing

mvrx-compose/src/main/kotlin/com/airbnb/mvrx/compose/MavericksComposeExtensions.kt
sav007: I might be missing a context here but why not just remember `mapper` lambda and provide API to specify keys this remember call should be depended on? Like the signature of `remember` function with keys.
It will be up to the user to ensure that `collectAsState` will be recomposed when provided key changed.

Another approach is just leave it to the user to ensure that provided lambda is not causing over recomposition, I think this is the well known and common pattern in compose to be careful with passed lambdas.

gpeal: I was thinking about having it be key based but was curious if there was some state snapshot capturing way to do it automatically

sav007: From what I can see the idea is that `mappedFlow` should be recomposed when `mapper` lambda is changed and skipped when it remains the same. To handle this I don't think there is any other way than remember `mappedFlow` with keys when invalidate the cached value (keys that signals cached value for provided mapper is invalid).

I think we should just do following:
```
val mappedFlow = remember(mapper) { stateFlow.map { updatedMapper(it) }.distinctUntilChanged() }
return mappedFlow.collectAsState(initial = withState(this) { updatedMapper(it) })
```

and delegate skipping unnecessary recomposition to user by providing remembered instance of `mapper`. As we don't really know under what conditions (keys if any) `mapper` should be remembered

gpeal: Did you mean to do `remember(key) { … }`?

sav007: I mean:
```
val mappedFlow = remember(mapper) { stateFlow.map { updatedMapper(it) }.distinctUntilChanged() } return mappedFlow.collectAsState(initial = withState(this) { updatedMapper(it) })
```

so it up to the user make `mapper` stable either to pass function reference or on user side do `remember { {} }`

gpeal: I think the most common usage would be:
```
viewModel.collectAsState { it.count }
```

which would cause it to change on every composition, right?

sav007: I don't think it's going to be recomposed, as I've read somewhere that compose is smart to skip recomposition if passed lambda is stable (compiler infers that from analyzing captured variables) and if it's stable it auto remembers it. So it will depend on the lambda implementation. In your example `it` is the VM state that I think should be stable in most case scenarios. If that not true the user can do `val mapper = remember { {}}` and pass it to `collectAsState`

mvrx-compose/src/main/kotlin/com/airbnb/mvrx/compose/MavericksComposeExtensions.kt
```diff
@@ -141,12 +142,17 @@ fun <VM : MavericksViewModel<S>, S : MavericksState> VM.collectAsState(): State<
...
+ *            If your mapper changes (for example, reading a different state property) then, by default, you won't receive an updated state value
+ *            until either the ViewModel emits a new state or if you change the key.
+ *            This is analogous to `remember(key) { … }`.
  */
 @Composable
-@SuppressLint("FlowOperatorInvokedInComposition")
-fun <VM : MavericksViewModel<S>, S : MavericksState, O> VM.collectAsState(mapper: (S) -> O): State<O> {
-    // TODO(gpeal) fix this lint error.
-    return stateFlow.map { mapper(it) }.distinctUntilChanged().collectAsState(initial = withState(this) { mapper(it) })
+fun <VM : MavericksViewModel<S>, S : MavericksState, O> VM.collectAsState(key: Any? = Unit, mapper: (S) -> O): State<O> {
```
gpeal, line 152: @sav007 @elihart I added an optional key here.

elihart (approved):
thanks, lgtm

PR #648, opened 2022-08-16
Renamed MvRxTestRule, Extension, and a few other comments and docs to Mavericks

gpeal:
We still had a number of mvrx references. This PR attempts to resolve the missing ones prior to 3.0.

Instead of making the junit rule and extension open, I copied the impl since they don't change frequently and we'll be able to remove the old one later.


elihart (approved):
thanks for the clean up 🙏

PR #649, opened 2022-08-16
Renamed gradle modules for consistency

gpeal:
All modules are now grouped with a prefix:
* mvrx: official artifact
* sample: sample app
* utils: a collection of utilities

No structural or other changes were done here other than simple renames. You can view `settings.gradle` for the clearest diff.

I went with dash-separated naming because some modules and artifacts already used that but I've seen dots or colons used in the past as well. I don't have a strong opinion but opted for consistency with existing modules here.


elihart (approved):
nice consistency improvement, thanks!

PR #652, opened 2022-09-23
Created an Anvil sample

gpeal:
This PR creates a sample app that uses Anvil and has a `@ContributesViewModel` code generator to let Anvil do all of the wiring necessary to do constructor injection for ViewModels. It also demonstrates how to handle Dagger subcomponents including those owned by Fragments/features.


gpeal:
@denis-bezrukov @romankivalin  FYI

elihart (approved):
Cool, this is a great sample to have

PR #661, opened 2022-10-31
Passing the arguments of a fragment if `mavericks:arg` is not available

rohithThammaiah:
This will be helpful for those who are using the fragment navigation library for deeplink since they match the argument key with the query or path parameter


rohithThammaiah:
I have also updated the `sample-dogs` sample to show the usage of this

PR #666, opened 2022-12-09
Optimize lifecyle aware flow collection

sav007:
Using androidx `Lifecycle.whenStarted` inside the flow collector creates new instance of dispatcher, queue, controller on every emission. To avoid this create `PausingDispatcher` outside of flow collector and reuse it to wrap action execution. Implementation of `PausingDispatcher` was copied without modification from AOSP, as it's non public API.

Follow up of https://github.com/airbnb/mavericks/pull/665

cc: @denis-bezrukov


sav007:
Oh my, nice number for this PR 👿👿👿

mvrx/src/main/kotlin/com/airbnb/mvrx/FlowExtensions.kt
```diff
@@ -42,11 +42,14 @@ internal fun <T : Any?> Flow<T>.collectLatest(
...
-        flow.collectLatest {
-            if (MavericksTestOverrides.FORCE_DISABLE_LIFECYCLE_AWARE_OBSERVER) {
-                action(it)
-            } else {
-                lifecycleOwner.whenStarted { action(it) }
+        if (MavericksTestOverrides.FORCE_DISABLE_LIFECYCLE_AWARE_OBSERVER) {
+            flow.collectLatest { action(it) }
+        } else {
+            val lifeCycleAwareDispatcher = lifecycleOwner.lifecycle.lifeCycleAwareDispatcher()
+            flow.collectLatest { item ->
```
denis-bezrukov, line 49: I feel there is one interesting consequence of this change... 🤔 before, the onEach block was cooperative to cancellation. That means a consumer should decide whether they want or don't want to allow cancellation for the block. 
e.g. 
```
val items = mutableList<Int>()
viewModel.onEach(State::someItem) { item ->
   items.add(item)
}
```
In this example, onEach's body doesn't support cooperative cancellation - the whole body is blocking. If VM emits 1,2,3,4,5. `items` will contain all of them. 

Now, the problem is withContext checks for cancellation, and won't execute the block if it was cancelled (in case if a new item will appear). So effectively it's an equivalent of
```
val items = mutableList<Int>()
viewModel.onEach(State::someItem) { item ->
   currentCoroutineContext.ensureActive()
   items.add(item)
}
```
That means if VM emits 1,2,3,4,5 the behavior is undefined - your `items` may contain a subsequence of these items, e.g. `1,5` or `5` or `1,2,3,4,5`. 

I think it's a huge behavior change that may break a lot of things. IIRC @gpeal was saying they had "conflation" in one of recent versions, but then intentionally avoided it. 

Actually, the same problem is here for pure `lifecycleOwner.whenStarted` call.

I'm trying to think how to change that behavior - outer withContext will definitely help, but it might affect the behavior when lifecycle is paused. Anyway we need a test for this (maybe we have that one? and it somehow passed due to "no dispatch" behavior or smth like that)

sav007: well you are right about the changing behaviour, but imo this behaviour is more correct as why even bother UI layer with the intermediate state if it's not relevant anymore in case of slow collector?

Wrapping the whole flow collection with paused dispatcher will break `DeliveryMode` I think, values emitted during paused won't be recorded

sav007: hm, shouldn't UI reset the state when it handles the state rather than VM take a guess if event was handled or not.

But anyway I agree with you about changing compatibility. I will try to think how to address it.

sav007: looks like wrapping the whole collector into `whenStarted` just works.

sav007:
Closing this PR as there is a major drawback with this implementation.
VM state is shared flow, if we pause entire collector for UI it will suspend any state updates when buffer limit reached and no new state will be emitted until UI collector resumed.

Issue #677, opened 2023-05-05
possible incompatibility with Kotlin coroutines 1.7.0


kenyee:
Using .flowWhenStarted() now throws this error:

```

java.lang.IllegalStateException: Flow invariant is violated:
  Emission from another coroutine is detected.
  Child of StandaloneCoroutine{Active}@ab35420, expected child of ScopeCoroutine{Active}@2bcc1d9.
  FlowCollector is not thread-safe and concurrent emissions are prohibited.
  To mitigate this restriction please use 'channelFlow' builder instead of 'flow'
 at kotlinx.coroutines.flow.internal.SafeCollector_commonKt$checkContext$result$1.invoke(SafeCollector.common.kt:68)
 at kotlinx.coroutines.flow.internal.SafeCollector_commonKt$checkContext$result$1.invoke(SafeCollector.common.kt:25)
 at kotlin.coroutines.CoroutineContext$Element$DefaultImpls.fold(CoroutineContext.kt:70)
 at kotlinx.coroutines.Job$DefaultImpls.fold(Job.kt:112)
 at kotlinx.coroutines.JobSupport.fold(JobSupport.kt:25)
 at kotlin.coroutines.CombinedContext.fold(CoroutineContextImpl.kt:131)
 at kotlinx.coroutines.flow.internal.SafeCollector_commonKt.checkContext(SafeCollector.common.kt:25)
 at kotlinx.coroutines.flow.internal.SafeCollector.checkContext(SafeCollector.kt:106)
 at kotlinx.coroutines.flow.internal.SafeCollector.emit(SafeCollector.kt:83)
 at kotlinx.coroutines.flow.internal.SafeCollector.emit(SafeCollector.kt:66)
 at com.airbnb.mvrx.MavericksLifecycleAwareFlowKt$flowWhenStarted$1$1$transform$1.invokeSuspend(MavericksLifecycleAwareFlow.kt:36)
```

This is Mavericks version 3.0.1

kenyee:
Also tried latest 3.0.2 and it has the same issue

elihart:
Thanks for reporting. We don't currently plant to update coroutines support in the near future, but if someone wants to contribute a PR that'd be great

ArunYogeshwaran:
Thanks a ton for this fix 🚀

PR #678, opened 2023-05-11
Changed `flowWhenStarted` to emit value on flow coroutine context

vulpeszerda:
Fixes: https://github.com/airbnb/mavericks/issues/677

1. Changed `flowWhenStarted` implmentation to emit value on original flow's context.
2. Fixed flow value is not emitted properly when flow's type `T` is nullable

### DIFF
--- a/mvrx/src/main/kotlin/com/airbnb/mvrx/MavericksLifecycleAwareFlow.kt
+++ b/mvrx/src/main/kotlin/com/airbnb/mvrx/MavericksLifecycleAwareFlow.kt
@@ -10,7 +10,6 @@ import kotlinx.coroutines.channels.ReceiveChannel
 import kotlinx.coroutines.channels.produce
 import kotlinx.coroutines.coroutineScope
 import kotlinx.coroutines.flow.Flow
-import kotlinx.coroutines.flow.collect
 import kotlinx.coroutines.flow.combineTransform
 import kotlinx.coroutines.flow.flow
 import kotlinx.coroutines.selects.SelectBuilder
@@ -31,31 +30,32 @@ fun <T : Any?> Flow<T>.flowWhenStarted(owner: LifecycleOwner): Flow<T> = flow {
         val startedChannel = startedChannel(owner.lifecycle)
         val flowChannel = produce { collect { send(it) } }
 
-        val transform: suspend (Boolean, T) -> Unit = { started, value ->
-            if (started) {
-                emit(value)
-            }
-        }
-
         var started: Boolean? = null
-        var flowValue: T? = null
+        var flowResult: FlowResult<T>? = null
         var isClosed = false
 
         while (!isClosed) {
-            select<Unit> {
-                onReceive(startedChannel, { flowChannel.cancel(); isClosed = true }) {
-                    started = it
-                    if (flowValue !== null) {
-                        @Suppress("UNCHECKED_CAST")
-                        transform(it, flowValue as T)
+            val result: SelectResult<T>? = select {
+                onReceive(startedChannel, { flowChannel.cancel(); isClosed = true }) { value ->
+                    started = value
+                    val result = flowResult
+                    if (result != null && value) {
+                        SelectResult(result.payload)
+                    } else {
+                        null
                     }
                 }
-                onReceive(flowChannel, { isClosed = true }) {
-                    flowValue = it
-                    if (started !== null) {
-                        transform(started as Boolean, it)
+                onReceive(flowChannel, { isClosed = true }) { value ->
+                    flowResult = FlowResult(value)
+                    if (started == true) {
+                        SelectResult(value)
+                    } else {
+                        null
                     }
                 }
+            }
+            if (result != null) {
+                emit(result.payload)
             }
         }
     }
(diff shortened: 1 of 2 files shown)


elihart:
Thanks for contributing a fix! This looks pretty good to me, and thank you for catching the null result bug. Let's just make sure the tests pass before merging

carlosmuvi-stripe:
Seems like all tests passed right? 🙂

mvrx/src/main/kotlin/com/airbnb/mvrx/MavericksLifecycleAwareFlow.kt
```diff
@@ -83,14 +83,22 @@ private fun startedChannel(owner: Lifecycle): Channel<Boolean> {
...
+            null
+        } else {
+            onReceive(result)
+        }
     }
 }
+
+private data class SelectResult<T : Any?>(val payload: T)
+
+private data class FlowResult<T : Any?>(val payload: T)
```
denis-bezrukov, line 104: For performance reasons it makes sense to switch to a token (see how coroutines team [uses](https://github.com/Kotlin/kotlinx.coroutines/blob/master/kotlinx-coroutines-core/common/src/flow/operators/Transform.kt#LL126C24-L126C24) it for example)

In this case it will be:
```
val NULL = Any()

var flowResult: Any? = NULL
var isClosed = false

val result = select {
    onReceive(startedChannel...) { value -> 
        started = value
        val result = flowResult
        if (started && result != NULL) {
           result
        } else { 
           NULL
        }
    }
    onReceive(flowChannel....) { value ->
       flowResult = value
        if (started && result != NULL) {
           flowResult
        } else { 
           NULL
        }
      }
if (result != NULL) {
   emit(result as T?)
}       
```

Or at least it doesn't need to be `data` class

vulpeszerda: Thanks for comment! 
I applied this in [5425029](https://github.com/airbnb/mavericks/pull/678/commits/5425029ed54890d9604b99f29c01e7e67550d140) 👍

mvrx/src/main/kotlin/com/airbnb/mvrx/MavericksLifecycleAwareFlow.kt
```diff
@@ -83,14 +85,17 @@ private fun startedChannel(owner: Lifecycle): Channel<Boolean> {
...
-    crossinline onClosed: () -> Unit,
-    noinline onReceive: suspend (value: T) -> Unit
+    crossinline onClosed: () -> R,
+    noinline onReceive: suspend (value: T) -> R
 ) {
     channel.onReceiveCatching {
         val result = it.getOrNull()
-        if (result === null) onClosed()
-        else onReceive(result)
+        if (result === null) {
```
denis-bezrukov, line 95: It seems null from upstream flow will be treated as closed event (in both wrapper/token impls), so it won't work as expected in case if flow emits null values. You can switch to isSuccess/isFailure/isClosed. 
```
if (it.isFailure) {
   onClosed()
} else {
   onReceive(it.getOrThrow())
}
```

Could you also please add a unit test to `MavericksLifecycleAwareFlowKtTest`, smth like this (it should fail on main/current pr):
```
    @Test
    fun testNullableFlow() = runTest(UnconfinedTestDispatcher()) {
        val flow = flowOf<Int?>(null, null)
        val owner = TestLifecycleOwner()
        val values = mutableListOf<Int?>()
        owner.lifecycle.currentState = Lifecycle.State.STARTED
        val job = flow.flowWhenStarted(owner).onEach {
            values += it
        }.launchIn(this)
        assertEquals(listOf(null, null), values)
        job.cancel()
    }
```

elihart: @vulpeszerda let me know when you get to the above ^ and then I will merge. thanks!

vulpeszerda: @denis-bezrukov Thanks for suggestion! 
I realized importance of test code 😓  my bad. 

BTW, how about checking result with `isClosed` instead of `isFailure`? 
Of cource It is not so meaningful in this code, because source channel will not fail in all case (if source channel fail, exception will be thrown in `produce { collect { send(it) } }`) ,but I think handling failure case in `getOrThrow` phrase is much readable.

vulpeszerda: @denis-bezrukov And your test code works great! 👍

denis-bezrukov: @vulpeszerda  Yep, since failure isn't possible I agree it's better to just use isClosed

denis-bezrukov (approved):
LGTM 🚀

mvrx/src/main/kotlin/com/airbnb/mvrx/MavericksLifecycleAwareFlow.kt
vulpeszerda: @denis-bezrukov @elihart I applied above suggestions

elihart:
Thanks! looks great

elihart:
I'll make a new release with this change this week sometime

Issue #352, opened 2020-02-20
is any plan to support RxJava3?


Dugin123:
is any plan to support RxJava3?

gpeal:
@BenSchwab @elihart This might be a good opportunity to the coroutines defualt + rxjava2  and rxjava3 extension libraries. What do you think?

Dugin123:
I prefer RxJava than coroutines. I think RxJava is more powerful and fitable functional programming.
  what do you think below structure? 
core+rxjava3
core+rxjava2
core+coroutines?

Anyway, regardless of module structure, RxJava3 has released official ver, so I hope it will be support.

elihart:
Yes, I think this would tie in nicely to split extension libraries for these options. I would think that an extension library for rxjava3 would be pretty simple once we have one for rxjava2.

jpventura:
@Dugin123 RxJava 2.x officially is [at maintenance mode](https://github.com/ReactiveX/RxJava/wiki/What's-different-in-3.0#2x-now-in-maintenance-mode).

Thus we could piggyback coroutines refactor to simply offer 3.x built over extensions when [JetBrains officially support it](https://github.com/Kotlin/kotlinx.coroutines/tree/master/reactive/kotlinx-coroutines-rx2)  (see #347).

Dugin123:
@jpventura 
Now JetBrains  support [kotlinx-coroutines-rx3](https://github.com/Kotlin/kotlinx.coroutines/tree/master/reactive/kotlinx-coroutines-rx3).

renaudcerrato:
RxJava2 is in maintenance mode and support will stop on March 1st 2021 ([source](https://github.com/ReactiveX/RxJava#version-2x)). Migrating is easy, will you accept a PR ?

elihart:
Yes, if you look at the 2.0 release branch (https://github.com/airbnb/MvRx/tree/release/2.0.0) there is a `mvrx-rxjava2` module. I think we would be open to a PR that adds an equivalent module for rxjava3. I imagine it would be pretty straightforward to mirror what is done in the rxjava2 module

JagadeeshDamarasingu:
hi @gpeal @elihart , I have opened a new PR #685 for this issue, please review

PR #685, opened 2023-06-08
RxJava3 support

JagadeeshDamarasingu:
Satisfies [352](https://github.com/airbnb/mavericks/issues/352)
Adds Support for RxJava3


mvrx-rxjava3/gradle.properties
```diff
@@ -0,0 +1,4 @@
+POM_NAME=Mavericks
+POM_ARTIFACT_ID=mavericks-rxjava2
```
elihart, line 2: wrong artifact id

.idea/kotlinc.xml
```diff
@@ -0,0 +1,6 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project version="4">
```
elihart, line 2: don't add this file. you can git ignore it

.idea/codeStyles/Project.xml
```diff
@@ -1,11 +1,6 @@
 <component name="ProjectCodeStyleConfiguration">
   <code_scheme name="Project" version="173">
     <JetCodeStyleSettings>
-      <option name="PACKAGES_TO_USE_STAR_IMPORTS">
```
elihart, line 4: don't change this

gradle/libs.versions.toml
```diff
@@ -59,12 +60,17 @@ picasso = "com.squareup.picasso:picasso:_"
 recyclerview = "androidx.recyclerview:recyclerview:_"
 retrofit = "com.squareup.retrofit2:retrofit:_"
 retrofitMoshi = "com.squareup.retrofit2:converter-moshi:_"
-retrofitRxJava = "com.squareup.retrofit2:adapter-rxjava2:_"
+retrofitRxJava2 = "com.squareup.retrofit2:adapter-rxjava2:_"
+retrofitRxJava3 = "com.squareup.retrofit2:adapter-rxjava3:_"
+coil = "io.coil-kt:coil:_"
```
elihart, line 65: why add coil?

mvrx-rxjava3/src/main/AndroidManifest.xml
```diff
@@ -0,0 +1,5 @@
+<manifest package="com.airbnb.mvrx.rxjava2">
```
elihart, line 1: i don't think this is needed. declare the package name in the build.gradle file via the `namespace` property

mvrx-testing/build.gradle
```diff
@@ -37,8 +37,8 @@ dependencies {
     implementation libs.junit
     implementation libs.junit5
     implementation libs.kotlinReflect
-    implementation libs.rxAndroid
+    implementation libs.rxAndroid3
     implementation libs.multidex
     implementation libs.kotlinCoroutinesTest
-    implementation project(":mvrx-rxjava2")
+    implementation project(":mvrx-rxjava3")
```
elihart, line 43: can you leave this as rxjava 2? i want to avoid changing published artifacts

docs/rxjava3.md
```diff
@@ -0,0 +1,11 @@
+# Mavericks + RxJava3
+
+[//]: # (TODO refine the documentation)
```
elihart, line 3: are you planning to address this?

mvrx-rxjava3/src/main/kotlin/com/airbnb/mvrx/MvRx.kt
```diff
@@ -0,0 +1,17 @@
+package com.airbnb.mvrx
+
+/**
+ * Exists for backwards compatibility.
+ *
+ * @see Mavericks
+ */
+object MvRx {
```
elihart, line 8: this  shouldn't be duplicated

mvrx-rxjava3/src/main/kotlin/com/airbnb/mvrx/MvRxState.kt
```diff
@@ -0,0 +1,9 @@
+package com.airbnb.mvrx
+
+/**
+ * MvRx state exists solely for MvRx 1.x backwards compatibility.
+ * [MavericksState] is a drop in replacement going forward.
+ *
+ * @see MavericksState
+ */
+interface MvRxState : MavericksState
```
elihart, line 9: this shouldn't be duplicated

mvrx-rxjava3/src/main/kotlin/com/airbnb/mvrx/BaseMvRxViewModel.kt
```diff
@@ -0,0 +1,464 @@
...
+import kotlinx.coroutines.flow.distinctUntilChanged
+import kotlinx.coroutines.flow.launchIn
+import kotlinx.coroutines.flow.map
+import kotlinx.coroutines.flow.onEach
+import kotlin.reflect.KProperty1
+
+/**
+ * Base ViewModel implementation that all other ViewModels should extend.
+ */
+abstract class BaseMvRxViewModel<S : MavericksState>(
```
elihart, line 33: some of this code doesn't need to be duplicated. each rx version can share a `InternalMvRxViewModel` super class which can be shared from a new module

mvrx-rxjava3/src/main/kotlin/com/airbnb/mvrx/MvRxViewModelFactory.kt
```diff
@@ -0,0 +1,7 @@
+package com.airbnb.mvrx
+
+@Deprecated(
+    message = "MvRx has been replaced with Mavericks",
+    replaceWith = ReplaceWith("MavericksViewModelFactory<VM, S>")
+)
+interface MvRxViewModelFactory<VM : MavericksViewModel<S>, S : MavericksState> : MavericksViewModelFactory<VM, S>
```
elihart, line 7: don't duplicate this

elihart (commented):
Thanks for working on this. Can you describe your method? Did you simply copy every file over exactly from the rx2 module and then update the references for rx3?

Some of these files have no relationship to rxjava and should not be duplicated - can you create a common rxjava module to share those?

It is a shame to duplicate most of this code at all, like the subscribe functions, but the functions need to return a Disposable that is specific to each Rx version. Ideally we would be able to provide a base class that contains the bulk of the code, and each rx version would just map the result disposable to the right version. If you can think of a good way to do that it would be great

JagadeeshDamarasingu:
Hi @elihart, when I initially developed this PR, my primary objective was to ensure compatibility with the existing rxJava2 API. I wanted users to be able to seamlessly transition from rxJava2 to rxJava3 by simply refactoring their codebase. To achieve this, I made sure not to disrupt the rxJava2 module, which is why you might notice a few duplicate classes.

Thank you for your feedback, I will make improvements to this MR accordingly.

docs/rxjava3.md
JagadeeshDamarasingu: I am currently in the process of building an rxjava demo app, and as part of that, I had intended to create comprehensive documentation that includes various examples taken directly from the demo itself,  However, I apologize for not including it at this time.

gradle/libs.versions.toml
JagadeeshDamarasingu: this one I added as part of rxjava demo, which is still in progress

.idea/codeStyles/Project.xml
JagadeeshDamarasingu: I have noticed that there might have been some automatic changes made by Android Studio itself. I will investigate this further to identify and verify any modifications that have occurred.

.idea/kotlinc.xml
JagadeeshDamarasingu: sure

PR #686, opened 2023-07-11
Add eleventh ViewModel mock builder

turbochan:
Adds an eleventh ViewModel mock builder for mocking MavericksViews with _so_ many ViewModels.


elihart (approved):
thanks!

PR #707, opened 2024-01-25
Fix crash in state restoration with multiple of 32 fields

rossbacher:
If the field count is exactly multiples of 32 the bitmap count would be be too high (e.g. 2 with field count 32 and 3 with field count 64) leading to a crash when calling the copy function via reflection.

Added test that tests that case.

### DIFF
--- a/mvrx/src/main/kotlin/com/airbnb/mvrx/PersistState.kt
+++ b/mvrx/src/main/kotlin/com/airbnb/mvrx/PersistState.kt
@@ -7,6 +7,7 @@ import android.os.Parcelable
 import java.io.Serializable
 import java.lang.reflect.Constructor
 import java.lang.reflect.Method
+import kotlin.math.ceil
 
 /**
  * Annotate a field in your [MavericksViewModel] state with [PersistState] to have it automatically persisted when Android kills your process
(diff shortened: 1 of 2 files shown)


mvrx/src/test/kotlin/com/airbnb/mvrx/PersistedStateTest.kt
```diff
@@ -238,6 +238,58 @@ class PersistedStateTest : BaseTest() {
         persistMavericksState(State())
     }
 
+    @Test
+    fun testClassWithExactly32Parameters() {
+        data class StateWith32Params(
+            val p0: Int = 0,
```
elihart, line 244: does it matter that there are only 16 persist state annotations here? it's confusing why only half of the params have it, but I suppose it shouldn't matter for the purposes of this bug?

mvrx/src/main/kotlin/com/airbnb/mvrx/PersistState.kt
```diff
@@ -128,7 +128,7 @@ fun <T : MavericksState> restorePersistedMavericksState(
     val fieldCount = constructor.parameterTypes.size
 
     // There is 1 bitmask for each block of 32 parameters.
-    val parameterBitMasks = IntArray(fieldCount / 32 + 1) { 0 }
+    val parameterBitMasks = IntArray(fieldCount / 32 + if (fieldCount % 32 != 0) 1 else 0) { 0 }
```
elihart, line 131: might be simpler to round up to the next int `ceil(fieldCount/32.0).toInt()`

elihart (approved):
thanks for the fix and test 🙏

mvrx/src/test/kotlin/com/airbnb/mvrx/PersistedStateTest.kt
rossbacher: It's the number of parameters in the constructor that trigger the bug not the number of persisted elements, I copied this from another test, that is why this is like this, but we could also persist them all or persist none of them.
The test was failing before the fix.

mvrx/src/main/kotlin/com/airbnb/mvrx/PersistState.kt
rossbacher: Yeah, that was what I was looking for yesterday but could not come up with it. I'll update it.

Trav5599 (approved):

PR #720, opened 2024-08-26
Better error message in PersistState 

rossbacher:
Log a better error message that contains the type of the state class that could not be persisted for easier debugging.

Update unit tests to also check the exception message.


elihart (approved):
Thanks!

